PR #50, opened 2026-04-22
fix: resolve SQL logical issue with `commit` keyword

miketheman:
commit is an SQLite reserved word.
The old query quoted it in the WHERE clause but not in SET commit = ... or SELECT commit FROM scans, so the parser rejected the whole statement. gdb.Exec logged the error but didn't propagate it, masking the fact that the backfill silently did nothing.

```go
2026/04/22 11:56:25 .../scrutineer/internal/db/db.go:496 SQL logic error: near "commit": syntax error (1)
[0.007ms] [rows:0]
		UPDATE findings
		SET commit = (
			SELECT commit FROM scans WHERE scans.id = findings.scan_id
		)
		WHERE `commit` IS NULL OR `commit` = ''
```

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -495,10 +495,10 @@ func BackfillFindingRepository(gdb *gorm.DB) {
 	`)
 	gdb.Exec(`
 		UPDATE findings
-		SET commit = (
-			SELECT commit FROM scans WHERE scans.id = findings.scan_id
+		SET "commit" = (
+			SELECT "commit" FROM scans WHERE scans.id = findings.scan_id
 		)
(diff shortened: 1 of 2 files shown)


internal/db/db.go
```diff
@@ -495,10 +495,10 @@ func BackfillFindingRepository(gdb *gorm.DB) {
 	`)
 	gdb.Exec(`
 		UPDATE findings
-		SET commit = (
-			SELECT commit FROM scans WHERE scans.id = findings.scan_id
+		SET "commit" = (
```
miketheman, line 498: Thinking about this more, this could probably be better represented as `commit_hash`, but that's a different conversation.

PR #57, opened 2026-04-24
Fix Docker builds for both scrutineer and -runner containers

edorian:
Address three issues when building.

I couldn't get the project on macOS/arm64 and had to solve three installation issues

 - semgrep==1.115.0 was no longer installable, minimal update to 1.116.0
 - brief now installs the latest released version github.com/git-pkgs/brief/cmd/brief@v0.5.2 over trying to find @v0.10 (which I don't think exists as a tag or release) 
 - dedicated rust:1.88-alpine stage for zizmor as it could not be build with Alpine’s older Rust/Cargo version

Tested with:

```
docker build -t scrutineer .
docker build -t scrutineer-runner -f Dockerfile.runner .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,11 @@ FROM node:22-alpine AS claude
 RUN npm install -g @anthropic-ai/claude-code@1.0.17
 
 FROM python:3.13-alpine AS python-tools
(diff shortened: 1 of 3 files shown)


andrew:
Thanks!

Issue #56, opened 2026-04-22
Reproducibility hardening: digest-pin base images, lockfile-based tool installs, Dependabot coverage


michaelwinser:
Independent follow-up to #53. The repo already has the supply-chain discipline for GitHub Actions (SHA-pinned refs, consistent with zizmor's `unpinned-uses` recommendation). The Dockerfile pins need the same treatment applied consistently — not just to the leaves.

### Tasks

- [x] **Digest-pin every base image in `Dockerfile.runner`.** Replace `FROM x:tag` with `FROM x:tag@sha256:<digest>`. `alpine:3.21` is a tag, not a digest — Alpine rebuilds it for security patches, which is how `git` silently disappeared from `golang:1.26.2-alpine`. Digest pins make the build content-addressed end-to-end. `docker buildx imagetools inspect <ref>` yields the digest.
- [x] ~**Python: hash-pinned requirements.**~ Superseded by #103: semgrep now comes from the digest-pinned `semgrep/semgrep` image, so there is no `pip install` step and no requirements file to maintain.
- [x] ~**npm: `npm ci` against a lockfile.**~ Superseded by #103: claude-code is now a checksum-verified musl binary from the GitHub release; the npm stage and `nodejs` are gone from the image.
- [x] **Cargo: `--locked` everywhere.** Done for zizmor in #53; audit nothing else invokes `cargo install` unlocked.
- [x] **Go: already safe.** `go.sum` covers transitive state; no change needed.
- [x] **Dependabot (or Renovate) coverage.** `package-ecosystem: docker` is in `.github/dependabot.yml` and now covers the semgrep base image too. No `pip`/`npm` stanzas needed after #103 since neither lockfile exists; the claude release URL + checksum is the one manual bump remaining.

### Why this over "use `@latest` everywhere"

`@latest` guarantees "works on build day" but not "reproducible builds." For a security scanner that users run months after release, reproducibility is load-bearing. The right answer is not *less* pinning — it's *complete* pinning, with an automated bump process that catches upstream changes early (CI, companion issue) and updates the pins in PRs (Dependabot).

### Related

- #53 — pin-drift bugs this prevents
- ~(companion) CI: build Dockerfile.runner on every PR — catches drift the day it happens~ done in #61
- ~(companion) Publish scrutineer-runner to GHCR — so users consume the immutable artifact, not their own rebuild~ done in #61

PR #61, opened 2026-04-25
Harden runner image: fix builds, digest-pin, CI, GHCR publish

### LINKED ISSUE #52: Running locally and in Docker results in error during scanning
Followed the instructions and made some changes to the Dockerfile (as I was receiving errors during the build for `semgrep` and `zizmor`). Here's my `Dockerfile.runner`:

```dockerfile
# Minimal image for per-job ephemeral containers. Contains only the analysis
# tools, not the web server or database. Built separately:
#   docker build -t scrutineer-runner -f Dockerfile.runner .

FROM node:22-alpine AS claude
RUN npm install -g @anthropic-ai/claude-code@1.0.17

FROM python:3.13-alpine AS python-tools
RUN pip install --no-cache-dir semgrep==1.159.0 zizmor

FROM golang:1.26.2-alpine AS go-tools
RUN GOBIN=/out go install github.com/git-pkgs/git-pkgs@v0.14.0

FROM alpine:3.21
RUN apk add --no-cache git ca-certificates python3 bash nodejs

# claude cli
COPY --from=claude /usr/local/lib/node_modules /usr/local/lib/node_modules
RUN ln -sf /usr/local/lib/node_modules/@anthropic-ai/claude-code/cli.js /usr/local/bin/claude

# semgrep
COPY --from=python-tools /usr/local/lib/python3.13/site-packages /usr/local/lib/python3.13/site-packages
COPY --from=python-tools /usr/local/bin/semgrep* /usr/local/bin/
COPY --from=python-tools /usr/local/bin/zizmor* /usr/local/bin/

# go tools
COPY --from=go-tools /out/* /usr/local/bin/

# Non-root
RUN adduser -D -h /home/runner runner
USER runner
WORKDIR /out
```

This built fine, started the application as documented and then ran a scan for a repository and saw this error in the scan logs:

```
$ git clone --depth 1 https://github.com/....git
$ docker run --rm --network none scrutineer-runner <skill:security-deep-dive>
/usr/bin/env: unrecognized option: S
BusyBox v1.37.0 (2025-11-23 13:10:04 UTC) multi-call binary.
Usage: env [-i0] [-u NAME]... [-] [NAME=VALUE]... [PROG ARGS]
Print current environment or run PROG after setting up environment
-, -i	Start with empty environment
-0	NUL terminated output
-u NAME	Remove variable from environment
[error] docker exited: exit status 1
```

### LINKED ISSUE #53: Dockerfile.runner pins have drifted; image fails to build and run on fresh checkouts
On a fresh checkout of `main` (at 5ced00c), `docker build -f Dockerfile.runner .` fails with multiple independent errors. Each is a consequence of pinning leaf tool versions at a point in time while upstream sources moved on, plus base images that aren't digest-pinned.

### Build-time failures

**1. `semgrep==1.115.0` yanked from PyPI.** PyPI jumps 1.114.0 → 1.116.0; 1.115.0 is missing.
```diff
-RUN pip install --no-cache-dir semgrep==1.115.0
+RUN pip install --no-cache-dir "semgrep>=1.114,<2"
```

**2. `brief@v0.10.0` never existed, and wrong import path.** The `git-pkgs/brief` repo's highest tag is `v0.6.0`. `brief`'s main package is at `cmd/brief`, not the repo root.
```diff
-    GOBIN=/out go install github.com/git-pkgs/brief@v0.10.0
+    GOBIN=/out go install github.com/git-pkgs/brief/cmd/brief@latest
```

**3. `golang:1.26.2-alpine` no longer ships `git`.** `go install` invokes git for VCS stamping.
```diff
 FROM golang:1.26.2-alpine AS go-tools
+RUN apk add --no-cache git
 RUN GOBIN=/out go install ...
```

**4. `zizmor@1.6.0` needs rustc ≥ 1.85, apk cargo ships 1.83; also unlocked installs pick up breaking transitives.** Without `--locked`, cargo resolves a newer `lsp-types` whose API zizmor's source wasn't updated for (missing field `offset_encoding`). Fix: move zizmor to its own stage on `rust:alpine` (tracks latest stable) and use `--locked`.
```diff
-# zizmor (inline in final stage)
-RUN apk add --no-cache --virtual .build-deps cargo && \
-    cargo install zizmor@1.6.0 && \
-    cp /root/.cargo/bin/zizmor /usr/local/bin/ && \
-    rm -rf /root/.cargo && \
-    apk del .build-deps
+FROM rust:alpine AS zizmor-tools
+RUN apk add --no-cache build-base
+RUN cargo install --locked --root /out zizmor
...
+# zizmor (in final stage)
+COPY --from=zizmor-tools /out/bin/zizmor /usr/local/bin/
```

### Runtime failures (image builds but scans fail)

**5. BusyBox env vs GNU `env -S`** — filed as #52. Fix: add `coreutils` to the runtime stage.

**6. `@anthropic-ai/claude-code@1.0.17` doesn't support `--effort`.** `internal/worker/docker.go` passes `--effort` on every scan; claude-code 1.0.17 predates that flag. Bumping the pin is needed — but then exposes (7).

**7. claude-code internal layout moved: `cli.js` → `bin/claude.exe`.** The Dockerfile hardcodes `cli.js` for its manual symlink, which is now dangling. Better: reuse npm's own symlink from the builder stage instead of hardcoding an internal path the package never guaranteed.
```diff
 COPY --from=claude /usr/local/lib/node_modules /usr/local/lib/node_modules
-RUN ln -sf /usr/local/lib/node_modules/@anthropic-ai/claude-code/cli.js /usr/local/bin/claude
+COPY --from=claude /usr/local/bin/claude /usr/local/bin/claude
```

### Root cause (why all these happened at once)

`Dockerfile.runner` pins leaf tool versions against *floating* base-image tags (`alpine:3.21`, `python:3.13-alpine`, `rust:alpine`, `golang:1.26.2-alpine`), with no lockfiles for transitive deps and no CI job that exercises the Dockerfile. That combination is maximally fragile: the author's build worked at its timestamp, and any upstream change (yank, rebuild, new rustc requirement, package layout move) silently invalidates it for later users — because nothing on the project side re-runs the build.

The repo's GitHub Actions refs are correctly SHA-pinned (consistent with zizmor's `unpinned-uses` discipline). The same discipline isn't applied to the Dockerfile.

### Structural fixes tracked separately

- CI builds `Dockerfile.runner` on every PR (catches this class of drift)
- Publish the image to GHCR (avoids every user rebuilding)
- Reproducibility hardening (digest-pin bases, lockfiles, Dependabot coverage)

All diffs above applied locally and building clean end-to-end. Happy to turn into a PR if useful.

### LINKED ISSUE #54: CI: build Dockerfile.runner on every PR and smoke-test tools
`.github/workflows/` currently runs `go test` but never exercises `Dockerfile.runner`. As a result, the entire class of pin-drift bugs summarised in #53 shipped undetected — nothing on the project side re-runs the Docker build until a user tries it.

### Proposal

Add a workflow (or a job in an existing workflow) that runs on PRs touching `Dockerfile.runner`, on every push to `main`, and on a daily schedule:

1. `docker buildx build --load -t scrutineer-runner -f Dockerfile.runner .`
2. Smoke-test each tool:
   - `docker run --rm scrutineer-runner claude --version`
   - same for `semgrep`, `zizmor`, `git-pkgs`, `brief`, `git`
3. Optional second pass: `--help` on each tool, to catch native-library breakage that `--version` doesn't exercise.

### Why daily

Many of the failures in #53 weren't triggered by a repo change — they came from upstream state moving (PyPI yanks, Alpine base rebuilds, etc.). A scheduled run catches those before the next unrelated PR trips over them.

### Why not just gate on the Dockerfile path

A `paths:` filter misses the case where a host-side change (e.g. adding `--effort` to `internal/worker/docker.go`) is incompatible with the pinned image. Running on every PR (or at least every PR that touches `cmd/`, `internal/`, or the Dockerfile) catches host/runner-image version skew like #53 item 6.

### Related

- #53 — bugs this CI would have caught
- (companion) Publish scrutineer-runner to GHCR
- (companion) Reproducibility hardening

### LINKED ISSUE #55: Publish scrutineer-runner to GHCR and pull by digest
Today the README instructs every user to `docker build -t scrutineer-runner -f Dockerfile.runner .` on first run. That's a multi-stage build with several native compilations (zizmor via cargo, semgrep's Python native deps) — ~15 minutes on a fresh machine — and each user builds from scratch against whatever state PyPI, crates.io, Docker Hub, and Alpine repos happen to be in on their build day. That is exactly the drift surface that produced #53.

### Proposal

Publish the runner image from CI to `ghcr.io/alpha-omega-security/scrutineer-runner`:

- `:sha-<short-sha>` per main push (immutable)
- `:latest` on main
- `:v<version>` on release tags

Flip scrutineer's default `--runner-image` to the GHCR ref so a freshly-cloned repo Just Works on `go run ./cmd/scrutineer` with no extra build step.

`Dockerfile.runner` stays in-tree as the source; the local build path remains available for users modifying the image.

### README change

Replace:
```
docker build -t scrutineer-runner -f Dockerfile.runner .
go run ./cmd/scrutineer -skills ./skills
```
with:
```
docker pull ghcr.io/alpha-omega-security/scrutineer-runner:latest
go run ./cmd/scrutineer -skills ./skills
```

(Or drop the pull entirely — docker does it on first `docker run` automatically.)

### Implementation

`docker/build-push-action@<sha>` in the same workflow that'll run the build (companion issue). Authenticate with `GITHUB_TOKEN` — no extra secret needed for GHCR under the same org.

### Follow-ups

- SLSA provenance + Sigstore signatures via `slsa-framework/slsa-github-generator`. Consistent with the project's zizmor-style supply-chain posture (workflows already SHA-pin actions). End-state is a verifiable, signed image provable to have come from this workflow.
- Consider pinning the runner image in config/UI by digest, not tag, so the host binary's expected image can't be swapped out from under it.

### Related

- #53 — pin-drift bugs a pre-built image sidesteps for new users
- (companion) CI: build Dockerfile.runner on every PR
- (companion) Reproducibility hardening

andrew:
Picks up where #57 left off and addresses the rest of the docker issue cluster.

`Dockerfile` and `Dockerfile.runner` fixes: add `coreutils` so claude's `env -S` shebang works on Alpine, bump `claude-code` to 2.1.119 (1.0.17 predates `--effort` which `internal/worker/docker.go` always passes), copy npm's own `/usr/local/bin/claude` symlink rather than hardcoding the package-internal `cli.js` path, add `git` to the go-tools stage, and bump `git-pkgs`/`brief` to current tags.

The smoke test also turned up a pre-existing break: semgrep never actually ran in either image. Its entry script's shebang is `#!/usr/local/bin/python3.13` from the build stage, but the final stage installed apk's `python3` at `/usr/bin/python3` (and 3.12 at that, so the copied 3.13 site-packages would not have resolved anyway). Final stage now bases on `python:3.13-alpine` so interpreter and site-packages match, with `pip` stripped to keep the T11 claim true. `setuptools<81` is added for `pkg_resources`, which an opentelemetry transitive of semgrep still imports.

All `FROM` lines are now digest-pinned. Dependabot already covers the `docker` ecosystem at `/` (which picks up `Dockerfile.<suffix>`); added a group block for it.

New `.github/workflows/runner-image.yml` builds `Dockerfile.runner` on every PR, every push to main, and once daily. It runs `--version` on each bundled tool, then on main pushes the image to `ghcr.io/alpha-omega-security/scrutineer-runner` as `:latest` and `:sha-<short>`. The default `--runner-image` now points at the GHCR ref, so a fresh checkout works with just `go run ./cmd/scrutineer -skills ./skills` and no local docker build. README, `scrutineer.sample.yaml`, and threatmodel T11 updated to match.

Verified locally: both images build, all six tools answer `--version`, `go test -race ./...` and `golangci-lint` pass, `zizmor --pedantic .github/` is clean.

After merge: the first push to main will create `ghcr.io/alpha-omega-security/scrutineer-runner` as a private package. It needs flipping to public in the org's package settings (and linking to this repo) so the new default `--runner-image` works for users without auth.

Follow-up left on #56: hash-pinned `runner-requirements.txt` via `pip-compile --generate-hashes`, `npm ci` against a committed lockfile for claude-code, and corresponding `pip`/`npm` Dependabot ecosystems.

Closes #52
Closes #53
Closes #54
Closes #55
Refs #56

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,3 +26,6 @@ updates:
       interval: weekly
     cooldown:
       default-days: 7
+    groups:
+      docker:
+        patterns: ["*"]
--- a/.github/workflows/runner-image.yml
+++ b/.github/workflows/runner-image.yml
@@ -1,0 +1,77 @@
+name: runner-image
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  schedule:
+    - cron: "30 6 * * *"
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write # push runner image to ghcr.io
+    env:
+      REGISTRY: ghcr.io
+      IMAGE: ghcr.io/${{ github.repository }}-runner
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
+
+      - uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
+        id: meta
+        with:
+          images: ${{ env.IMAGE }}
+          tags: |
+            type=raw,value=latest,enable={{is_default_branch}}
+            type=sha,prefix=sha-
+
+      - uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+        if: github.event_name == 'push'
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f # v7.1.0
+        with:
+          context: .
+          file: Dockerfile.runner
+          load: true
+          push: false
+          tags: |
+            scrutineer-runner:ci
+            ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: smoke test
+        run: |
+          set -euxo pipefail
+          docker run --rm scrutineer-runner:ci claude --version
+          docker run --rm scrutineer-runner:ci semgrep --version
+          docker run --rm scrutineer-runner:ci zizmor --version
+          docker run --rm scrutineer-runner:ci git-pkgs --version
+          docker run --rm scrutineer-runner:ci brief --version
+          docker run --rm scrutineer-runner:ci git --version
+
+      - name: push
+        if: github.event_name == 'push'
+        env:
+          TAGS: ${{ steps.meta.outputs.tags }}
+        run: |
+          set -euo pipefail
+          echo "$TAGS" | while read -r tag; do
+            [ -n "$tag" ] && docker push "$tag"
+          done
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,43 @@
-FROM golang:1.26.2-alpine AS build
+FROM golang:1.26.2-alpine@sha256:f85330846cde1e57ca9ec309382da3b8e6ae3ab943d2739500e08c86393a21b1 AS build
 WORKDIR /src
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . .
 RUN CGO_ENABLED=0 go build -o /scrutineer ./cmd/scrutineer
 
-FROM node:22-alpine AS claude
-RUN npm install -g @anthropic-ai/claude-code@1.0.17
+FROM node:22-alpine@sha256:8ea2348b068a9544dae7317b4f3aafcdc032df1647bb7d768a05a5cad1a7683f AS claude
+RUN npm install -g @anthropic-ai/claude-code@2.1.119
 
-FROM python:3.13-alpine AS python-tools
-RUN pip install --no-cache-dir semgrep==1.116.0
+FROM python:3.13-alpine@sha256:420cd0bf0f3998275875e02ecd5808168cf0843cbb4d3c536432f729247b2acc AS python-tools
+RUN pip install --no-cache-dir semgrep==1.116.0 "setuptools<81"
 
-FROM rust:1.88-alpine AS zizmor-build
+FROM golang:1.26.2-alpine@sha256:f85330846cde1e57ca9ec309382da3b8e6ae3ab943d2739500e08c86393a21b1 AS go-tools
+RUN apk add --no-cache git
+RUN GOBIN=/out go install github.com/git-pkgs/git-pkgs@v0.15.3 && \
+    GOBIN=/out go install github.com/git-pkgs/brief/cmd/brief@v0.6.0
+
+FROM rust:1.88-alpine@sha256:9dfaae478ecd298b6b5a039e1f2cc4fc040fc818a2de9aa78fa714dea036574d AS zizmor-build
 RUN apk add --no-cache build-base linux-headers
 RUN cargo install --locked --root /out zizmor@1.24.1
 
-FROM alpine:3.21
-RUN apk add --no-cache git ca-certificates python3 bash nodejs
+FROM python:3.13-alpine@sha256:420cd0bf0f3998275875e02ecd5808168cf0843cbb4d3c536432f729247b2acc
+RUN apk add --no-cache git ca-certificates bash nodejs coreutils && \
+    rm -f /usr/local/bin/pip* /usr/local/bin/idle* /usr/local/bin/pydoc*
 
 # scrutineer binary
 COPY --from=build /scrutineer /usr/local/bin/scrutineer
 
 # claude cli
 COPY --from=claude /usr/local/lib/node_modules /usr/local/lib/node_modules
-RUN ln -sf /usr/local/lib/node_modules/@anthropic-ai/claude-code/cli.js /usr/local/bin/claude
+COPY --from=claude /usr/local/bin/claude /usr/local/bin/claude
 
 # semgrep
 COPY --from=python-tools /usr/local/lib/python3.13/site-packages /usr/local/lib/python3.13/site-packages
 COPY --from=python-tools /usr/local/bin/semgrep* /usr/local/bin/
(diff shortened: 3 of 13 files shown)


andrew:
Pushed 81d7dc4 which replaces `--network none` with an allowlisting egress proxy so model-backed skills actually work under the docker runner.

How it works: scrutineer starts a small CONNECT/forward proxy on the host (ephemeral port, all interfaces so the docker bridge can reach it on Linux). The runner container goes on the default bridge with `--add-host host.docker.internal:host-gateway` and `HTTPS_PROXY`/`HTTP_PROXY`/`ALL_PROXY` pointed at the proxy. The proxy 403s anything not on the allowlist and requires a per-process random token via `Proxy-Authorization`, so it isn't an open relay on the LAN.

Default allowlist covers what the bundled skills need: `*.anthropic.com`, `*.ecosyste.ms`, GitHub/GitLab/Codeberg/Bitbucket, npm/PyPI/RubyGems/crates.io/Go proxy/Packagist/Hex/NuGet, semgrep.dev/OSV/NVD/cwe.mitre.org, and `host.docker.internal` for the local skill API. Extra hosts can be added with `egress_allow:` in `scrutineer.yaml`.

The proxy rewrites `host.docker.internal` to `127.0.0.1` when dialing, so `context.json` now points skills at `http://host.docker.internal:<port>/api` and that reaches the loopback-bound web server through the proxy. The token gates that path too.

This is cooperative, not enforced: a process in the container that ignores the proxy env can still dial out directly. Everything in the image is ours and pinned (T11), so that's accepted for now; hard enforcement via an `--internal` network is the follow-up under #3. New T13 in `threatmodel.md` spells this out.

@mechko to re-test: pull the branch, make sure `ANTHROPIC_API_KEY` or `CLAUDE_CODE_OAUTH_TOKEN` is in your env, and run a scan. The startup log will show `egress_proxy_port=<n>`; denied hosts log as `egress denied host=<h>`.

mechko:
Thanks Andrew, I just ran a scan using the runner, it seems to work now.

PR #76, opened 2026-04-27
fix: show Findings tab based on all repo findings, not just latest scan

mbarbero:
Agent-Logs-Url: https://github.com/mbarbero/scrutineer/sessions/6eaad190-28e6-49e6-9008-a96fe8c1c060


mbarbero:
already fixed :D

PR #85, opened 2026-04-28
feat: add support for configurable clone depth

### LINKED ISSUE #73: Configurable clone depth: shallow vs full
Clones are currently hardcoded to `--depth 1` in `internal/worker/clone.go` and `internal/skills/git.go`. That's the right default for speed and disk, but some skills want full history (blame-based authorship checks, commit cadence heuristics, looking for secrets in old commits).

Add a config option, something like:

    clone: shallow   # or "full"

Default stays shallow. When set to `full`, drop the `--depth 1` arg from the worker clone and the skills_repo clone. Existing shallow clones in `data/work/` should probably be unshallowed with `git fetch --unshallow` on next use rather than re-cloned, but blowing them away and re-cloning is also fine for a first pass.

Might be worth allowing per-repo or per-scan override later, but a global toggle covers the immediate need.

alexandre-daubois:
Fix #73

As said in the issue, this is a global setting instead of a per-repo setting.


andrew:
Thanks!

andrew (approved):

PR #93, opened 2026-04-29
config: add anthropic_base_url for custom API endpoints

andrewjkerr:
## Summary

This MR adds `-anthropic-base-url` flag and `anthropic_base_url` YAML field so operators can point `scrutineer` at a custom Anthropic API endpoint (e.g. an LLM proxy, like I wanted to!). Falls back to the `ANTHROPIC_BASE_URL` env var when neither the flag nor config is set. This means the resolution order: CLI flag > YAML config > env var > default.

> [!NOTE]
> I wasn't sure if this should be a config option or an env var - I chose to do both, because my understanding is that `ANTHROPIC_BASE_URL` _may_ be set on a local system, but it may not always be. More than happy to update based on feedback, though! 

The resolved URL is passed as `ANTHROPIC_BASE_URL` to Docker containers and its hostname is automatically added to the egress proxy allowlist, so a single config change is all that's needed. The scan log now surfaces the resolved URL so operators can confirm it's being forwarded.

## Test

Tests cover the three merge precedence cases (config fills default, explicit flag wins, zero-value config leaves default alone) plus `baseURLHost` extraction. I also made sure this ran locally, confirming that the `ANTHROPIC_BASE_URL` is passed to the runner container & the requests hit the LLM proxy a-ok. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ You need [Go 1.26+](https://go.dev/dl/), [Docker](https://docs.docker.com/get-do
     git clone https://github.com/alpha-omega-security/scrutineer
     cd scrutineer
     export ANTHROPIC_API_KEY=sk-ant-...
+    export ANTHROPIC_API_URL=https://...  # optional: custom API endpoint
     go run ./cmd/scrutineer -skills ./skills
 
 Then open http://127.0.0.1:8080.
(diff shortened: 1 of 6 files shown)


andrew:
Thanks, I fixed up the linting errors

andrew (approved):

PR #98, opened 2026-04-29
feat: implement CSAF document generation and validation for findings

alexandre-daubois:
First pass at the CSAF 2.0 VEX export (part of #8). This emits a single product per finding, schema-validated at runtime.

Per-dependent VEX statements are deferred as they consume the `finding_dependents` table created by #9. Once #9 is in, wiring its rows into the CSAF generator is a small follow-up 🙂

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ toolchain go1.26.2
 require (
 	github.com/git-pkgs/sbom v0.1.0
 	github.com/glebarez/sqlite v1.11.0
+	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2
 	github.com/yuin/goldmark v1.8.2
 	gopkg.in/yaml.v3 v3.0.1
 	gorm.io/gorm v1.31.1
@@ -22,7 +23,6 @@ require (
 	github.com/mattn/go-isatty v0.0.22 // indirect
 	github.com/ncruces/go-strftime v1.0.0 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
-	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2 // indirect
 	golang.org/x/sys v0.43.0 // indirect
 	golang.org/x/text v0.36.0 // indirect
 	modernc.org/libc v1.72.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/mattn/go-isatty v0.0.22 // indirect
 	github.com/ncruces/go-strftime v1.0.0 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
+	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2 // indirect
 	golang.org/x/sys v0.43.0 // indirect
 	golang.org/x/text v0.36.0 // indirect
 	modernc.org/libc v1.72.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+github.com/dlclark/regexp2 v1.11.0 h1:G/nrcoOa7ZXlpoa/91N3X7mM3r8eIlMBBJZvsz/mxKI=
+github.com/dlclark/regexp2 v1.11.0/go.mod h1:DHkYz0B9wPfa6wondMfaivmHpzrQ3v9q8cnmRbL6yW8=
 github.com/dustin/go-humanize v1.0.1 h1:GzkhY7T5VNhEkwH0PVJgjz+fX1rhBrR7pRT3mDkpeCY=
 github.com/dustin/go-humanize v1.0.1/go.mod h1:Mu1zIs6XwVuF/gI1OepvI0qD18qycQx+mFykh5fBlto=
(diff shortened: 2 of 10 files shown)


andrew (commented):
Thanks for this — the structure is good, the embedded-schema approach with the provenance README is exactly right, the test coverage is solid, and `santhosh-tekuri/jsonschema/v6` is a clean leaf dep with no transitive bloat. A few semantic issues to sort before merge:

`pickJustification` doesn't work as written. It picks a not-affected justification based on which prose field is non-empty, but the security-deep-dive schema makes `boundary` and `validation` required on every finding, so every rejected finding hits the first branch and emits `vulnerable_code_cannot_be_controlled_by_adversary` regardless of why it was rejected. Field presence says nothing about content here. Either add an explicit justification column on `Finding` or drop `flags` from this first pass.

`mapProductStatus` has two wrong mappings. `FindingPublished` means the advisory went out, not that the bug is fixed — it should stay `known_affected` unless `FixVersion`/`FixCommit` is set. `FindingDuplicate` means "tracked elsewhere", not "not vulnerable"; mapping it to `known_not_affected` is misleading. Duplicates probably shouldn't export at all.

`baseSeverity` can contradict `baseScore`. `severityLabel` prefers `f.Severity` over the score-derived band, so the seed fixture produces `baseScore: 9.8, baseSeverity: HIGH`. The CVSS schema doesn't cross-check those so it passes validation, but per FIRST a 9.8 is CRITICAL. Derive it from the score only.

The workaround remediation uses `f.Trace` as `details`. Trace is the input-flow prose, not a workaround description, so that emits "this variable, assigned from this method's return..." as the remediation text. `f.Notes` or just the fallback string for now.

Smaller bits:

- `go.mod` lists jsonschema as `// indirect` but it's imported directly in `finding_csaf.go` and `server.go`; needs `go mod tidy`.
- The `if initial == ""` / `if current == ""` checks are dead — `time.Time.Format` never returns empty, a zero time becomes `0001-01-01T00:00:00Z`. Check `f.CreatedAt.IsZero()` instead.
- `csafReference.Summary` is `omitempty` but the CSAF schema requires `summary`; a `FindingReference` with empty `Summary` and empty `Tags` will 500 on export. Fall back to the URL.
- A `CVSS:4.0/...` vector gets its version coerced to `"3.1"` while keeping the original `vectorString`, which then fails the v3.1 pattern at validation time. Return nil for non-3.x prefixes rather than coercing.
- The `sync.Once` cache plus the `Server.csafSchema` field is redundant since `New()` runs once; either is fine on its own.

alexandre-daubois:
Thanks for the thorough review! I addressed your comments in dd04bee8436cf680ce91b0a3637c435731ea45fb.

andrew (commented):
Two things worth addressing before merge:

1. **No PURL in the product tree.** The `csafProductIdentifier` struct has a `PURL` field but `buildCSAF` never populates it. The `packages` table already has PURLs for 10k packages -- wiring the repo's packages into `product_identification_helper` would make the document much more useful for automated tooling, since PURL is how downstream consumers match a VEX statement to their dependency graph.

2. **No `flags` on `known_not_affected` products.** CSAF consumers expect a `flags` entry with a `label` (e.g. `vulnerable_code_not_present`, `vulnerable_code_cannot_be_controlled_by_adversary`) when a product is marked `known_not_affected`. Without it the document is schema-valid but not very useful as a VEX statement. This is the natural place to wire in per-dependent reachability justifications from #9, but even without that, rejected findings could carry a justification derived from `resolution` or the analyst's notes.

Everything else looks good. The schema validation at runtime, the status mapping logic, and the test coverage are solid. CVSS v4.0 and monorepo product trees can be follow-ups.

alexandre-daubois:
> 1. **No PURL in the product tree.** The `csafProductIdentifier` struct has a `PURL` field but `buildCSAF` never populates it. The `packages` table already has PURLs for 10k packages -- wiring the repo's packages into `product_identification_helper` would make the document much more useful for automated tooling, since PURL is how downstream consumers match a VEX statement to their dependency graph.

A finding maps to a repo, and a repo can have multiple packages rows. Should we emit one product branch per package, or pick a canonical one? And emit no PURL when the repo has no packages row, does it sound good?

> 2. **No `flags` on `known_not_affected` products.** CSAF consumers expect a `flags` entry with a `label` (e.g. `vulnerable_code_not_present`, `vulnerable_code_cannot_be_controlled_by_adversary`) when a product is marked `known_not_affected`. Without it the document is schema-valid but not very useful as a VEX statement. This is the natural place to wire in per-dependent reachability justifications from #9, but even without that, rejected findings could carry a justification derived from `resolution` or the analyst's notes.

The `Resolution` enum doesn't map cleanly to CSAF's fixed label set, and notes are free-form so no reliable mapping there either. Should it default everything to `vulnerable_code_not_present` for now and revisit with #9, or hold this until #9?

andrew:
> default everything to `vulnerable_code_not_present`

Yeah lets go with that for now

andrew:
> A finding maps to a repo, and a repo can have multiple packages rows. Should we emit one product branch per package, or pick a canonical one? And emit no PURL when the repo has no packages row, does it sound good?

yep sounds good

alexandre-daubois:
Comments addressed and rebased the branch

andrew:
perfect, thanks!

PR #165, opened 2026-05-06
web: show skill name with link on scan detail page

### LINKED ISSUE #127: Show skill name on scan detail page
The scan show page has tabs for result, log, and prompt but doesn't show which skill was used. Add a visible skill indicator, either as a tab linking to the skill definition or as metadata in the page header, so you can tell at a glance what ran.

alexandre-daubois:
Fix #127, feedback welcome on the styling!

Capture d’écran 2026-05-06 à 15 31 49


andrew:
lgtm

PR #173, opened 2026-05-07
feat: add preferred model configuration for skills

### LINKED ISSUE #152: Allow skills to declare a preferred model in SKILL.md metadata
Right now every scan uses the single model configured on the server, regardless of which skill is running. Several of the simpler skills (`metadata`, `maintainers`, `subprojects`, `sbom`, `zizmor`) are mostly mechanical extraction or tool-wrapping and would run fine on sonnet, which is faster and cheaper. The heavier reasoning skills (`security-deep-dive`, `reachability`, `verify`, `patch`) still want the bigger model.

Proposal: add an optional `scrutineer.model` key to the SKILL.md `metadata` block alongside the existing `scrutineer.output_file` / `scrutineer.output_kind`:

```yaml
metadata:
  scrutineer.output_file: report.json
  scrutineer.output_kind: structured
  scrutineer.model: claude-sonnet-4-6
```

Wiring:
- parse it in `internal/skills/parse.go` next to `metaOutputFile` / `metaOutputKind`
- store on `db.Skill`
- when enqueueing a scan in `internal/worker/skill.go`, use the skill's model if set, otherwise fall back to the server default
- `SkillJob.Model` and the `--model` flag plumbing already exist so the runner side needs no change

A scan-level override (e.g. from the API or UI) should still win over the skill's preference, so precedence would be: explicit scan model > skill metadata > server default.

alexandre-daubois:
Fix #152

image

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -216,6 +216,7 @@ func run(log *slog.Logger) error {
 		return fmt.Errorf("queue: %w", err)
 	}
 
+	skills.ModelValidator = web.ValidModel
 	if err := loadSkills(log, gdb, f.dataDir, f.skillLocal, f.skillsRepo, f.fullClone()); err != nil {
 		return err
 	}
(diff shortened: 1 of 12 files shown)


andrew (approved):

andrew:
Thanks for the continued excellent prs!

PR #174, opened 2026-05-07
feat: add support for "Retry all failed scans"

### LINKED ISSUE #128: Add bulk retry for failed scans
Right now retrying failed scans means clicking each one individually or scripting curl calls against `/scans/{id}/retry`. With hundreds of failures after an outage or config change this is tedious.

Add a way to retry all failed scans at once, either a button on the scans index when filtered to `status=failed`, or a `POST /scans/retry-failed` endpoint that requeues everything in one go.

alexandre-daubois:
Fix #128

image

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -197,6 +197,7 @@ func (s *Server) Handler() http.Handler {
 	mux.HandleFunc("GET /advisories", s.advisoriesList)
 	mux.HandleFunc("GET /scans/{id}", s.scanShow)
 	mux.HandleFunc("POST /scans/{id}/retry", s.scanRetry)
+	mux.HandleFunc("POST /scans/retry-failed", s.scansRetryFailed)
 	mux.HandleFunc("POST /scans/{id}/cancel", s.scanCancel)
 	mux.HandleFunc("GET /scans/{id}/log", s.scanLog)
 	mux.HandleFunc("GET /usage", s.usage)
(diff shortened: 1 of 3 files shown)


andrew (changes requested):
Thanks for picking this up. One blocker, then a few smaller notes.

Failed scans never leave `status=failed`. `scanRetry` (and this bulk handler) creates a new queued scan and leaves the original row alone, so the set only grows. After an outage: 200 failures, click, 200 new scans, 180 succeed, 20 fail again, now 220 rows at `status=failed`. Click again for the stragglers and you've just re-enqueued all 220, including 200 that already succeeded on the previous retry. That's exactly the scenario in #128.

I think the fix is to skip any failed scan where a later scan with the same `(repository_id, skill_id, sub_path, ref, finding_id)` already exists at `queued`, `running`, or `done`. That matches what "retry all failed" means in practice: re-run the ones that are still failed, not every historical attempt. Deduping to the latest failed per tuple before enqueueing would also work.

Smaller bits:

- The `hx-confirm` text says "current view" but the handler ignores pagination and re-enqueues every match across all pages. After the dedupe that's the right behaviour, but the wording should say so ("Re-enqueue every failed `{{.Skill}}` scan?" or similar).
- Button visibility keys on `.Scans`, which is the paginated page slice. `.Page.Total` would be exact, though in practice nobody lands on an empty trailing page.
- `q.Find(&scans)` loads every failed row including `Log` and `Report` text. Fine at our scale, but a `Select("id, repository_id, skill_id, kind, model, finding_id, sub_path, ref")` keeps it cheap if someone's sitting on thousands.

The rest looks good: route doesn't collide with `/scans/{id}/retry`, the per-scan enqueue mirrors `scanRetry` exactly, the skill filter and redirect target are right, and the tests assert on rendered text and cover the skip conditions and SubPath preservation.

alexandre-daubois:
Thanks! Comments should all be addressed

andrew (approved):

andrew:
Thanks!

PR #186, opened 2026-05-07
skills/zizmor: handle full exit code range and 0-indexed rows

andrew:
zizmor encodes the highest finding severity in its exit code (10 through 14), but `scan.py` only accepted 0 or 14. Any repo whose worst finding was below High, for example uriparser which tops out at Medium and exits 13, was treated as a tool error and reported zero findings.

Switched to `--no-exit-codes` so zizmor returns 0 whenever it runs cleanly and non-zero only on real failures, which lets the returncode check become a plain `!= 0`.

Also fixed location line numbers. zizmor's `start_point.row` is 0-indexed, so reported locations were one line early. Changed the truthiness check to `is not None` at the same time so a finding on row 0 still gets a line number.

### DIFF
--- a/skills/zizmor/scripts/scan.py
+++ b/skills/zizmor/scripts/scan.py
@@ -29,13 +29,12 @@ def main():
         return
 
     proc = subprocess.run(
-        ["zizmor", "--format", "json", ".github/workflows"],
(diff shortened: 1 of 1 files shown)


mechko:
I tested this locally, lgtm now.

PR #195, opened 2026-05-08
security-deep-dive: check existing mitigations before recommendations

### LINKED ISSUE #190: CSP recommendations should check for existing CSP configuration
In Apache Tooling we took Scrutineer for a spin and we are really happy with the results — the tool is very polished and works great, and the analyst-style commentary that contextualizes Semgrep findings is genuinely a step above what we've seen from other static analysis frontends. Most of the triage was spot-on. We did spot one pattern worth flagging, in the spirit of helping the tool get even sharper.

## Summary

When Scrutineer analyzes XSS-adjacent findings (e.g. `generic.html-templates.security.var-in-href`), it recommends adding a Content Security Policy as a defense-in-depth mitigation without first checking whether the application already sets one. This produces misleading recommendations in projects that already have CSP configured.

## Example

In a recent run on [`apache/tooling-trusted-releases`](https://github.com/apache/tooling-agents/tree/main/scrutineer/tooling-trusted-releases), Scrutineer produced 102 `var-in-href` findings and recommended:

> Add CSP header (`Content-Security-Policy: default-src 'self'; script-src 'self'`) — *Eliminates `var-in-href` exploit path globally*

The repository already configures a more comprehensive CSP in [`atr/server.py`](https://github.com/apache/tooling-trusted-releases/blob/main/atr/server.py), inside a function named `_app_setup_security_headers` whose docstring is `"""Setup security headers including a Content Security Policy."""`. The relevant code:

```python
csp_directives = [
    "default-src 'self'",
    "script-src 'self'",
    "style-src 'self' 'unsafe-inline'",
    "img-src 'self' https://apache.org https://incubator.apache.org https://www.apache.org data:",
    "font-src 'self'",
    "connect-src 'self'",
    "frame-src 'none'",
    "object-src 'none'",
    "base-uri 'none'",
    "form-action 'self'",
    "frame-ancestors 'none'",
]
csp_header = "; ".join(csp_directives)

# ...

@app.after_request
async def add_security_headers(response: quart.Response) -> quart.Response:
    # ...
    response.headers["Content-Security-Policy"] = csp_header
    # ...
```

The deployed policy is strictly stronger than Scrutineer's recommendation: it adds `object-src 'none'`, `base-uri 'none'`, `frame-ancestors 'none'`, `form-action 'self'`, and others, all of which are relevant to the `var-in-href` exploit class.

## Suggested behavior

Before recommending CSP as a remediation, Scrutineer could:

1. Search the codebase for existing CSP configuration. The cheap signals include: string literal `"Content-Security-Policy"`, header assignments matching `headers["Content-Security-Policy"]`, common middleware imports (`flask-talisman`, `secure`, `quart-csp`), and `after_request` / `@app.after_request` handlers that set headers. Function and docstring search would also have caught this case immediately.
2. If a CSP is found, either suppress the recommendation or reframe it as "verify the existing CSP at `<path>:<line>` covers `script-src` and `object-src`" with a pointer to the location.
3. If no CSP is found, the current recommendation is fine — but the recommended directives should match what's needed for the findings (the suggested two-directive CSP is weaker than typical baselines).

The same principle probably applies to other defense-in-depth recommendations (HSTS, X-Frame-Options, Referrer-Policy, Permissions-Policy) — checking the project's existing posture before recommending should be a general behavior. In this same `_app_setup_security_headers` function the project also sets `Permissions-Policy` and `X-Permitted-Cross-Domain-Policies`, so a tool that misses the CSP is likely missing those too.

**Suggested labels:** `bug`, `skill`

andrewmusselman:
Closes #190.

Step 2's trust-boundary check decides whether a sink is exploitable based on whether the input is attacker-controlled. It does not currently look at whether the project has already mitigated the danger somewhere else. Without that step the model writes recommendations against codebases that have already addressed the issue, ranging from noise to actively misleading advice (a stricter existing control taken at face value gets weakened).

The reporter case in #190: a deep-dive against a Quart application flagged 102 `var-in-href` template sinks and recommended adding a Content Security Policy. The repo already configured a stricter CSP in an `after_request` handler. The recommendation, applied literally, would have removed directives the project already had.

Add an explicit pass between Step 2's rule-out branch and the precondition-subsumes branch: grep the codebase for the relevant control, cite its file and line in the rule-out reason if found, and do not recommend adding what already exists. The wording covers controls across domains (centralised handlers, framework defaults, build flags, sanitiser libraries, type-system rules, project-wide lints), with CSP as one illustrative instantiation rather than the focus.

### DIFF
--- a/skills/security-deep-dive/SKILL.md
+++ b/skills/security-deep-dive/SKILL.md
@@ -95,6 +95,8 @@ For agentic sinks, the boundary is crossed when user-controlled content reaches 
 
 If the boundary check rules the sink out — input is internal, or comes from a trusted documented source — write the reason and move to the next sink.
 
+Even where the input is attacker-controlled, check whether the project has already mitigated the danger. The mitigation may live in a centralised handler, a framework default that escapes or binds, a build flag, a sanitiser library, a type-system rule, or a project-wide lint. Grep for it — the directive, the helper, the import, the config line. If the control is in place, rule the sink out at this step and cite the file and line of the existing control in the reason. Do not recommend adding a control the project already has; if the existing one is stricter than what you would have suggested, an analyst applying the advice literally weakens the project. As an example: a template `var-in-href` warning resolves here in a project whose request lifecycle sets `Content-Security-Policy: script-src 'self'`, with the rule-out reason citing that handler.
+
 Even where the input is attacker-controlled, check the precondition does not subsume the conclusion. If reaching the sink requires the attacker to already hold a capability equal to or stronger than what the sink grants — write access to a directory documented as holding executable hooks, MITM position on a connection the finding claims to let them influence — the finding is circular. The attack path's first step already arrives at its last. Write "precondition subsumes conclusion" and move to the next sink.
 
 ### Step 3: Validate


andrew:
Thanks!

PR #196, opened 2026-05-09
feat(skills): add Authorization & Access Control (IDOR) to security-deep-dive

N0tre3l:
Hi team,

While reviewing the `security-deep-dive` skill, I noticed that the current sink classes comprehensively cover technical vulnerabilities (RCE, SSRF, memory safety, etc.) but completely miss logical flaws like Broken Object Level Authorization (BOLA/IDOR). 

Automated LLM scanning often assumes that if an object ID is provided, the operation is legitimate, overlooking the crucial step of verifying whether the object belongs to the current user's session.

I added `Authorization and access control` to the list of sink classes to ensure the model actively looks for proper session/role verification before sensitive operations.


andrew:
Thanks for this. The gap is real for application targets, but the placement doesn't fit the skill's audit model.

`security-deep-dive` is library-first: the trust-boundary section says "the attacker is not the developer calling the library", and every other sink class is a code primitive you can grep for (eval, exec, open, deserialise, `messages=`). IDOR needs an authenticated user, per-user data ownership, and request handlers, which most of what scrutineer scans (packages, parsers, CLIs) doesn't have. Adding it to the universal sink list would have the model looking for "the current user's session" in repos with no sessions, and there's no grep target in the line as written; the other entries tell the model what to search for, this one describes the absence of a check.

The right home is conditional on the threat-model output. Phase 1 already loads the threat-model report and uses its `entry_points` and `adversaries` to fill `boundaries[]`. An authz sink class that only applies when `adversaries` includes authenticated users with distinct data, and that enumerates handlers loading a record by request-supplied ID without comparing the loaded record's owner to the session principal, would fit that flow and give the model something concrete to find.

Happy to take a reworked version gated that way, or we can track it as an issue if you'd rather leave it with us.

N0tre3l:
@andrew 
Thanks for the detailed feedback! You're absolutely right—unconditional checks in non-application targets would generate noise and confuse the model.

I have updated the PR to make the IDOR sink conditional. It now explicitly requires `authenticated users with distinct data` in the Phase 1 threat model boundaries, and gives the model a concrete grep target (`handlers loading a record by a request-supplied ID without comparing the loaded record's owner to the session principal`).

Let me know if this version fits the audit flow better!

andrew:
Thanks for this — gating the IDOR sink on Phase 1's authenticated-users-with-distinct-data boundary is the right scoping.

Two things before merge:

- #195 just landed in `skills/security-deep-dive/SKILL.md` (different section, but your base is a few commits behind). Could you rebase onto current `main`?
- CI hasn't run yet because this is a first contribution from your fork; once you push the rebase I'll approve the workflow run.

andrew (approved):

andrew:
Thanks!

PR #198, opened 2026-05-10
Add a note about using Scrutineer when you have a sandboxed Claude Code config

### LINKED ISSUE #197: Usage with Claude Code's sandbox mode
I have a pretty strict global Claude config to force it to always use sandbox mode and other restrictions. Given that this is running Claude Code in a Docker container, I would assume it's fine to remove these protections, is there a recommended way to do that / recommended way of making doing that safe?

```json
{
  "$schema": "https://json.schemastore.org/claude-code-settings.json",
  "permissions": {
    "disableBypassPermissionsMode": "disable"
  },
  "disableSkillShellExecution": true,
  "allowedHttpHookUrls": [],
  "httpHookAllowedEnvVars": [],
  "sandbox": {
    "enabled": true,
    "failIfUnavailable": true,
    "autoAllowBashIfSandboxed": false,
    "allowUnsandboxedCommands": false,
    "network": {
      "allowedDomains": [
        "github.com",
        "api.github.com"
      ],
      "allowLocalBinding": true
    },
    "enableWeakerNetworkIsolation": true
  },
  "advisorModel": "opus"
}
```

connorshea:
I can put this in another section of the README if there's a specific place we want it, but nothing really stood out to me as fitting.

I should note that we may want to clarify a few things before merging:

- Does Claude Code run within the Docker container when run with Scrutineer, I'm fairly certain it does not?
- Rather than recommending that the user disable sandboxing to use Scrutineer, we would ideally provide a sandbox config that works and allows the necessary ports/commands/filesystem locations for Scrutineer to be run.
Fixes #197.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,15 +204,13 @@ The config file can also replace the model pick list and pin the default model:
       - name: Opus
         id:   claude-opus-4-7
 
-## Interaction with Claude Code Sandboxing
+## Sandboxed Claude Code configs
 
-If you use [sandboxing for Claude Code](https://code.claude.com/docs/en/sandboxing) and your `~/.claude/settings.json` file enforces it, Scrutineer scans will fail due to network restrictions and other limitations.
-
-You should not disable or remove your sandboxing config globally just to run Scrutineer, you can instead create a separate config file with relaxed restrictions and point Claude at it while using Scrutineer.
-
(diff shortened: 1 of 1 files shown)


andrew:
Pushed a polish commit on top: scoped to `--no-docker` mode, tightened the wording.

To answer your two questions: claude doesn't run inside the container in docker mode (the runner image has its own claude install with its own env), so the host sandbox config only matters in `--no-docker`. A pre-baked sandbox profile that works with scrutineer is a fair idea but not urgent enough to block this — happy to merge once you give the polish a once-over.

connorshea:
Yeah definitely not a blocker, the changes you made look good to me!

andrew:
Thanks @connorshea!

PR #200, opened 2026-05-11
feat: implement exposure skill for assessing dependent reachability

### LINKED ISSUE #9: Exposure analysis and dependent testing
For each top-N dependent of a vulnerable package, trace whether the dependent actually reaches the vulnerable code and record an affected or not_affected statement with the right VEX justification.

- [ ] `finding_dependents` join (finding_id, dependent_id, status, justification)
- [ ] Exposure skill per top-N dependents
- [ ] Cache dependent clones
- [ ] Per-dependent table on the finding page

alexandre-daubois:
Fix #9

A few design choices I made, open to debate 🙂 

- The status column mirrors the CSAF `product_status` buckets verbatim, so the VEX export is a passthrough rather than a translation that can drift.
- Each (finding, dependent) pair is its own scan. Prompts stay short, failures are isolated, and the shared clone cache amortises the extra git work.
- Top 10 dependents per click, should we change it maybe or add a way to customize it?
- The new skill is pretty minimal, happy to improve it one way or another

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -48,7 +48,8 @@ One row per skill execution. Every scan is a skill; `kind` is always `skill`. `s
 | skill_id | integer FK | References `skills.id`. Null for legacy non-skill rows. |
 | skill_version | integer | Version of the skill at run time; the skill row's `version` bumps on every edit so older scans stay readable. |
 | skill_name | text | Denormalised skill name for UI display. |
-| finding_id | integer FK | Set when the scan is finding-scoped (verify/patch/disclose). References `findings.id`. |
+| finding_id | integer FK | Set when the scan is finding-scoped (verify/patch/disclose/exposure). References `findings.id`. |
+| dependent_id | integer FK | Set on `exposure` scans only. References `dependents.id`; identifies which downstream consumer the skill is auditing for reachability of the upstream finding. |
 | api_token | text | Per-scan bearer token that the skill presents when calling `/api`. Only valid while the scan is running. |
 | ref | text | Git ref to checkout after cloning. Empty means the default branch. |
 | sub_path | text | Scopes code analysis to a sub-folder of the clone (monorepo packages). Empty means repo root. |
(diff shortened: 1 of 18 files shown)


andrew (changes requested):
Thanks for picking this up — the shape is right (CSAF status enum stored verbatim so VEX export is a passthrough, one scan per `(finding, dependent)`, upsert on the join table). A few things I'd want fixed before merge, the first two in particular.

### Must fix

**Shared dependent-cache races under concurrent workers.** `doExposure` symlinks `{workRoot}/src` to `{DataDir}/dependent-cache/<sha(url)>`, then `RunSkill` calls `ensureClone` which operates through that symlink on the shared dir. Click "Run exposure" on two findings in the same library and the worker pool can pick up `(findingA, depX)` and `(findingB, depX)` simultaneously; both run `git fetch`/checkout in the same physical directory. Either flock the cache dir for the duration of `ensureClone`, or clone into the cache once under lock and copy into `workRoot/src` per scan. The copy approach also makes the SKILL.md "treat as read-only" claim true — right now nothing stops the model writing into the shared cache.

**Wrong VEX justification in SKILL.md step 2.** Pinning a version outside the affected range is labelled `component_not_present`. Per CSAF 2.0 that label means the component isn't in the product at all. A dependent that ships a non-vulnerable version of the library has the component present; the right label is `vulnerable_code_not_present`. This flows straight into the VEX export so downstream tooling will misread it.

**`newExposureWorker` opens `file::memory:?cache=shared`.** Existing worker tests use plain `file::memory:` (see `cna_test.go`). With `cache=shared` all five exposure tests share one DB, which is why the fixture needs `t.Name()` salting and `FirstOrCreate` on the skill row. Drop `?cache=shared` and the workarounds go with it.

**Hardcoded `/findings/1/exposure` in the three web tests.** Use `fmt.Sprintf("/findings/%d/exposure", f.ID)`; these break as soon as another test seeds a finding first.

### Should fix

- `buildFlags` ranges over the `byLabel` map so the CSAF `flags` array order is non-deterministic. Sort by label so the export is reproducible.
- Dead code: `queued` in `findingExposureRun` is incremented but never read; `_ = i` in `TestFindingExposureRun_enqueuesScanPerDependent` should be `for _, name := range`; `contains()` in `finding_csaf_test.go` is a one-line wrapper around `slices.Contains`.
- Docs: `docs/skills.md` needs an `exposure` row in the output_kind table and `dependent_id` added to the `context.json` paragraph; `docs/database.md` needs `FindingDependent` and `Scan.DependentID` (same as f873e8b did for the last schema change).
- No `allowed-tools` on the exposure skill. If you fix the cache race by copying instead of symlinking this is moot; otherwise the model can `Edit` files under `./src` and pollute the shared cache.

### Nits

- `loadFindingDependents` takes `*Server` as a param rather than being a method, and duplicates the same load-dependents-by-id lookup that `findingShow` does inline. Not blocking.
- The check-then-create in `recordSkippedExposure` / `upsertExposure` is TOCTOU but `idx_finding_dependent` covers it; fine for single-operator.

On your open questions: the hardcoded top-10 const is fine for now, a `?n=` param can come later if anyone asks; and the skill body is the right size, just the `component_not_present` label needs swapping.

alexandre-daubois:
Thanks for the review! I addressed your comments in the last commit

andrew (approved):
LGTM

PR #204, opened 2026-05-11
feat: add local frontend assets locally

### LINKED ISSUE #142: Vendor all CDN assets so the frontend works offline
All frontend dependencies are loaded from cdn.jsdelivr.net:

- `@tailwindcss/browser`
- `basecoat-css`
- `htmx.org`
- `htmx-ext-sse`
- `lucide`
- `highlight.js` (being added for syntax highlighting)

This means the UI breaks without internet access. Download each asset into `internal/web/static/vendor/` and serve them locally. Keep the CDN versions noted somewhere so they can be updated.

alexandre-daubois:
Fix #142

Using a local script downloading the whole thing + being the only source of truth for installed versions seems to be the best DX we can offer without using a package manager.


andrew (changes requested):
One thing before merge: the tailwind entry in `scripts/vendor-assets.sh` points at `@tailwindcss/browser@4` (floating major) while the filename says `4.3.0`. Re-running the script later would silently write a newer 4.x under the `4.3.0` name. Pin it to `@tailwindcss/browser@4.3.0` so the filename stays honest.

While you're there: `mkdir -p` the vendor dir before `cd`ing into it, and consider clearing `*.{js,css}` first so old versioned files don't linger after a bump. Neither blocks.

Otherwise looks good, thanks for picking this up.

alexandre-daubois:
Here you go 🙂

andrew (approved):
Thanks, all addressed.

andrew:
@alexandre-daubois somewhat inspired by this (but also i've been wanting the same for other projects too), I started working on https://github.com/git-pkgs/pin

alexandre-daubois:
Oh that's such a good idea. Keeping an eye on it!

PR #205, opened 2026-05-11
feat: Add OpenAI-compatible backend support (Ollama, etc.)

MatthewKhouzam:
- Add openai backend worker using go-openai client
- Add -backend flag to select between anthropic and openai
- Support OPENAI_API_KEY and OPENAI_BASE_URL env vars
- Document usage with local Ollama and functiongemma model in README

This code was made with heavy assistance of claude-opus-4.6

It was tested with functiongemma:270m on several SBOMs

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ When the docker runner is active, scrutineer starts an authenticated egress prox
 | `-scan-timeout` | `1h` | Wall-clock limit per scan; exceeded scans fail |
 | `-max-turns` | `0` | Passed as `--max-turns` to claude-code (0 = unlimited) |
 | `-schema-strict` | `false` | Fail a scan when its `report.json` does not validate against the skill's `schema.json` (default: warn in the scan log and parse anyway) |
+| `-backend` | `anthropic` | Backend to use: `anthropic` or `openai` |
 | `-anthropic-base-url` | - | Custom Anthropic API base URL (env: `ANTHROPIC_BASE_URL`) |
 
 ## Config file
@@ -203,6 +204,25 @@ The config file can also replace the model pick list and pin the default model:
         id:   claude-sonnet-4-6
       - name: Opus
         id:   claude-opus-4-7
+
+## OpenAI-compatible backend (Ollama, etc.)
+
+Scrutineer can use any OpenAI-compatible API instead of Anthropic. This is useful for running against local models via [Ollama](https://ollama.com). You need a model that supports tool calling (e.g. `functiongemma`):
+
+    ollama pull functiongemma:270m
+    export OPENAI_API_KEY=unused
+    export OPENAI_BASE_URL=http://localhost:11434/v1
+    go run ./cmd/scrutineer -skills ./skills -backend openai
+
+Or configure it in `scrutineer.yaml`:
+
+    backend: openai
+    default_model: functiongemma:270m
+    models:
+      - name: Function Gemma 3
+        id: functiongemma:270m
+
+The `OPENAI_API_KEY` env var is required by the client library but can be set to any value when talking to a local server that doesn't check it.
 
 ## Security
 
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -54,6 +54,7 @@ type flags struct {
 	addr             string
 	dataDir          string
 	effort           string
+	backend          string
 	noDocker         bool
 	runnerImage      string
 	skillsRepo       string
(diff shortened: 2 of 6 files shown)


MatthewKhouzam:
image
Here is a screenshot of it in action.

andrew:
@MatthewKhouzam we're going to need good test coverage to ensure that future changes don't break non-claude users

MatthewKhouzam:
Wow, fast review, thanks! I will add some more coverage. For now I released this patch as a "let's talk about it" patch. I did not yet experience a real vulnerability, so I don't even know how claude behaves when things go wrong.

andrew:
@MatthewKhouzam I think we'd be better off switching between coding harnesses, rather than raw model calls, because all these security coding agents have been fine tuned against their harnesses, codex and opencode would be good to start with

MatthewKhouzam:
> @MatthewKhouzam I think we'd be better off switching between coding harnesses, rather than raw model calls, because all these security coding agents have been fine tuned against their harnesses, codex and opencode would be good to start with

That's probably a good direction. I will look into that, either way, we need to decouple from Claude which is done here. I will look into it. Also, and I need to emphasize this. I am OK with this code not getting in. I want to see what's needed to decouple this tool from claude-code so that we can have zero leaks, in case we want to run it in enterprise environments. I do not care strongly about ollama/vllm right now, I want to make it more "adoptable". You can also put giant banners on ollama saying "this is for experimentation purposes only!" or something like that. 

Again thanks! Sorry I missed the message before pushing the change.

MatthewKhouzam:
My personal motivation for this patch is to minimize data exfiltration chances. So Anthropic (even if I trust them) is one. By replacing it with the localhost openai api, we have 1/3 risks removed. the other two I see are Docker (which we can handle) and ecosyste.ms which seems benign by design (package names only).

andrew:
Thanks Matthew. Closing thought before we decide on this one: for the Anthropic-but-not-Anthropic-SaaS case, `claude` already honours `ANTHROPIC_BASE_URL`, so pointing scrutineer at Bedrock, Vertex, or a self-hosted gateway works today without code changes — that covers a lot of the enterprise-adoption goal.

For genuinely non-Anthropic stacks, the better lever is swapping the harness (codex, opencode) rather than the model call, since the agents are fine-tuned against their harnesses. Opened https://github.com/alpha-omega-security/scrutineer/issues/211 for that.

Separately, on data minimisation: the `packages.ecosyste.ms` enrichment is opportunistic — it adds context but the scan still works without it. Tracked at https://github.com/alpha-omega-security/scrutineer/issues/212 for a disable flag and graceful degradation when the domain is blocked.

N0tre3l:
hi @MatthewKhouzam 
Thanks for putting this together! The abstraction looks clean and the test coverage is a great addition.

While reviewing the new `openai.go` implementation, I noticed a few technical details that might need some adjustment before this gets merged:

**1. Sandbox bypass with OpenAI backend**
When `backend=openai` is selected, `selectRunner` returns before starting the Docker runner or the Egress Proxy. This means that tools like `run_command` will execute directly on the host, and `web_fetch` can reach any internal network endpoints. This might give users a false sense of security, as choosing OpenAI (to avoid data exfiltration) currently results in a less isolated environment than the default runner.

**2. Path traversal vulnerability in workspace checks**
In functions like `toolReadFile` and `toolWriteFile`, the containment check uses `strings.HasPrefix`:
`if !strings.HasPrefix(full, workRoot)`
This can be bypassed if a path resolves to a directory that shares the same prefix string (e.g., if `workRoot` is `/tmp/scan-1`, a path resolving to `/tmp/scan-10/evil` will pass). A safer check would be:
`if full != workRoot && !strings.HasPrefix(full, workRoot+string(os.PathSeparator))`

**3. Uncapped response body in `callAPI`**
In `callAPI`, the response body is read entirely into memory with `io.ReadAll(resp.Body)`. A malicious or misconfigured endpoint could potentially exhaust worker memory. It might be safer to use `io.LimitReader(resp.Body, maxResponseBody)` to match the rest of the package.

**4. Turn counter accuracy**
In the main execution loop, the event emitter reports the configured maximum turns rather than the actual turns taken:
`emit(Event{..., Turns: maxTurns, ...})`
This could make per-scan cost and usage tracking inaccurate for scans that complete early.

Hope this helps! Let me know if you'd like any help adjusting these.

MatthewKhouzam:
I see one major issue that I was using the openai api for ollama, and neglecting that openai is also compatible with it. I will add a harness for OpenAI too. I cannot test it as I have no openai key.. yet.. I will look into the feedback, thank you @andrew and @N0tre3l .

andrew:
Closing in favour of #239 which carries this forward with codex and opencode. Thanks for kicking off the discussion, the enterprise-egress framing was useful and led to #211 and #212.

MatthewKhouzam:
Thanks, this is the ideal result for me. I wanted to make sure you can handle this use case in particular.

PR #213, opened 2026-05-13
fix(db): safely strip line numbers to prevent Windows path truncation

N0tre3l:
### Summary
The `normaliseLocation` function in `internal/db/fingerprint.go` naively split the location string on the first `:` to strip line and column numbers. However, when processing findings with Windows absolute paths (e.g., `C:\project\src\main.go:42`), this logic incorrectly truncates the entire path, leaving only the drive letter (`"c"`).

This causes massive false-positive deduplication collisions for findings generated on Windows, as they all map to the exact same file location.

### Changes
- Updated `normaliseLocation` to iterate backwards and only strip the trailing segment if it is purely numeric.
- This safely removes `:line:col` or `:line` suffixes without truncating drive letters.
- Added test cases in `fingerprint_test.go` to explicitly cover Windows absolute paths.

### DIFF
--- a/internal/db/fingerprint.go
+++ b/internal/db/fingerprint.go
@@ -38,8 +38,23 @@ func FingerprintFinding(skillName, subPath, cwe, location, title string) string 
 func normaliseLocation(loc string) string {
 	loc = strings.TrimSpace(loc)
 	loc = strings.TrimPrefix(loc, "./")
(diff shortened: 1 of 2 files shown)


andrew (approved):

andrew:
Thanks!

PR #193, opened 2026-05-08
findings: group same-fingerprint matches and carry full location set

andrew:
Closes #191.

Adds `db.Finding.Locations` (newline-joined `file:line` set) with `LocationList()` and `ExtraLocationCount()` helpers; GORM AutoMigrate adds the column on startup.

The findings parser now reads an optional `locations[]` array from `report.json`, and the within-scan dedupe is replaced by a `groupByFingerprint` pass that runs before the DB loop: entries that share a fingerprint accumulate their match positions into one row instead of silently dropping all but the first. On re-observation the existing row's `location` and `locations` are refreshed so the set tracks current HEAD.

`semgrep/scripts/scan.py` groups results by `(check_id, message)` and `zizmor/scripts/scan.py` by `(ident, desc)`, each emitting `location` (first sorted) plus the full `locations[]`. The rule's message template interpolates whatever metavars the rule author considered significant, so identical messages are the rule author's signal that the matches are equivalent; different messages stay as separate findings. Both schemas gain the array.

The finding page shows the primary location as before, then a collapsed "N more locations" block with each entry linked through the existing forge-URL helper. Findings, scan, and repo tables show a `+N` badge next to the primary location.

For the issue's example, 102 `var-in-href` hits with one message become one finding with a `+101` badge and an expandable list. The fingerprint function is unchanged so existing rows keep their identity across the upgrade; pre-existing rows have an empty `Locations` column until the next rescan populates it.

Not done here: the AST-normalisation grouping key the issue suggests for splitting same-message matches by metavar context. The `(check_id, message)` key is the practical signature semgrep already gives us; a per-language normaliser can layer on top of the `locations[]` plumbing later.

PR #214, opened 2026-05-13
web: export individual scan results as markdown

### LINKED ISSUE #194: Export individual scan results as markdown
## Summary

Each scan's results page (e.g. `/scans/18`) currently shows the skill's output as JSON. There's no way to grab *just this scan*'s findings as markdown for sharing, archiving, or pasting into an email or another issue. The existing markdown exports work at higher granularities — `/repositories/{id}/report.md` covers the whole repo across all scans, `/orgs/{login}/findings.md` covers an org — but neither isolates a single scan.

## Example

A semgrep run we did against `apache/tooling-trusted-releases` produced scan #18: 120 findings, of which 119 are medium-severity and 102 are the same `generic.html-templates.security.var-in-href` rule firing across many template files. As JSON it's 120 entries of near-identical structure to scroll through. As markdown — a table grouped by rule, or one section per rule with a list of file:line locations — it would be skimmable in a few seconds and pasteable into a triage thread.

The use case is "here, look at scan #18's results" without having to either share a localhost URL or paste raw JSON.

## Suggested behavior

Add an "Export as markdown" button on the scan detail page that downloads the scan's findings as `.md`. Concrete shape:

- New route `GET /scans/{id}/report.md` that reads the scan's `Report` JSON and dispatches on `output_kind` (`findings`, `maintainers`, `packages`, `freeform`, etc.) to a per-kind renderer
- A `findings`-kind renderer that groups by rule so the 102 `var-in-href` entries collapse into one section with a file:line list rather than 102 separate rows
- For `output_kind: freeform`, the markdown is the raw report wrapped with metadata (skill name, commit, started, duration, cost)
- A button in the scan template alongside the existing scan controls, mirroring the existing pattern in `repo_show.html`: `<a href="/scans/{{.Scan.ID}}/report.md" class="btn-outline" download>...</a>`

The route shape matches the existing `/repositories/{id}/report.md` and `/orgs/{login}/findings.md` patterns, which keeps URL structure consistent. The renderer would naturally live in `internal/web/scan_report.go` next to the existing per-repo and per-org report files.

**Suggested labels:** `enhancement`

andrewmusselman:
Closes #194.

The scan detail page at /scans/{id} shows the raw report JSON in the UI but provides no way to grab just that scan's results for sharing or archiving. The existing markdown exports work at coarser granularities — /repositories/{id}/report.md is repo-wide, /orgs/{login}/findings.md is org-wide — and neither isolates a single scan.

Add GET /scans/{id}/report.md mirroring the existing per-repo and per-org report patterns. The handler:

- Loads the scan and its skill, falling back gracefully if the skill row is gone (deleted skill, scan retained for audit).
- Writes a metadata header (repo, skill, status, model, commit, started, finished, duration, turns, cost, findings count, error if any).
- For findings-kind scans, queries findings with scan_id = the scan ID (matching what the /scans/{id} UI shows via scan.Findings preload), so the markdown stays in lockstep with the page. Each finding is rendered by the existing writeReportFinding helper in repo_report.go, so the scan report and the repo report describe the same finding the same way.
- For other output kinds, parses the report.json and walks the tree: top-level scalars become a metadata table, top-level arrays of objects each become their own section with a table whose columns are picked from the most-common scalar keys (preferring name, version, license, severity, cwe, etc.). CycloneDX SBOMs are recognised by their bomFormat marker and given a tuned heading.
- Falls back to a fenced JSON block only for shapes the walker can't tabulate (non-object top level, nested-only top level). Uses prettyJSON from jsontree.go for the indenting.

Leaf scalar formatting (numbers, booleans, null) delegates to scalarString in jsontree.go, keeping the HTML Data tab and the markdown export consistent.

Filename: scrutineer-<repo>-scan-<id>-<skill>-<date>.md so a directory of downloaded reports stays unambiguous without renaming.

UI: a download button next to the existing Retry control on /scans/{id}, gated on Scan.HasExportableReport so scans with empty results (e.g. a clean subprojects scan emitting {"subprojects": []}) don't offer a download that wouldn't be worth the click. The route itself is unaffected; the URL still works if hit directly.

Tests cover the findings dispatch, the freeform-table rendering for non-findings kinds, the CycloneDX heading detection, the genuine raw-JSON fallback path, the 404 path, and the HasExportableReport threshold logic.

## Testing
```
$ go test ./internal/web/... ./internal/db/...
ok  	scrutineer/internal/web	(cached)
ok  	scrutineer/internal/db	(cached)
```


andrewmusselman:
I'll fix the failing lints @andrew

andrew (commented):
Looks good — the freeform JSON→table rendering is a nice upgrade over what #194 originally asked for, and reusing `writeReportFinding` keeps the scan and repo reports consistent.

One question: #194's motivating case was the 102 `var-in-href` findings collapsing into one section with a file:line list. As written, a 120-finding semgrep scan renders 120 full `writeReportFinding` sections (table + trace + boundary + …). Is grouping-by-rule planned as a follow-up?

andrewmusselman:
> Looks good — the freeform JSON→table rendering is a nice upgrade over what #194 originally asked for, and reusing `writeReportFinding` keeps the scan and repo reports consistent.
> 
> One question: #194's motivating case was the 102 `var-in-href` findings collapsing into one section with a file:line list. As written, a 120-finding semgrep scan renders 120 full `writeReportFinding` sections (table + trace + boundary + …). Is grouping-by-rule planned as a follow-up?

Good catch @andrew, that was a real gap, not just the legacy-scan case you were pointing at. Even post-#193 grouped findings were only rendering the primary location; pushed an amend so writeReportFinding lists the rest under "Additional locations" (sorted naturally so :110 doesn't sort before :56). Also caught that the query was using scan_id instead of last_seen_scan_id, which made re-observed scans render "No findings" — fixed. scan_show.html has the same query bug, can file a follow-up.

andrew:
Thanks!

andrew (approved):

PR #215, opened 2026-05-13
feat: add support for skill repo ref

### LINKED ISSUE #69: Pin -skills-repo to a ref for reproducible scans
`-skills-repo https://github.com/org/skills` clones HEAD on startup, so two scans a week apart can run different prompts with no record of which version was used.

Accept `owner/repo@ref` (sha, tag, or branch) and record the resolved sha on each `Scan` row.

- [ ] Parse `@ref` suffix in `internal/skills/git.go`, default to default branch when absent
- [ ] Resolve and store `SkillsRepoSHA string` on `Scan`
- [ ] Show the sha on the scan detail page and in the markdown export
- [ ] Same for the config file: `skills_repo: org/skills@v0.3.1`

Related to #56.

alexandre-daubois:
Fix #69

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -52,6 +52,7 @@ One row per skill execution or external import. `skill_name` / `skill_version` p
 | dependent_id | integer FK | Set on `exposure` scans only. References `dependents.id`; identifies which downstream consumer the skill is auditing for reachability of the upstream finding. |
 | api_token | text | Per-scan bearer token that the skill presents when calling `/api`. Only valid while the scan is running. |
 | ref | text | Git ref to checkout after cloning. Empty means the default branch. |
+| skills_repo_sha | text | Commit of `-skills-repo` resolved at startup and stamped on every skill scan. Empty when `-skills-repo` is unset or for `import` scans. |
 | sub_path | text | Scopes code analysis to a sub-folder of the clone (monorepo packages). Empty means repo root. |
 | commit | text | Git HEAD at scan time. |
 | started_at | datetime | |
--- a/internal/skills/git.go
+++ b/internal/skills/git.go
@@ -11,16 +11,63 @@ import (
 
 const dirPerm = 0o755
 
-// CloneOrPull prepares a local copy of a git repo at dst. On first call it
-// clones; on subsequent calls it fetches and resets to origin/HEAD so skill
-// updates propagate without needing to wipe the cache. fullClone toggles
-// between --depth 1 and full history, and unshallows an existing shallow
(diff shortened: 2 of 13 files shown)


internal/skills/git.go
```diff
@@ -11,16 +11,59 @@ import (
...
+// ref is empty when none is given, meaning "use the repo's default branch".
+func ParseRepoSpec(raw string) (url, ref string, err error) {
+	raw = strings.TrimSpace(raw)
+	if raw == "" {
+		return "", "", fmt.Errorf("empty skills_repo spec")
+	}
+	if i := strings.Index(raw, "://"); i >= 0 {
+		scheme := raw[:i+3]
+		rest := raw[len(scheme):]
+		if at := strings.LastIndex(rest, "@"); at >= 0 {
```
andrew: This breaks `https://<token>@github.com/org/repo` (no ref), which previously worked for private skills repos — `LastIndex` turns everything after the `@` into the ref. Could you only split when the `@` appears after the last `/`, or note in the docs that private repos need a credential helper instead of token-in-URL?

docs/database.md
```diff
@@ -52,6 +52,7 @@ One row per skill execution or external import. `skill_name` / `skill_version` p
 | dependent_id | integer FK | Set on `exposure` scans only. References `dependents.id`; identifies which downstream consumer the skill is auditing for reachability of the upstream finding. |
 | api_token | text | Per-scan bearer token that the skill presents when calling `/api`. Only valid while the scan is running. |
 | ref | text | Git ref to checkout after cloning. Empty means the default branch. |
+| skills_repo_sha | text | Commit of `-skills-repo` resolved at startup and stamped on every scan that ran a remote skill. Empty when `-skills-repo` is unset or for `import` scans. |
```
andrew: `enqueueSkillWith` stamps this on every skill scan regardless of the skill's source (`remote`/`local`/`ui`), not just remote ones. That behaviour seems fine, just tweak the wording here to match.

andrew (commented):
Nice work, this covers everything in #69 and the git integration tests are appreciated. Two small things below.

docs/database.md
alexandre-daubois: Fixed 👍

internal/skills/git.go
alexandre-daubois: Added support with your first suggestion. The ref must be a single segment, not something like `refs/heads/main`. I think it's fine like this, most people will use SHA or tags anyway

andrew (approved):

andrew:
Thanks!

PR #216, opened 2026-05-13
fix: stay on the "Findings" tab when applying a CWE category filter

alexandre-daubois:
Without this change, applying a CWE filter gets the user back to the summary tab, even though the list is indeed filtered when getting back on the right tab.

### DIFF
--- a/internal/web/templates/org_show.html
+++ b/internal/web/templates/org_show.html
@@ -78,12 +78,12 @@
         </button>
         <div data-popover aria-hidden="true" data-align="end">
           <div role="menu">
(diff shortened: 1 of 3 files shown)


andrew (approved):
Nice catch. The severity and sort dropdowns in the same panel on `sbom_show.html` (around lines 122 and 157) have the same issue, mind adding `#st2` to those too while you're here?

alexandre-daubois:
Did it in #217 but I can merge both into on PR if you prefer 🙂

andrew:
Ah sorry, my fault, was reviewing them in reverse!

PR #218, opened 2026-05-13
fix: put the Settings tooltip at bottom of the settings icon

alexandre-daubois:
Otherwise it is cut off by the screen


andrew (approved):

andrew:
Thanks, i noticed that i totally forgot to fix it!

PR #226, opened 2026-05-14
fix: Update zizmor repo link.

connorshea:
The zizmor repo moved and the old URL redirects, so link it directly here.

https://github.com/zizmorcore/zizmor


andrew (approved):

andrew:
Good catch

PR #229, opened 2026-05-14
fix: Stage schema.json at the workspace root

### LINKED ISSUE #221: Maintainers report fails to parse/unmarshal in some cases
I've seen this error a few times from the maintainers job, though only sometimes: `parse maintainers report: json: cannot unmarshal array into Go struct field .maintainers.evidence of type string`

```
[error] schema: report.json does not validate against schema.json:
/maintainers/0: validation failed
/maintainers/0: missing properties 'login', 'status'
/maintainers/0/role: value must be one of 'lead', 'maintainer', 'contributor', 'bot'
/maintainers/0/evidence: got array, want string
/maintainers/0: additional properties 'activity_level', 'is_rubygems_owner', 'github_username', 'website', 'location', 'last_commit_date', 'emails', 'company' not allowed
/: additional properties 'repository', 'notable_contributors', 'bots', 'security_contact', 'analysis' not allowed
[error] parse maintainers report: json: cannot unmarshal array into Go struct field .maintainers.evidence of type string
```

In this specific case I was running the tool on https://github.com/Casecommons/pg_search, though I'm unsure if that is reproducible per-repo.

andrew:
Every SKILL.md tells the model to read `./schema.json`, but `stageSkill` only wrote it to `.claude/skills/{name}/schema.json`. The model would try `./schema.json`, get file-not-found, then glob around for it. In `--no-docker` runs that glob falls through to the scrutineer source tree so it always recovered, which is why this never showed up locally; in a docker workspace there is no fallback and the hunt sometimes gave up, at which point the model invented its own report shape (`github_username`, `is_rubygems_owner`, `evidence` as an array, top-level `analysis` etc) and `parseMaintainersOutput` choked on it.

Now `stageSkill` also writes `schema.json` to the work root alongside `context.json` and `report.json`, so the `./schema.json` path every skill references resolves on the first read. Reordered `stageContext` before `stageSkill` in `doSkill` since `stageContext` is what creates the work root.

Fixes #221.

### DIFF
--- a/internal/worker/exposure.go
+++ b/internal/worker/exposure.go
@@ -159,7 +159,7 @@ func (w *Worker) doExposure(ctx context.Context, scan *db.Scan, emit func(Event)
 	scan.Commit = cacheCommit
 
 	skillDir := filepath.Join(workRoot, ".claude", "skills", skill.Name)
-	if err := stageSkill(&skill, skillDir); err != nil {
+	if err := stageSkill(&skill, workRoot, skillDir); err != nil {
 		return "", fmt.Errorf("stage skill: %w", err)
 	}
 	if err := stageContext(workRoot, w.APIBase, w.ForkOrg, scan, &scan.Repository); err != nil {
--- a/internal/worker/skill.go
+++ b/internal/worker/skill.go
@@ -90,11 +90,11 @@ func (w *Worker) doSkill(ctx context.Context, scan *db.Scan, emit func(Event)) (
 		return "", err
 	}
 	skillDir := filepath.Join(workRoot, ".claude", "skills", skill.Name)
-	if err := stageSkill(&skill, skillDir); err != nil {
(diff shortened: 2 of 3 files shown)


N0tre3l:
@andrew  
This is honestly one of the funniest and smartest bugs I've seen! The fact that the AI was silently covering up the issue by globbing the local source tree is hilarious.

Great catch on finding this silent failure inside the Docker container. Reordering stageContext before stageSkill is a neat and logical fix. Thanks for the detailed write-up!

connorshea:
Is there a reason not to also copy the relevant python scripts for a given skill in the same way? I noticed the sbom skill gets caught up trying to find git-pkgs and it seems like it may be fixable in the same way as this PR.

andrew:
@connorshea can you open a separate issue for this?

PR #232, opened 2026-05-15
feat: add support for code display in browser

### LINKED ISSUE #15: Keep repos locally and add a code browser
Keep clones persistent across scans and add a simple in-UI code browser so findings link to locally rendered source instead of bouncing out to the forge.

alexandre-daubois:
Fix #15

Some additional JS code is required, otherwise strings inside backticks shift the whole thing in the UI:

<details>
<summary>Why some additional JS is required</summary>

image
</details>

Another solution could be to use Chroma and do the rendering/highlighting on the server-side, but that would be an additional dependency and may not be worth it. Maybe later if there are plans to go further with syntax highlighting.

No tree view for now, just the code snippet with an highlighted line if provided:

https://github.com/user-attachments/assets/e6a33a9b-af21-4150-8eb9-8494da9a4779

### DIFF
--- a/docs/skills.md
+++ b/docs/skills.md
@@ -115,13 +115,15 @@ If you need an output shape that is not in this list, see "When you need Go chan
 
 When a scan starts, the worker creates `./data/work/scan-{id}/` with:
 
-    ./src/                       clone of the target repository at the requested ref
+    ./src/                       working copy of the target repository at the requested ref
     ./context.json               who you are scanning and how to call scrutineer back
     ./.claude/skills/{name}/     this skill's SKILL.md, schema.json, and any aux files
     ./schema.json                copy of the skill's schema for the model to read
     ./report.json                the skill writes its output here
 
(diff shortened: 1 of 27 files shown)


andrew:
Needs a rebase

andrew (changes requested):
A few things before this can land:

- `code_browser.html` has ~60 lines of inline `<script>` for the line-wrapping and `balanceSpansAtNewlines` logic. We just merged a CSP with `script-src 'self'` (#240) so this will be blocked at runtime now. Move it into `static/app.js` (or a new `static/code_browser.js`).
- There's a fair amount of unrelated comment trimming in `db.go`, `repo_report.go`, `scan_report.go`, `scan_report_test.go` and a few test files. Could you split those into a separate PR? Makes this one much easier to review.
- `prepareRepoSrc` now always passes `fullClone=true` to `ensureClone`. I see why (need history for `git show <old-sha>`), but it means every scan keeps full history on disk where shallow used to do. Is that intentional for all scans, or should the full clone be deferred until the browser actually requests an old commit?

The path-traversal handling and tests look solid otherwise.

alexandre-daubois:
Comment addressed 🙂

andrew (approved):
Thanks, all three addressed. The on-demand unshallow in  is a nice solution.

andrew (approved):
Thanks, all three addressed. The on-demand unshallow in `EnsureCommit` is a nice solution.

PR #233, opened 2026-05-15
Security: fix SSRF in dependency resolution (T4) AND Prompt Injection (T5)

N0tre3l:
I noticed that the T4 SSRF threat (from `threatmodel.md`) still had some pending mitigations, so I went ahead and implemented them.

- Added `safeClient` with a `CheckRedirect` hook in `internal/worker/metadata.go` to block any redirects pointing to RFC1918/loopback IPs.
- Updated `depScan` and `dependentScan` in `internal/web/server.go` to validate the resolved repo URLs against `worker.DefaultEgressAllow` before they get queued.

This covers the remaining items for T4. Let me know if anything needs changing!

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -378,6 +378,9 @@ type Finding struct {
 	// scrutineer produced itself. Used as the skill-name input to the
 	// fingerprint so re-importing the same external report dedupes.
 	ImportedFrom string `gorm:"index"`
+	// SourceJob records the name of the tool or model that generated this finding.
+	// Used for provenance tagging (T5) and to conditionally render AI caveats.
+	SourceJob string `gorm:"index"`
 
 	// Disclosure / triage fields. Any of these may be set by a tool, a
 	// model-backed skill, or the analyst; see FindingHistory for the trail.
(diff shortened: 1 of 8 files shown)


N0tre3l:
Also, included a fix for the open T5 (Prompt Injection) issue from threatmodel.md. Added a SourceJob column to the Finding schema for provenance tracking, updated the parsers to capture the tool name, and made the UI caveat conditional so it only warns on 'claude' findings instead of hardcoding it for everything.

andrew:
Closing this one rather than iterating, since the T4 hunks are duplicated in #236 and the T5 change regresses the warning it's meant to strengthen.

On T5: `SourceJob` is set to `scan.SkillName` (`security-deep-dive`, `triage`, etc.), so the template's `eq .SourceJob "claude"` never matches and every model finding drops from "verify it independently" to the softer "review the evidence". The existing `ImportedFrom` field already distinguishes tool-imported from model-generated findings; if you want to redo the provenance UI, branching on `ImportedFrom == ""` in the template gets you there without a new column.

On T4: same `DefaultEgressAllow` issue as noted on #236. Happy to look at a standalone T4 PR that uses a proper forge list and respects `cfg.EgressAllow`.

PR #236, opened 2026-05-16
worker: implement missing safeURL validation for html_url and icon_url (T7)

N0tre3l:
The threat model marks T7 as mitigated via safeURL() validation, but the 
function was never implemented. Both html_url and icon_url were stored 
verbatim from the metadata skill response, meaning a javascript: or data: 
URI would be written to the database and rendered as a link in the UI — 
a stored XSS vector against the operator.

This PR adds safeURL() in clone.go and applies it in 
parseRepoMetadataOutput() so only http/https URIs are accepted before 
the UPDATE reaches the database. All existing tests pass.

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -696,6 +696,11 @@ func (s *Server) depScan(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "could not resolve repository URL for "+dep.Name, http.StatusUnprocessableEntity)
 		return
 	}
+	u, err := url.Parse(repoURL)
+	if err != nil || !worker.HostAllowed(worker.DefaultEgressAllow, u.Hostname()) {
+		http.Error(w, "resolved repository URL is not on the forge allowlist", http.StatusForbidden)
+		return
+	}
 	s.addRepoAndScan(w, r, repoURL)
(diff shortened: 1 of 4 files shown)


andrew (changes requested):
The `safeURL` helper in `clone.go` and its use in `skill_parsers.go` are right and close the gap `threatmodel.md:168` already claims is done. Could you cut this PR down to those two files plus a small table test for `safeURL` (http/https pass, `javascript:`/`data:`/empty/whitespace fail)?

The `server.go` and `metadata.go` hunks are T4, not T7, and they're duplicated verbatim in #233 so merging both would conflict. They also have problems as written: `DefaultEgressAllow` is the egress-proxy list (it contains `*.anthropic.com`, `host.docker.internal`, registries) so it's the wrong filter for "is this a forge", and it ignores the operator-configured `egress_allow` extras assembled in `cmd/scrutineer/main.go:264`. Better handled in a focused T4 PR.

andrew:
Superseded by #407, which carries the same `safeURL` helper and call-site changes without the T4 SSRF hunks. Thanks for the original work — the commit credits you as co-author.

Issue #211, opened 2026-05-12
Support codex and opencode as alternative coding harnesses


andrew:
Scrutineer currently drives scans through `claude` (the Claude Code CLI). For users who can't or don't want to route data through Anthropic — typically enterprises with a specific vendor relationship, or operators who want to keep model traffic on a non-Anthropic stack — there's no path today that preserves the agent harness.

Raw model API calls (see #205) aren't a substitute: security coding agents are fine-tuned against their harnesses, and dropping to a bare chat-completions loop loses most of what makes them work on this task.

The reasonable shape: a `--harness` flag (or per-worker config) that can select `claude`, `codex`, or `opencode`, with the runner image carrying whichever binary is selected. Each harness has its own auth/config story, so the operator brings their own credentials and endpoint.

For the Anthropic case specifically, `ANTHROPIC_BASE_URL` already lets `claude` point at Bedrock, Vertex, or a self-hosted gateway, so that path doesn't need new code — it needs docs. This issue is about the harness-swap case, where the model isn't Anthropic at all.

Out of scope: raw OpenAI / Ollama clients (see #205 discussion). Revisit if a model+harness combo emerges that handles the workload usefully offline.

andrew:
Status: the `Harness` interface is on main (#529, #530, #532, #533, #534), covering binary, argv, stream parsing, skill staging path, project-memory filename, egress hosts, container env, state-dir env, and account-error detection. `ContainerRunner` goes through it for everything that varies per agent CLI; the container, egress proxy, language profiles and workspace stay the same.

Two implementation PRs are up:

- #535 `-backend codex`: `CodexHarness`, `HarnessByName` registry, `-backend` flag + `backend:` config key, codex binary in `Dockerfile.runner`, `docs/codex.md`
- #536 `-backend opencode` (stacked on #535): `OpencodeHarness`, opencode binary, `docs/opencode.md`

Both stream parsers are built from reading the upstream source rather than capturing a live run, so they're permissive (unknown event shapes pass through as text). I'll do an end-to-end scan with each harness in the morning and tighten the parsers against real output before merging.

One gap to close on #535 before merge: `db.Scan` doesn't record which backend produced it, only `Model`/`Effort`/`Profile`. Without that a codex-run scan retried after switching backends would try to resume a codex thread id under claude. Needs a `Backend` column on `db.Scan`, stamped at run time and threaded through the retry paths the same way `Profile` and `ScanGroup` are.

[...]

PR #239, opened 2026-05-19
Opencode support

MatthewKhouzam:
This addressed the out of scope but really interesting suggestions from #205


andrew (changes requested):
This is heading the right way but the harness choice is wired at the wrong layer. `selectRunner` returns `CodexRunner`/`OpencodeRunner`/`OpenAIRunner` *instead of* `DockerRunner`, so picking any non-claude backend drops the container and the egress proxy entirely, and `CodexRunner` additionally sets `SandboxModeDangerFullAccess` + `ApprovalPolicyNever`. For a tool that scans untrusted repos that's a step backwards from the default.

What #211 is after is the harness varying *inside* the container: `DockerRunner` stays the only runner, gains a `Harness string` field, and the image execs `claude -p` / `codex exec` / `opencode run` accordingly. The runner image picks up the extra binaries. That keeps isolation and egress filtering intact regardless of backend, and the diff gets a lot smaller.

With that shape, `openai.go` can go (raw-model loop is out of scope per #211), and the SDK deps aren't needed since the container shells out to the CLIs. Also `-backend` flag help still says "claude-code (default) or openai".

MatthewKhouzam:
Hi @andrew I added a gemini branch and codex branch to scrutineer on my personal GH. This is to show the end goal too. I applied opencode changes. Sad about openai api not making the cut, but I get it. 
  https://github.com/MatthewKhouzam/scrutineer

andrew:
Thanks, the refactor is where it needs to be — harness on `DockerRunner`, no SDK deps, non-claude + `--no-docker` errors out. The `runCommand` extraction is a nice cleanup on its own.

A few things still outstanding:

- README is stale: it documents `-backend openai` (which `ValidateBackend` now rejects) and still describes the SDK/JSON-RPC approach and host-side `npm install -g @openai/codex`. The harness runs inside the container now, so those sections need rewriting.
- `Dockerfile.runner` doesn't bundle `codex` or `opencode`, so the container will hit "executable not found".
- The egress allowlist doesn't include `api.openai.com` (or whatever opencode dials), so the proxy will 403 the harness's own API calls.
- `runCommand` pipes everything through `ParseStream`, which is claude-JSON-specific; codex/opencode output won't parse into events. Not fatal (report.json still lands) but the scan log will be noise.

This also now conflicts pretty heavily with #274, which extracted the same block of `main.go` into `setupRunner()` and added hardened-mode fields to `DockerRunner` — that rebase is going to be more than mechanical.

Given that, would you be up for splitting the `runCommand` helper (`internal/worker/run.go` + the `claude.go`/`docker.go` callers) into its own PR? That part is independently useful and would merge straight away, and it'd shrink this one to just the harness switch + image + egress once rebased.

edorian:
Small addition: 

We now have a PROFILE.md for custom runner containers.

https://github.com/alpha-omega-security/scrutineer/blob/5664bff96d44ecbdb6f7c7a58b4931e25967f03d/internal/worker/docker.go#L309

For OpenCode (and bascially everything that isn't the claude-code cli) should should be copied into AGENTS.md instead of CLAUDE.md

andrew:
I'm reworking this and adding both codex and opencode harness support over here now: https://github.com/alpha-omega-security/scrutineer/issues/211

PR #242, opened 2026-05-20
feat: add support for local directory scans

### LINKED ISSUE #130: Support scanning local directories
Currently scrutineer only accepts git repository URLs. It would be useful to point it at a local directory on disk, for example to scan a project before pushing or to audit something not hosted on a git forge.

This would need a way to provide a local path instead of a URL, skip the clone step, and run the skill directly against the directory.

alexandre-daubois:
Fix #130


andrew (changes requested):
Needs a rebase after the recent merges. Two small things while you're in there:

- `internal/web/server_test.go`: the assertion on `class="alert-destructive` couples the test to CSS. Assert on the rendered error text (the Title string) instead.
- `internal/worker/clone.go` `prepareLocalSrc`: if the supplied path is itself a symlink to a directory, `os.Stat` passes but `filepath.Walk` (lstat-based) recreates it as a single dangling link, so the container sees an empty `src`. An `os.EvalSymlinks` on the root before walking covers it.

Otherwise this looks good, happy to merge once those land.

alexandre-daubois:
Here you go!

andrew (approved):
Thanks, all addressed.

alexandre-daubois:
(Rebasing...)

PR #248, opened 2026-05-20
fix: normalize ecosystem names for packages and dependencies

connorshea:
Previously I was getting a lot of empty strings for the ecosystem value in the packages table, this should improve the situation, especially for the most common ecosystems.

Previously I'd get things like this in the Packages view:

Screenshot 2026-05-20 at 3 21 19 PM

Now it'll do a better job normalizing the ecosystem name it finds.

This change was generated using Claude Code, reviewed and tested by me. I looked in the SQLite database and confirmed that the `ecosystem` value for rubygems gems was saved as `gem` for whatever reason originally, in the dependencies table, and just an empty string in the `packages` table.

This is maybe something that should be handled/normalized elsewhere (in git-pkgs maybe?), but I'm not totally sure. This PR can be closed if this solution doesn't fit in correctly.

### DIFF
--- a/internal/worker/ecosystem.go
+++ b/internal/worker/ecosystem.go
@@ -1,0 +1,117 @@
+package worker
+
+import "strings"
+
+// normalizeEcosystem canonicalises the ecosystem string for a package row.
+// Upstream sources (packages.ecosyste.ms, git-pkgs manifest scans) are
+// inconsistent: the same logical ecosystem can arrive as "rubygems" or "gem",
+// "go" or "golang", and frequently as an empty string when the upstream
+// record is incomplete. The Packages tab displays this value directly and
+// builds its filter dropdown from the distinct set, so any variance becomes
+// visible as either empty badges or duplicate filter entries.
+//
+// Resolution order: the explicit field, then the PURL type segment, then a
+// registry-host fallback. Synonyms collapse onto the canonical name that the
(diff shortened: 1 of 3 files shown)


andrew (changes requested):
The empty-ecosystem case is real but a normalizer adds a third place where ecosystem names get mapped (this, `db.dependencyEcosystemAlias`, and `git-pkgs/purl` upstream). PURL is already the canonical identifier and both ecosyste.ms and git-pkgs emit one, so I'd rather derive ecosystem from the PURL type than reconcile strings.

Concretely: import `github.com/git-pkgs/purl`, and in `skill_parsers.go` set `Ecosystem: purl.Parse(p.PURL).Type` (falling back to `purl.EcosystemToPURLType(p.Ecosystem)` if the PURL is absent). Same in `DependencyFindings` — key the join on PURL type+name and `dependencyEcosystemAlias` can go.

Which repo gave you the empty strings? I'd like to check whether the PURL is also missing there or just the ecosystem field.

connorshea:
> Which repo gave you the empty strings? I'd like to check whether the PURL is also missing there or just the ecosystem field.

https://github.com/mceachen/bloomer, https://github.com/igorkasyanchuk/active_storage_validations, https://github.com/brianmario/charlock_holmes, and a few other gems

andrew (approved):
LGTM, thanks

PR #259, opened 2026-05-26
feat: add profile support for skill scans

### LINKED ISSUE #254: Ecosystem specific runner profile (container + skill)
Scrutineer runs every scan in the same runner container, which requires extra setup steps for projects whose audit or reproduce phase needs a specific toolchain (PHP, C, ...). Operators are currently able to work around this by maintaining a custom fat container locally.

Here's a proposal to solve this friction. Introduce *runner profiles*: Dockerfile under `docker/profiles/` + manifest-file match rules + allowed skills. After clone, the worker picks the first matching profile (e.g. `composer.json` for `php`, `go.mod` for `go`) or falls back to the implicit `default`. Each profile builds locally on demand (no registry pull, sha-tagged for cache invalidation). The operator can override the auto-detection via `?profile=` when needed.

I can take care of it if this is the right direction. We could ship with one concrete non-default profile (`php` for example) so the mechanism is exercised on day one.

cc @edorian

alexandre-daubois:
Fix #254

@edorian don't hesitate to comment if the new Dockerfile is not exactly what you would need!

For now, it uses Brief to autodetect the profile. A followup could add an input select in the UI for example to force the profile to use.

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -56,6 +56,7 @@ type flags struct {
 	effort           string
 	noDocker         bool
 	runnerImage      string
+	profilesDir      string
 	skillsRepo       string
 	concurrency      int
 	cloneMode        string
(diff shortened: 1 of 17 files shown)


edorian:
Looks great to me from the feature set.

@andrew Feel free to merge when you're happy with the changes and I'll make a follow up for a more detailed PHP container and collaborate on the details with @alexandre-daubois there

andrew (changes requested):
Nice work, this is well structured. One change before merge, rest are notes.

**Blocker:** `imageTag()` hashes only the Dockerfile contents. When `--runner-image` changes, the cached `scrutineer-profile-php:<sha>` was built FROM the old base and won't rebuild. Fold the base image ref into the hash input so a runner bump invalidates profile images too.

Non-blocking:
- `docker/profiles/php/Dockerfile` installs `php84*` unpinned; the main runner Dockerfile pins tool versions, would be good to match that here.
- `TestLockForTag_sameTagSameMutex` checks `got != 0` right after `wg.Go(...)` with no synchronisation, so it passes whether or not the lock blocks. Fine to leave or drop the assertion.
- Auto-detect runs an extra `docker run ... brief` per scan now that `profiles_dir` defaults on. Cheap, just noting it.
- #254 also proposed per-profile allowed-skills filtering which isn't here; might want to reopen it or file a follow-up after this merges since `Fix #254` will auto-close.

alexandre-daubois:
I addressed your comments and created #266 for the followup

andrew (approved):
All review feedback addressed: `imageTag` now folds the runner image ref into the hash (with test coverage), php packages are pinned, the lock test was rewritten without the race, and #266 tracks the allowed-skills follow-up. Thanks!

PR #262, opened 2026-05-26
fix: include skill name in scan queue.

connorshea:
Otherwise we end up with the skill name displayed as "skill" in the Scans page, like this:

Screenshot 2026-05-26 at 4 42 27 PM


andrew (approved):
LGTM, thanks

PR #264, opened 2026-05-27
fix: Improve the code browser UI to have shorter line height.

connorshea:
Before:

Screenshot 2026-05-26 at 8 47 18 PM

After:

Screenshot 2026-05-26 at 8 47 07 PM

The downside is that the highlight line isn't full-width now, but I think this is much better.


andrew (approved):
LGTM, thanks

PR #268, opened 2026-05-27
feat: add path filtering support with `scrutineer.paths` and `scrutineer.ignore_paths`

### LINKED ISSUE #66: Per-skill path scoping and a builtin skip list
Every skill currently sees the whole clone. On large repos the deep-dive spends turns reading lockfiles, minified bundles, and generated code. Let skills declare which paths they care about, and ship a default skip list.

Frontmatter additions under `metadata`:

    scrutineer.paths:        ["src/**", "lib/**"]
    scrutineer.ignore_paths: ["**/*.test.*"]

Builtin skip list applied before any skill runs (overridable by `scrutineer.paths`): `**/pnpm-lock.yaml`, `**/package-lock.json`, `**/yarn.lock`, `**/Cargo.lock`, `**/go.sum`, `**/Gemfile.lock`, `**/poetry.lock`, `**/composer.lock`, `**/*.min.js`, `**/*.min.css`, `**/dist/**`, `**/build/**`, `**/node_modules/**`, `**/vendor/**`, `**/generated/**`, `**/__generated__/**`.

- [ ] Glob matcher in `internal/skills` (no new dependency; `**`, `*`, `?` is enough)
- [ ] Parse `scrutineer.paths` / `scrutineer.ignore_paths` in `internal/skills/parse.go`
- [ ] Apply filters when materialising the workspace in `internal/worker/skill.go`
- [ ] Builtin skip list constant, documented in README skill section
- [ ] Show "N files excluded by path filters" on the scan page

alexandre-daubois:
Fix #66


andrew:
This now conflicts with #269 (both add to the `parse.go` const block / `scrutineerKeys` / `Parsed` / `ToModel`, the `db.Skill` struct, and the docs tables). Should be a mechanical rebase.

One thing while you're in there: `doExposure` in `internal/worker/exposure.go` also materialises a `workRoot/src` from the dependent repo but doesn't get `applyPathFilters`. Dependent repos can have huge `node_modules` too — was that intentional or just missed?

Otherwise looks good, the Copilot points are all addressed.

alexandre-daubois:
That was just missed indeed. `applyPathFilters` is now called after `prepareDependentSrc` in `doExposure`, and since the exposure skill declares no scrutineer.paths, the builtin skip list kicks in and those subtrees get RemoveAll'd before the skill runs.

alexandre-daubois:
I'm taking care of the lint error in a separate PR, see #276

andrew (changes requested):
The mechanism is solid — hand-rolled matcher, `DirAllExcluded` short-circuit, `.git` preservation, parse-time `ValidateGlob`, all good.

One change before merge: drop `**/vendor/**` and `**/build/**` from `BuiltinSkipPaths` (and from the three `SKILL.md` `ignore_paths` blocks plus the README/docs lists).

- `**/vendor/**` strips vendored Go deps, so `go build` fails inside the container unless it can reach the module proxy — which hardened mode blocks. The `patch` skill needs to build.
- `**/build/**` catches projects that keep build scripts or source under `build/`, not just output.

Skills that genuinely want those gone can add them to their own `scrutineer.ignore_paths`. I know #66 spec'd both; we'll update the issue.

Also worth a quick look at `subprojects` (probably detects subprojects by lockfile presence) and `patch` — they may want the same `paths: ["**"]` opt-out the dependency skills got. Fine as a follow-up if you'd rather keep this PR scoped.

Trivial overlap with #277 in `exposure.go` / `skill.go` near the `SkillJob` construction; whichever lands second adds one line.

alexandre-daubois:
First comment addressed, I'm taking care of the second in a follow-up PR. Working on it now. Thanks!

alexandre-daubois:
Actually, neither needs the opt-out: `patch` only emits a diff (it never builds or touches lockfiles), and `subprojects` keys off manifests that were never in the skip list.

andrew (approved):
Thanks — vendor/build dropped everywhere and tests cover it.

PR #269, opened 2026-05-28
feat: add `requires_profile` constraint to skills

### LINKED ISSUE #266: Allowed-skills filter per ecosystem profile
Follow-up of #254 

#259 ships the runner-profile mechanism from #254 but not the *allowed skills* piece from the same proposal. Add a per-profile filter so a profile can declare which skills it's allowed to run (e.g. the `php` profile is the only one allowed to run skills that require a PHP interpreter, and the `default` profile cannot).

alexandre-daubois:
Fix #266


andrew (approved):

andrew:
Thanks!

PR #276, opened 2026-05-28
fix: reduce cognitive complexity of Docker command execution function

alexandre-daubois:
Fix CI


andrew (approved):
LGTM, this also unblocks lint on the other open PRs which are all tripping the same gocognit check on main.

PR #277, opened 2026-05-28
feat: create ephemeral docker networks on hardened mode

### LINKED ISSUE #275: hardened mode: per-scan ephemeral docker networks
Follow-up to #274.

All hardened scans currently attach to a single shared `scrutineer-hardened` `--internal` network. Concurrent scan containers on that network can reach each other, so a hostile clone in one scan could probe or interfere with another scan running at the same time.

Non-hardened mode already has this property on the default bridge so it's not a regression, but for `--hardened` per-scan isolation would be the expected posture.

Proposed: create an ephemeral `--internal` network per scan (e.g. `scrutineer-hardened-<scanID>`), attach only that scan's container, and `docker network rm` it when the scan completes. `ResolveHostGatewayIPv4` already accepts a network name so the gateway probe carries over.

alexandre-daubois:
Fix #275

This also implements a "sweep orphans" mechanism that erases potentially remaining networks from previous crashes.


alexandre-daubois:
Rebasing...

andrew (commented):
Design looks right — per-scan networks, the `ScanID == 0` guard, and the prefix re-check in the sweep are all the right calls.

This now conflicts with #276 (your own `buildDockerArgs` extraction, just merged): the `gwTarget` block and the `--network` flag both moved into that helper. On rebase `buildDockerArgs` will need `perScanNetwork` as a parameter, or the gateway probe moves up into `RunSkill` before the call and the resolved IP gets passed in. Either is fine.

One small thing: the per-scan `ResolveHostGatewayIPv4` spins up a probe container on the new network before the real one, so every hardened scan pays that startup latency. Fine for the strict profile, just noting it.

Happy to approve once it's rebased and CI is green.

alexandre-daubois:
> One small thing: the per-scan ResolveHostGatewayIPv4 spins up a probe container on the new network before the real one, so every hardened scan pays that startup latency. Fine for the strict profile, just noting it.

Happy to investigate whenever it'll become necessary/relevant!

andrew (approved):

PR #278, opened 2026-05-29
feat: Ecosystem runner specific agent instructions

edorian:
This is a two-part change.

The main focus point is runner-specific agent instructions; the PHP runners are a good example and demonstrate the need and approach.

As we have #254 / #259 now with ecosystem-specific containers, this is something we can build on in two ways:

## Runner-specific agent instructions

PROFILE.md auto-injection. When a profile resolves, its docker/profiles/<name>/PROFILE.md is copied to the workspace as  CLAUDE.md before the container starts. claude-code reads that as  project memory. Each profile can give the agent a short orientation without touching the skills.

## Runner-containers more specific than just the ecosystem

**Why?**

PHP has extensions written in C, which need a different container than PHP userland code. But also should have different instructions than other C code.

My proposal is to have simple matching on top of brief for these cases. For PHP extensions it's easy: `config.m4` with `PHP_ARG_` in there should match well enough:

ProfileMarker. Match a profile on a file in the cloned source (path + optional substring), not just on what `brief` reports.  A PHP C-extension repo with a config.m4 containing PHP_ARG_ is now  detectable, even though it has no composer.json. The profile list  is ordered: marker-specific profiles (php-ext) win over general  ecosystems (php / composer).

---

The file is named PROFILE.md to not trigger any auto-injection when the repo is worked on and so it can be copied into an AGENTS.md for a codex or other runners in the future.

---

The main changes are in: f333cea0235c4608d6db1521d8d96cea1121051f

Cleanup diff is a big bigger than needed to avoid `//nolint:gocognit`

### DIFF
--- a/docker/profiles/php-ext/Dockerfile
+++ b/docker/profiles/php-ext/Dockerfile
@@ -1,0 +1,71 @@
+# PHP-extension profile.
+#
+# Ships PHP, built from source with --enable-debug + ASan + UBSan.
+
+ARG RUNNER_IMAGE=ghcr.io/alpha-omega-security/scrutineer-runner:latest
+ARG PHP_GIT_URL=https://github.com/php/php-src.git
+ARG PHP_GIT_REF=php-8.5.6
+
+FROM composer:2.9.8 AS composer
+
+FROM ${RUNNER_IMAGE}
+ARG PHP_GIT_URL
+ARG PHP_GIT_REF
+
+USER root
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    PATH=/usr/local/php/bin:/usr/local/php/sbin:$PATH \
+    COMPOSER_ALLOW_SUPERUSER=1 \
+    USE_ZEND_ALLOC=0 \
+    ASAN_OPTIONS="symbolize=1:strict_string_checks=1:detect_stack_use_after_return=1:detect_leaks=0:print_summary=1" \
+    UBSAN_OPTIONS="print_stacktrace=1:print_summary=1"
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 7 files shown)


andrew (approved):
Nice, both mechanisms are clean and the php-ext PROFILE.md is genuinely useful guidance.

Two small things before merge:

- #276 just landed and also touched `RunSkill` in `internal/worker/docker.go` (extracted `buildDockerArgs` for the same gocognit reason). Your `checkHardenedWorkspace` extraction is in a different part of the function so it may auto-merge, but please rebase onto main and re-push so CI runs against the combined shape.
- Typo in `docker/profiles/php/PROFILE.md`: "preperations" → "preparations".

Heads up that #279 (Ruby profile) will trivially conflict on `builtinProfiles` with this; whichever lands second just appends a line.

andrew:
#279 (Ruby profile) just merged too, so when you rebase there are a couple of extra things in `profile_test.go` to reconcile:

- `TestBuiltinProfiles_registrySanity` (new) asserts every profile has a non-empty `Ecosystem` — `php-ext` is marker-only, so that check needs to allow empty `Ecosystem` when `Markers` is set.
- Two new ordering test cases (`"ruby before php picks ruby"` / `"php before ruby picks php"`) encode the old first-package-manager-wins behaviour. With your registry-order matching, a repo with both Composer and Bundler will deterministically pick whichever comes first in `builtinProfiles` — those cases need updating to reflect that.
- `TestRepoShipsProfileDockerfiles` was generalised to loop over `builtinProfiles`; your version of the same test also checks for `PROFILE.md`. Easy to combine, just flagging the textual conflict.

edorian:
@andrew Looks good from my side now.

Rebuild all containers (PHP, PHP-Ext, Ruby) and everything works and runners execute as expected

PR #279, opened 2026-05-29
feat: Add a Ruby Dockerfile profile and basic tests.

connorshea:
Not sure if we need more comprehensive tests here. Also, ruby-build is pulling a tag, I can update this to pin to a SHA hash if preferred.

Generated with Claude Code, built the Docker image myself to confirm it works, and also used this branch to scan a ruby gem and it worked and was able to reproduce a vuln by running ruby :D

### DIFF
--- a/docker/profiles/ruby/Dockerfile
+++ b/docker/profiles/ruby/Dockerfile
@@ -1,0 +1,58 @@
+# Ruby profile. Extends the scrutineer runner with a modern Ruby
+# interpreter and Bundler so scans of Bundler/RubyGems projects can audit
+# or reproduce in-place.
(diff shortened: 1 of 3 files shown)


connorshea:
https://github.com/alpha-omega-security/scrutineer/pull/276 should fix the lint failure here

andrew (approved):
LGTM. Pinned ruby-build + source compile is the right call over the distro package.

One thought for a follow-up: Ruby 4.0.0 is out, might be worth bumping `RUBY_VERSION` once it's had a point release, though 3.4 is the safer default for compatibility with what's actually being scanned today.

Issue #255, opened 2026-05-26
dependencies skill: Claude hand-authors rows when git-pkgs returns empty


andrew:
When `git-pkgs list` returns nothing for a repo (e.g. a Bones Rakefile project with no Gemfile or gemspec), the model reads the manifest itself and writes dependency rows by hand. Scan 7419 (`twp/little-plugger`) log: "I wrote report.json by hand from the Rakefile". Same pattern for `janlelis/value_struct` (Ore `gemspec.yml`) and `forgejo/forgejo` (`.forgejo/workflows/*`).

The hand-written rows use whatever ecosystem string the model picks (`rubygems`, `githubactions`) rather than the git-pkgs vocabulary, and sometimes omit the PURL entirely (all 14 forgejo-action rows have `p_url=""`). That breaks the `DependencyFindings` join and the Packages filter.

Fix options:

- Tighten `skills/dependencies/SKILL.md`: if git-pkgs returns nothing, write `{"dependencies": []}` and stop. Do not infer dependencies from files git-pkgs does not parse.
- Or, if augmentation is wanted, make `purl` a required field in `skills/dependencies/schema.json` so every row has a canonical identifier, and validate against the schema before accepting the report.

Related: https://github.com/alpha-omega-security/scrutineer/pull/248 (ecosystem normalisation discussion) and https://github.com/git-pkgs/git-pkgs/issues/207 (the `null` vs `[]` output that triggers the improvisation).

PR #285, opened 2026-05-29
feat: enhance ecosystem handling with PURL integration

### LINKED ISSUE #256: DependencyFindings: join on PURL type+name instead of ecosystem string
`db.DependencyFindings` (internal/db/ecosystem.go:53) currently joins Dependency rows to Package rows on `(name, CanonicalEcosystem(ecosystem))`. The two tables are populated from different sources (git-pkgs vs packages.ecosyste.ms) which use different ecosystem vocabularies, so `dependencyEcosystemAlias` exists to translate between them at read time.

Both sources emit a PURL. Keying the join on parsed PURL `(type, namespace, name)` removes the need for the alias map and for any write-time normalisation.

Concretely:

- Add `github.com/git-pkgs/purl` as a direct dep.
- In `skill_parsers.go`, set `Ecosystem` from `purl.Parse(p.PURL).Type` with `purl.EcosystemToPURLType(p.Ecosystem)` as the fallback when no PURL is present.
- In `DependencyFindings`, build the join key from the parsed PURL on both sides and drop `dependencyEcosystemAlias` / `CanonicalEcosystem`.

Existing rows converge on next scan since the parsers delete-then-insert per repo.

Related: https://github.com/alpha-omega-security/scrutineer/pull/248 (where this came up) and https://github.com/alpha-omega-security/scrutineer/issues/255 (model-authored rows that sometimes lack a PURL).

alexandre-daubois:
Fix #256


andrew (approved):
Matches #256 as spec'd — PURL-keyed join, alias map gone, and the mixed-PURL fallback tests cover #255 nicely. Dep graph is fine (same org as `sbom`, three small modules).

Two notes, neither blocking:

- The registry-URL fallback is dropped. #256 assumes both sources always emit a PURL; if that turns out not to hold, rows with no ecosystem and no PURL will store `""` where they previously got a value from `RegistryURL`. Happy to take that and see if it bites.
- Stored `ecosystem` flips vocab (`rubygems`→`gem`, `go`→`golang`, `packagist`→`composer`), so the Packages filter will show both until repos rescan. Expected per #256.

This now conflicts with main after #278 — give it a rebase and I'll merge.

alexandre-daubois:
Rebased!

andrew:
Thanks!

PR #292, opened 2026-05-30
feat: add support for Claude Opus 4.8

alexandre-daubois:


connorshea:
will this work without upgrading the claude-code package we ship?

andrew:
Agree with @connorshea — let's bump the shipped CLI alongside this so there's no doubt. Could you update `CLAUDE_VERSION` in `Dockerfile.runner` (and the checksum lines) and the `@anthropic-ai/claude-code` pin in `Dockerfile` to `2.1.160` in this PR? Then we can land both together.

andrew (approved):

Issue #294, opened 2026-05-30
Top-level Repositories list shouldn't show Findings that have been rejected or resolved


connorshea:
Right now my Repositories list looks like this and so I have a bit of trouble remembering which of these are new / rejected / resolved, it might be good to make the column _only_ include non-reject non-resolved findings.

<img width="92" height="569" alt="Image" src="https://github.com/user-attachments/assets/65bc790f-5f7c-4105-9921-af7a0217864a" />

alexandre-daubois:
Should it be closed as fixed by #296?

PR #296, opened 2026-05-31
Fix repository findings count for closed findings

abhinavgautam01:
Closess #294

updates the top-level Repositories list so the Findings column only counts open findings. Findings with closed lifecycle statuses are now excluded from the repo list count:

- `fixed`
- `published`
- `rejected`
- `duplicate`

The `sort=findings` behavior now uses the same filtering, so repos with only closed findings no longer sort above repos with active findings.

## Testing

- `go test ./internal/web`
- `go test ./...`
- `git diff --check`


andrew (approved):
lgtm

PR #298, opened 2026-05-31
fix: update allowed tools to include Skill in permission mode

alexandre-daubois:
I systematically ran into Claude telling that `the skill invocation failed` for every skill I ran. It seems that we're missing `Skill` in the allowed tools and the model then tries what it can with the skill name only, but often fails.

Also removed max turns of `maintainers`, I ran every time into the limit.


andrew (approved):
lgtm — skills weren't usable under acceptEdits without this.

Issue #160, opened 2026-05-05
Capture Claude session ID and pass --resume on retry


andrew:
When goqite redelivers a failed skill job the worker starts a fresh `claude -p` invocation, so a `security-deep-dive` that hits an API timeout at turn 130 loses everything and begins again from turn 0. `claude -p` supports `--resume <session-id>` to continue an existing conversation with full history; the harness already does the hard part, scrutineer just isn't wiring it through.

`claude --output-format stream-json` emits a `session_id` in both the init system message and the final result message. `internal/worker/stream.go` currently ignores it. Proposal:

- Extend `streamMsg` and `ParseStream` in `internal/worker/stream.go` to capture `session_id` and surface it on `SkillResult`.
- Add a `SessionID string` column to `db.Scan` and write it as soon as the init event arrives, before the run finishes, so it survives a crash.
- In `internal/worker/claude.go` and `internal/worker/docker.go`, when the incoming `Scan` already has a non-empty `SessionID` and `Status != "done"`, build args as `claude -p --resume <id> ...` instead of passing the activation prompt again.
- Clear `SessionID` when the scan reaches `done` so a deliberate re-run from the UI starts fresh.

The docker runner needs the Claude Code session store to persist across container restarts for this to work; mounting `~/.claude` (or wherever `CLAUDE_CONFIG_DIR` points) into the container alongside the workspace volume should be enough.

This is the cheap fix for the all-or-nothing problem on long skills; #20 can stay focused on repos that are too large for a single context even with resume.

Issue #20, opened 2026-04-20
Better strategies for huge projects


andrew:
spec-deep phase 1 inventory struggles on large codebases. Explore chunking, per-directory scans, and model-backed prioritisation of where to look first.

andrew:
This is partly done, but needs more testing

connorshea:
One interesting thing I've found to work with scanning a large codebase using "raw" Claude has been providing it with a "lead", e.g. "look for XSS/HTML injection vulns in the comments system for this app" or "look at the security of the auth/login system" or "PDF exports", etc. I wonder if it'd make sense to allow the user to trigger "New Scan" with an additional short prompt to guide what Claude should focus on for the given scan.

andrew:
[google/mantis](https://github.com/google/mantis) `mantis-researcher/SKILL.md` (Apache-2.0) does this in two waves:

Wave 1 runs 10-20 cheap-model sub-agents over every file in scope, each returning only `{"potentially_flawed": bool, "reason": ...}`. Wave 2 runs 4-8 deep audits on the flagged set with the heavier model, and each sub-agent writes its own finding file and returns just the id so the parent's context window doesn't fill up.

That maps onto the model-tier config we already have (flash for wave 1, mid/frontier for wave 2) and the per-focus-area scan split. #106's per-class audit skills are effectively the wave-2 half.

They also have `mantis-summarize` doing per-directory summaries bottom-up (children first, parent sees child summaries not child source), which is the "chunking" option, and `mantis-structural-index` building a call-edge SQLite index to rank where wave 1 looks first. The index is a lot of machinery and their own rule is that it only ranks and never filters, so grep alone finds the same set. Probably not the first thing to reach for.

andrew:
[...]

PR #301, opened 2026-06-01
Capture Claude session id and --resume failed scans on retry

connorshea:
Implements #160.

Generated with Claude Code, reviewed and tested by me (though I'm not particularly familiar with Go, so take that with a grain of salt). I'm unsure if this needs special handling for hardened mode.

Confirmed it works by starting a scan, killing the Scrutineer server mid-run, and then hitting "Retry" and seeing it successfully resume the same session:

Screenshot 2026-05-31 at 7 28 39 PM

---

A long skill (e.g. `security-deep-dive`) that hit an API timeout at turn 130 used to restart from turn 0 on retry, losing everything. claude-code supports `--resume <session-id>`; this wires it through so a retry of a failed scan continues the conversation with full history.

## What it does
- Captures `session_id` from the stream-json **init** event and persists it on the scan the moment it appears, so it survives a crash. Cleared on `done` (a deliberate re-run starts fresh); kept on `failed` (a retry can resume).
- The UI **Retry** button (and retry-all-failed) carries a failed scan's session into the new run. claude keys its session store by **working directory**, so the retry reuses the lineage-root workspace path (`ResumedFromScanID`) to keep cwd stable across the retry chain.
- The **docker** runner mounts a persistent per-lineage `CLAUDE_CONFIG_DIR` so the session store survives a container exit (works under hardened mode's `--read-only` rootfs). The **local** runner relies on the host `~/.claude`.
- **Resume-failure fallback:** if `--resume` can't find the session (expired/pruned — no init event, exit 1), the runner restarts once *fresh* in the same call, so a lost session doesn't permanently wedge the retry lineage.

## Note on the trigger (divergence from the issue text)
The issue framed this around goqite redelivering a failed job. In practice `wrap()` returns `nil` to goqite on handler errors *specifically* to suppress auto-retry (`internal/worker/worker.go`), so redelivery never fires — and a crash + restart path is killed by `SweepRunning` + the terminal-drop. So the trigger here is an **explicit UI retry** instead. This avoids reversing the deliberate no-auto-retry contract and spending API budget without a human in the loop. (#20 can stay focused on repos too large for a single context even with resume.)

## Verified
- `--resume` is cwd-scoped (confirmed empirically): the session `.jsonl` lives at `$CLAUDE_CONFIG_DIR/projects/<encoded-cwd>/`, and resume fails from a different cwd — hence the lineage-stable workspace.
- A *failed* `--resume` echoes a throwaway `session_id` in its `result` event but emits no `system/init`; capturing from init only (and using "no init seen" as the fallback signal) handles this correctly.
- Positional prompt + `--resume` parses and loads the conversation.

## Tests
Stream capture, failed-resume-emits-no-session, `buildClaudeArgs` resume, an end-to-end local fallback driving a fake `claude`, docker config-mount, worker persist/clear/lineage, and web `resumeOpts`. `go build`, `go vet`, and `go test ./...` all pass.

## Known wart (out of scope)
Config dirs and local `~/.claude` transcripts for lineages that fail and are *never* retried accumulate; a startup sweep could clean those later.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -157,6 +157,24 @@ type Scan struct {
 	// operator's override and the UI can show the chosen ecosystem.
 	Profile string `gorm:"index"`
 
+	// SessionID is the claude-code session this scan's run belongs to,
+	// captured from the stream-json init/result events. It is written as
+	// soon as the init event arrives (before the run finishes) so it
+	// survives a crash, and cleared once the scan reaches "done" so a
+	// deliberate re-run from the UI starts a fresh conversation. A retry
+	// of a failed scan carries this value forward so the runner can pass
+	// `claude -p --resume <id>` and continue from where it left off
+	// instead of restarting from turn 0.
+	SessionID string
+	// ResumedFromScanID points at the lineage-root scan whose claude session
+	// and workspace a retry reuses. Nil on a fresh scan. claude keys its
+	// session store by working directory, so a resuming run must execute
+	// in the same per-scan workspace path as the original; this pins that
+	// path across the whole retry chain. Always the root of the lineage,
+	// not the immediate parent, so N retries deep still resolve to one
+	// workspace.
+	ResumedFromScanID *uint `gorm:"index"`
+
 	Commit     string
 	StartedAt  *time.Time
 	FinishedAt *time.Time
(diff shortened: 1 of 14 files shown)


andrew (approved):
lgtm

andrew:
Merged. One follow-up worth flagging: `clearSessionStore` only runs on `ScanDone`, so a cancelled scan leaves its `SessionID` and `claude-config/scan-N/` dir on disk. Same kind of leak you already noted for failed-and-never-retried lineages — probably worth handling in the same cleanup pass.

PR #302, opened 2026-06-01
feat: Add "About" section to Settings page

connorshea:
And also split "Findings" from "Scanner Findings" in the statistics.

Updated Settings content:

Screenshot 2026-05-31 at 8 05 54 PM

Generated with Claude Code, reviewed and tested by me. I'm unsure if there are any security considerations to make about this information being available inside scrutineer, but I don't think there's any real problem with it?

Also, `Runner image` is just "ghcr.io/alpha-omega-security/scrutineer-runner:latest" which feels... goofy. We could probably fix that to specify the SHA somehow?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,12 @@ WORKDIR /src
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . .
-RUN CGO_ENABLED=0 go build -o /scrutineer ./cmd/scrutineer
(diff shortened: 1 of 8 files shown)


andrew (approved):
lgtm

andrew:
Merged — let's look at pinning the runner image to a digest (rather than `:latest`) as a follow-up. Same goes for surfacing it on this page once we do.

PR #303, opened 2026-06-01
feat: add support for specific branch scanning

alexandre-daubois:
Many projects maintain multiple branches at the same time, I thought it would be convenient to provide an optional branch to the "Add Repository" feature:

Capture d’écran 2026-06-01 à 11 01 50

Branches are automatically fetched, providing autocompletion. A branch badge is also added to repositories and scans:

Capture d’écran 2026-06-01 à 11 01 09

Capture d’écran 2026-06-01 à 11 01 18


andrew (changes requested):
Looks good overall — the `fetchRef` move to `FETCH_HEAD` and the `--` guards are nice.

One inconsistency: on the existing-checkout path `fetchRef` happily takes a tag or commit SHA (`git fetch origin <ref>` → `FETCH_HEAD`), but on first clone the ref goes through `git clone --branch <ref>`, which only accepts branch and tag names. So a SHA in the branch field works on the second scan but fails on the first.

Could you make the two paths agree? Easiest is probably to drop `--branch` from the initial clone and call `fetchRef(ctx, dst, ref, fullClone, emit)` right after cloning, so both first-clone and re-fetch resolve refs the same way.

alexandre-daubois:
Should be good now 🙂

andrew (approved):
Thanks — both clone paths now resolve refs identically, and the SHA case in the test pins exactly that.

andrew:
Hit a conflict merging — 312 just touched the same templates. Rebase and I'll merge straight away.

alexandre-daubois:
Here you go

andrew:
@alexandre-daubois got a vuln check test failure

alexandre-daubois:
It is unrelated to this PR, I'm creating a new one to fix it

andrew:
@alexandre-daubois ah its go version, i've got it

andrew:
Bumped Go on main, so just needs a rebase

alexandre-daubois:
Rebased and other branches as well 🙂

andrew:
Conflicts again after merging #310 and #307 — could you rebase once more? Should be the last one.

alexandre-daubois:
No worries, here you go

andrew:
🫶

PR #297, opened 2026-05-31
Improve scan UX for Claude plan limits

abhinavgautam01:
closes #293

## Summary

Improves the operator experience when Claude plan/session limits are hit during scan runs...

Changes include:

- Add a new `paused` scan status.
- Add **Pause queued** and **Resume paused** actions on the Scans page.
- Add per-scan resume support for paused scans.
- Make workers safely ignore paused/stale queue messages.
- Detect Claude usage/session/rate-limit style failures and surface a clearer error message.
- Show a Scans page warning when Claude plan-limit failures exist while scans are still queued.

## Testing

- `go test ./internal/db ./internal/worker ./internal/web`
- `go test ./...`
- `git diff --check`

PR #307, opened 2026-06-01
feat: add concurrency and max turns settings tuning

alexandre-daubois:
Adds two "knobs" in the settings to quickly customize concurrency and max turns:

Capture d’écran 2026-06-01 à 15 05 04

Because concurrency requires a runner reboot, it is done automatically. If scans are running, a warning is displayed:

Capture d’écran 2026-06-01 à 15 05 20

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ Every index page has a search box plus filter and sort dropdowns; the specifics 
 - **Scans** -- every scan that has run. Queued scans can be paused/resumed, running or queued scans can be cancelled and failed ones retried.
 - **Skills** -- installed skills from disk and from the UI; view, edit, or run any of them.
 - **Usage** -- token and cost totals across all scans, broken down by skill.
-- **Settings** -- theme, colour scheme, default model, and system stats (record counts, DB size, concurrency, paths).
+- **Settings** -- theme, colour scheme, default model, runner concurrency (restarts the runner to apply, cancelling in-flight scans) and default turn cap (applied to the next scan), plus system stats (record counts, DB size, paths).
 
 ## Finding workflow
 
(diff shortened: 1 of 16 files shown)


andrew:
Code looks good — the `Reconfigure` swap and the save-then-confirm flow are both nicely done.

It now conflicts with #297 in `scans.go`, `server.go`, `sse.go` and `theme.go`: both PRs pulled the `"success"` flash literal out into a constant, this one as `flashSuccess` and #297 as `successKey` (alongside `errorKey`/`warningKey`). #297 has landed, so could you rebase and drop `flashSuccess` in favour of the existing `successKey`? Happy to merge after that.

alexandre-daubois:
Comments addressed!

andrew (approved):
The queue refactor is the substantive part and the tests cover it well — `ReconfigureLive` proves the end-to-end behaviour (cancel in-flight, apply new limit, fresh runner picks up new jobs), `ReconfigureDuringShutdown` guards the WaitGroup panic, and `ReconfigureBeforeStart` handles the lazy-init case. I checked `finishScan` — `context.Canceled` already maps to `ScanCancelled`, so in-flight scans get cleanly marked rather than stuck in `running` when Reconfigure tears down the runner.

The `db.Setting` table being generic (key/value) is a reasonable bet — easy to add more knobs without another column. Persisted-but-flag-overrides precedence in `run()` mirrors the existing pattern.

Two small things, neither blocking:
- The OOB confirm leaves the new value persisted even if the user clicks "Not now". That matches what the help text says ("takes effect on the next restart"), but it does mean reloading the settings page shows the new value as active when it's actually not in effect yet. Minor.
- The race window where Reconfigure could spawn a new runner just after parentCtx is cancelled is acknowledged in the comment and not worth fixing for a single-operator app.

Conflicts after #311 — rebase and I'll merge.

alexandre-daubois:
Rebased!

PR #310, opened 2026-06-02
feat: add support for repository removal

alexandre-daubois:
Allows removing a repository and its related data. Space used by the repository appears in the repositories list and in the summary page.

Capture d’écran 2026-06-02 à 11 12 21

A cache or a persistance of the disk size could be useful if you ever come across performance issues. But I think it's fine with 20 repos per page?

Capture d’écran 2026-06-02 à 11 12 34

### DIFF
--- a/internal/web/repo_delete_test.go
+++ b/internal/web/repo_delete_test.go
@@ -1,0 +1,300 @@
+package web
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"scrutineer/internal/db"
+	"scrutineer/internal/worker"
+)
+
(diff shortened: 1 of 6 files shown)


andrew (approved):
The transaction ordering rationale is well documented and the test forcing `SetMaxOpenConns(1)` + `PRAGMA foreign_keys=ON` to actually enforce FKs is a nice touch — sqlite's per-connection pragma would otherwise let a wrong order pass silently.

One thing worth following up later: if a scan is actively running when delete fires, the worker keeps writing into a removed inode until it returns. Not catastrophic for the single-operator model, but the confirm could surface "running scans on this repo, cancel first?" at some point.

Conflicts with `repo_list.html` / `repo_show.html` after #312 — rebase and I'll merge.

alexandre-daubois:
I just rebased. I'll take care of the follow-up today!

PR #346, opened 2026-06-08
worker: include rendered SKILL.md in scan.Prompt

### LINKED ISSUE #308: Expose the full skill file in the Prompt logged for a scan
It'd be nice to have the actual skill file in the Prompt section, right now it only displays a generic message like this:

Image

andrew:
The Prompt tab on a scan only showed the short activation wrapper ("Use the X skill on ./src..."). The actual instructions live in the SKILL.md the skill loads, which the UI never surfaced — operators had to dig into the skills page or the source tree to see what Claude actually ran.

Append the rendered SKILL.md to `scan.Prompt` at scan time so the Prompt tab shows both the wrapper and the body that was active for this run.

Closes #308

### DIFF
--- a/internal/worker/claude.go
+++ b/internal/worker/claude.go
@@ -305,6 +305,15 @@ func buildSkillPrompt(name, outputFile string) string {
 	return p
 }
 
+// buildLoggedPrompt is what scrutineer records on scan.Prompt for the UI. It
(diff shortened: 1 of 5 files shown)


connorshea (approved):
Nice, thanks!

PR #349, opened 2026-06-09
worker: gate skill dispatch on declared prereqs

### LINKED ISSUE #332: "Dependent" skills
Should it be possible to start the `security-deep-dive` skill before `threat-model` is done? It seems like we're just burning tokens unnecessarily and sending Claude in blind if it doesn't have a threat model defined before it starts the security scan. I'm not sure of the best way to resolve this, but it seems worth resolving?

andrew:
A SKILL.md may now declare `scrutineer.requires: [name, ...]`. Before dispatching a job the worker checks each name has a `done` scan on the same repository. If any are missing the queue message is re-published with a delay and the scan row stays `queued`; after 20 retries (default, configurable via `Worker.MaxPrereqAttempts` / `Worker.PrereqRetryDelay`) the scan fails with a "prereqs not satisfied" error.

A prereq skill that has never been seeded is treated as satisfied so that a triage-gated upstream skill (e.g. `dependents` on a no-packages repo) does not deadlock its dependents.

Two skills now declare prereqs:

- `security-deep-dive` waits on `threat-model`, `semgrep`, `repo-overview`, `advisories`, `packages`, `dependents`, `maintainers`.
- `dependents` waits on `packages`.

Closes #332.


andrew:
Two follow-ups not in this PR:

1. **`dependents` early-exit on empty packages.** The gate now ensures `packages` has run before `dependents` starts, but `dependents` still hits packages.ecosyste.ms directly rather than reading scrutineer's `/repositories/{id}/packages` first. A small SKILL.md edit to short-circuit when the packages list is empty would actually realize the token saving the gate sets up.
2. **Visible "waiting on X" scan status.** A scan deferred by the prereq gate currently looks identical to a normal queued scan. If operators want to see what it's blocked on, that needs a new column on `scans` (e.g. `waiting_on`) and a UI surface for it. Re-queue with delay was chosen for v1 over a real `waiting` status to keep the change small.

internal/worker/worker.go
```diff
@@ -35,6 +35,17 @@ const (
...
 
+// Prereq gate defaults. A skill declaring scrutineer.requires has its
+// dispatch deferred when any listed upstream scan is not yet done; the
+// queue message is re-published with DefaultPrereqRetryDelay between each
+// attempt, up to DefaultMaxPrereqAttempts. The two together cap the wait
+// at roughly DefaultPrereqRetryDelay * DefaultMaxPrereqAttempts before the
+// scan fails with a "prereq not satisfied" error.
+const (
+	DefaultPrereqRetryDelay  = 30 * time.Second
+	DefaultMaxPrereqAttempts = 20
```
edorian, line 49: I'd suggest that 10 minutes isn't enough. And 60-120 prereq attempts are better, as things tend to run 5-15 minutes for me at times.

edorian:
Did a bit of testing, and this looks like it will resolve my current issues well 👍

naderman:
Just ran into this too, realizing all the "security deep dive" executions were pretty basic and saying they couldn't find any threat models or semgrep results. Turns out that's because they were started around the same time, and simply not done yet.

edorian:
Is there anything I can help to move this along? Setting up Nils for scanning, we ran into this:

>  "inventory_notes": "No threat-model or semgrep scan was available from scrutineer (both endpoints returned empty)" aber die scans haben geklappt und resultate waren non-empty

and 

> "prior_art": "Scrutineer advisory cache `GET /repositories/2/advisories` returned `[]` — no CVE/GHSA has ever been published against any ..... No prior threat-model or semgrep scan exists for this repo (`GET .../scans?skill=threat-model` and `.../findings?skill=semgrep` both empty), so Phase-1 boundaries were derived from source. 

due to concurrency :)

connorshea:
as long as you rebase this branch I assume you can just run it locally for the moment, until it's merged?

PR #350, opened 2026-06-09
Add finding deduplication skill

### LINKED ISSUE #348: Finding deduplication skill
The built-in fingerprinting works pretty well already, but I've been getting a lot of duplicates that I think could be resolved by an extra skill run.

For example, because the finding fingerprint is based on the location, any minor difference in the location is treated as a different finding:

`Finding A: Insufficient permissions check at foo/bar.rb:45-51`

`Finding B: Insufficient permissions check for DELETE at foo/bar.rb:45-52`

These are the exact same finding, but the LLM will arbitrarily include or exclude whitespace or method name, and so get a distinct line range.

In some cases this may be valid and two findings may overlap, so I don't think there's a good heuristic we can use to automatically handle this. So maybe we should have an extra step/skill that has Claude evaluate each finding on a given Repository to determine whether any are dupes of one another?

Another common case where I've seen this occur has been multi-file findings. Different scans will output very different Location values because we have no guardrails (afaik) around how to communicate multi-file Locations. 

`foo/bar.rb:45 (sink), foo/baz.rb:57 (vuln)` and things like that end up duplicating often because the format Claude emits here is a lot more random than even line ranges.

abhinavgautam01:
closes #348 
## Summary

Adds a repository-scoped `finding-dedup` skill that compares open findings and marks overlapping reports as duplicates when they describe the same underlying vulnerability.

## Changes

- Add bundled `skills/finding-dedup` skill with JSON schema
- Add new `finding_dedup` output kind and worker parser
- Mark duplicate findings through `db.WriteFindingField` so status history is preserved
- Add a note to duplicate findings pointing to the canonical finding
- Skip invalid, cross-repository, self-referential and closed findings
- Document the new skill and parser kind

## Testing

- `go test ./internal/worker ./internal/skills`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ When a repo is added, the `triage` skill is enqueued. Its SKILL.md lists the ski
 | `semgrep` | Static analysis mapped into findings shape |
 | `zizmor` | GitHub Actions workflow audit mapped into findings shape |
 | `security-deep-dive` | The model-backed audit producing structured findings |
+| `finding-dedup` | Compares open findings and marks overlapping reports as duplicates |
 | `verify` | Re-checks one finding against current HEAD; records reproduces / fixed / can't-reproduce |
 | `disclose` | Drafts a GHSA-shaped advisory (title, description, CVSS, CWEs, references) for one finding |
 | `patch` | Proposes a unified diff fixing one finding, written back as a note for analyst review |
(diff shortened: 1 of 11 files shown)


andrew (approved):
This is well put together - the parser re-validates everything server-side (repo ownership, open status on both sides, self-reference), status changes go through `WriteFindingField` so history is preserved, and the conflict ordering resolves safely since statuses are re-read per group.

One non-blocking note: the README adds `finding-dedup` to the table introduced by "the `triage` skill is enqueued. Its SKILL.md lists the skills it can dispatch", but `skills/triage/SKILL.md` wasn't updated, so nothing actually dispatches this skill yet - it's manual-only. Worth either teaching the triage prompt about it or adjusting the README placement in a follow-up.

PR #351, opened 2026-06-09
Add Rust ecosystem profile and support

walterhpearce:
PR includes two changes:
- Addition of a rust runner profile 
- Added additional language to the extraction skill. The tooling was struggling extracting crate file sources, so we needed to specify they are tar files.

### DIFF
--- a/docker/profiles/rust/Dockerfile
+++ b/docker/profiles/rust/Dockerfile
@@ -10,8 +10,8 @@ ARG RUST_STABLE_VERSION=1.96
 ARG RUST_NIGHTLY_VERSION=nightly-2026-06-24
 
 USER root
-RUN apt update \
- && apt install -y --no-install-recommends \
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends \
         build-essential \
         pkg-config \
         gcc-multilib \
--- a/docker/profiles/rust/Dockerfile
+++ b/docker/profiles/rust/Dockerfile
@@ -14,7 +14,6 @@ RUN apt-get update \
  && apt-get install -y --no-install-recommends \
         build-essential \
         pkg-config \
-        gcc-multilib \
         ca-certificates \
         autoconf \
         automake \
@@ -24,9 +23,13 @@ RUN apt-get update \
         wget \
  && rm -rf /var/lib/apt/lists/*
 
-# Add rust to the container
+# Add rust to the container. MIRI_SYSROOT is pinned so the sysroot built
+# below as root is the one the runner user finds at scan time (the worker
+# runs with HOME=/tmp, which is noexec, so the default ~/.cache/miri
+# location would force a rebuild that then fails to execute).
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
+    MIRI_SYSROOT=/usr/local/miri-sysroot \
     PATH=/usr/local/cargo/bin:$PATH \
     RUSTUP_VERSION=${RUSTUP_VERSION} \
     RUST_STABLE_VERSION=${RUST_STABLE_VERSION} \
@@ -80,10 +83,14 @@ RUN set -eux; \
     cargo --version; \
     rustc --version;
 
-# Install nightly
+# Install the pinned nightly with rust-src (for -Zbuild-std / sanitizers)
+# and miri, then run miri's one-time sysroot build so scans don't pay it.
+# The bare `nightly` channel is deliberately not installed; select this
+# toolchain with `+${RUST_NIGHTLY_VERSION}` (exported via ENV above).
 RUN rustup toolchain install ${RUST_NIGHTLY_VERSION} \
-       --component rust-src
-RUN rustup component add --toolchain nightly-x86_64-unknown-linux-gnu miri \
-       && cargo +nightly miri setup
+        --component rust-src \
+        --component miri \
+ && cargo +${RUST_NIGHTLY_VERSION} miri setup \
+ && chmod -R a+w $RUSTUP_HOME $CARGO_HOME $MIRI_SYSROOT
 
 USER runner
--- a/docker/profiles/rust/Dockerfile
+++ b/docker/profiles/rust/Dockerfile
@@ -1,0 +1,89 @@
+# docker pull ghcr.io/alpha-omega-security/scrutineer-runner:latest && \
+# docker build -t scrutineer-profile-rust:latest docker/profiles/rust
+
+ARG RUNNER_IMAGE=ghcr.io/alpha-omega-security/scrutineer-runner:latest
+
+FROM ${RUNNER_IMAGE}
+
+ARG RUSTUP_VERSION=1.29.0
+ARG RUST_STABLE_VERSION=1.96
+ARG RUST_NIGHTLY_VERSION=nightly-2026-06-24
+
+USER root
+RUN apt update \
+ && apt install -y --no-install-recommends \
+        build-essential \
+        pkg-config \
+        gcc-multilib \
+        ca-certificates \
+        autoconf \
+        automake \
+        libssl-dev \
+        libssh-dev \
+        git \
+        wget \
+ && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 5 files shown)


andrew (changes requested):
Thanks for this - a Rust profile is very welcome, and the `.crate` extraction hint in the skill is a genuinely useful addition.

One blocker: the profile is never registered, so it can't be selected. Profiles are picked from `builtinProfiles` in `internal/worker/profile.go` - the Dockerfile needs a matching entry, something like:

```go
{Name: "rust", Ecosystem: "Cargo"},
```

Without it, `ProfileByName("rust")` falls back to the default runner and ecosystem matching never picks the new image up.

A few smaller things on the Dockerfile:

- The other profiles pin their toolchains for reproducibility (the ruby profile pins the interpreter version and the ruby-build release tag). `curl https://sh.rustup.rs | bash` plus an unpinned `nightly` makes the image a moving target - a dated toolchain like `nightly-2026-06-01` would match the repo's conventions.
- The apt list looks copied from the ruby profile: `libyaml-dev`, `libreadline-dev`, `libgdbm-dev`, `libncurses-dev` etc. are Ruby build deps and shouldn't be needed for Rust. The apt lists also aren't cleaned up afterwards (`rm -rf /var/lib/apt/lists/*`).
- `ENV RUSTFLAGS='-Zsanitizer=address'` forces ASan on every cargo invocation in the container, which is slow and can break proc-macro and build-script compilation in surprising ways. If ASan is the point of the profile, it might be better for the skill to opt in per command rather than baking it into the image.
- Cosmetic: trailing whitespace after `libgdbm-dev`, and the file is missing a trailing newline.

In the SKILL.md, suggest tightening the new sentence to: "Rust `.crate` files are gzipped tarballs."

Muscraft:
> The other profiles pin their toolchains for reproducibility (the ruby profile pins the interpreter version and the ruby-build release tag). `curl https://sh.rustup.rs | bash` plus an unpinned `nightly` makes the image a moving target - a dated toolchain like `nightly-2026-06-01` would match the repo's conventions.

I would agree that pinning at least the nightly version would be a good idea, but I would go a step further and also pin the `rustup` version as well. That is what [Rust's official nightly images](https://github.com/rust-lang/docker-rust/blob/1c612e8be5586f5288dc5a1d58b11a5c51f978a8/nightly/trixie/Dockerfile) do, and it would probably be good to match[^1].



[^1]: Note: I am biased on this as I am [one of the maintainers](https://rust-lang.org/governance/teams/infra/#team-docker) of Rust's official images

andrew:
Thanks for this. One thing to flag: as it stands the Dockerfile will not actually be picked up at scan time. A profile is three pieces wired together, and this PR has only the Dockerfile:

1. a registry entry in `internal/worker/profile.go` (`builtinProfiles`) so detection maps an ecosystem to the profile. For Rust that is `{Name: "rust", Ecosystem: "Cargo"}` (`brief` reports Cargo projects as `Cargo`).
2. a `docker/profiles/rust/PROFILE.md`, which the worker injects into the container as the per-ecosystem guide.
3. coverage in `internal/worker/profile_test.go` (the registry-sanity and detection tests).

Without the registry entry the worker never selects or builds the image, so the Dockerfile is inert today.

A few other things I noticed while building the other language profiles:

- rustup installs to `$HOME/.cargo/bin` and the PR only appends to `.bashrc`. The worker runs skills non-interactively with `HOME=/tmp`, so `.bashrc` is not sourced and `cargo`/`rustc` will not be on `PATH` at scan time. Installing with `CARGO_HOME`/`RUSTUP_HOME` under a fixed path and putting that on `ENV PATH` avoids it. That same `/tmp` is mounted `noexec`, so cargo's build cache and tmp dir want an exec-capable path too (I hit this with the Go and JVM profiles, where `cargo run`/`go test` execute the binaries they build).
- the apt list looks like it came from the ruby profile (`libyaml-dev`, `libreadline-dev`, `libgmp-dev`, `libgdbm-dev`, `libncurses-dev`); Rust does not need most of those.
- rustup is installed via an unpinned `curl | bash`, which drifts from how the other profiles pin and checksum their toolchains.

I am happy to push the registry entry, `PROFILE.md`, tests, and the `PATH`/cache fixes straight onto this branch so it lands as one complete profile, if you would rather not chase all of it. Just say the word and I will add it in.

docker/profiles/rust/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+USER runner
+
+# Install rust
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+RUN echo 'source $HOME/.cargo/env' >> $HOME/.bashrc
+# Install nightly
+RUN . $HOME/.cargo/env && rustup toolchain install nightly && rustup default nightly
+# Install tools and associated
+RUN . $HOME/.cargo/env && \
+       rustup component add miri
```
Eh2406: Miri also depends on the source of STD being available.

There's also a one time setup step `cargo +nightly miri setup`, that we should not make every runner rediscover.

Eh2406:
When we tell the AI to make sure to use the tools that are available, we should remember to point out that MIRI can emulate running on other targets.

walterhpearce:
I've addressed the majority of issues mentioned in here in the latest commits. I have 1 open question however - When pinning Rust nightly - what kind of strategy do you want for building? I've currently pinned it to a specific day, but we can also pull latest with our tools available. The nightly and stable toolchains are needed because features like ASAN and miri need nightly currently.

andrew:
Pushed fa1322c with the last review fixes plus two things I hit while actually building the image:

- The miri layer referenced `nightly-x86_64-unknown-linux-gnu` and `+nightly`, but only the dated `nightly-2026-06-24` toolchain is installed. rustup treats those as different toolchains, so the build either fetched an unpinned nightly or failed. Folded miri into the dated install and run setup via `+${RUST_NIGHTLY_VERSION}`, which also drops the hardcoded `x86_64` so the multi-arch case block above it actually holds.
- `gcc-multilib` has no installation candidate on arm64 Debian (or any non-amd64 arch), so the apt step failed outright there. Dropped it; cross-building to i686 from inside the scanner is niche enough not to be worth breaking five of the six supported arches.
- `cargo miri setup` was writing the sysroot to `/root/.cache/miri`, which the `runner` user never sees (and the worker sets `HOME=/tmp`, mounted noexec, so a scan-time rebuild would fail too). Pinned `MIRI_SYSROOT=/usr/local/miri-sysroot` and chmod'd it along with `RUSTUP_HOME`/`CARGO_HOME` after the nightly install.
- Updated `PROFILE.md` to select nightly via `$RUST_NIGHTLY_VERSION` everywhere, added a line about `miri --target` for foreign-arch UB per @Eh2406's note, and moved the `.crate` = gzipped tarball hint here from `SKILL.md` (it was sitting between two unrelated steps in the cross-language doc; this profile is where the runner actually reads it for Cargo projects).
- `nolint:maintidx` on `TestMatchProfile`: the three new cases pushed the table over the repo's configured threshold. Same justification pattern as `internal/web/repo_delete_test.go`.
- Trailing newline on the Dockerfile.

Verified the image builds on arm64 and that `cargo +$RUST_NIGHTLY_VERSION miri run` works as the `runner` user with `HOME=/tmp` without rebuilding the sysroot.

On the nightly-pinning question: a fixed date is the right call and matches the other profiles (ruby pins `RUBY_VERSION` + the ruby-build tag, python pins the uv release and CPython build). The worker tags profile images by the Dockerfile's content sha (`internal/worker/profile.go`, `EnsureImage`), so bumping `RUST_NIGHTLY_VERSION` in a PR is enough to invalidate the cache and rebuild. We bump manually when miri or a sanitizer needs a newer compiler; no need to pull latest at build time.

I think this is ready now. @Muscraft, if you have a moment to sanity-check the rustup-init pinning against what the official images do, that would be appreciated.

andrew (approved):
Image builds on amd64 and arm64, miri runs as the runner user with the pre-built sysroot, tests and lint pass. Thanks @walterhpearce, @Muscraft and @Eh2406.

PR #352, opened 2026-06-10
Split dependency phases in dependency views

### LINKED ISSUE #44: Dependencies skill should split test/build/runtime prereqs
Surfaced while validating #28 (Perl coverage).

Perl distributions separate runtime dependencies from `test_requires`, `build_requires`, and `configure_requires` in `META.yml`/`META.json`. Today the \`dependencies\` skill merges them all — e.g. `Test::More` shows up alongside genuine runtime modules like `File::Spec`.

The same concern applies to other ecosystems with the same shape: npm `devDependencies` vs `dependencies`, Python `[project.optional-dependencies].test` vs `[project.dependencies]`, Gemfile groups, Cargo `[dev-dependencies]`, etc.

- [ ] Extend \`dependencies\` skill to record the prereq phase (`runtime`/`test`/`build`/`dev`)
- [ ] Add a column on \`Dependency\` for the phase (or repurpose \`dependency_type\`)
- [ ] Default UI filter hides test/build rows; toggle to include them

abhinavgautam01:
Closes #44

## Summary

Splits dependency phases so runtime dependencies are handled separately from test, build and development prerequisites.

## Changes

- Normalize dependency phase/type values to `runtime`, `dev`, `test` or `build`
- Preserve unknown dependency types instead of hiding them
- Keep test/build/dev dependencies in dependency-finding reachability matching (they are still supply-chain edges); the phase is returned to the caller instead
- Show runtime dependencies by default on the repository Dependencies tab
- Add a toggle to include test/build/dev dependency rows
- Update dependency skill docs/schema, OpenAPI, README and database docs
- Add regression tests for parser normalization, dependency matching and UI filtering

## Testing

- `go test ./...`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ Each finding page has a notes section for recording triage reasoning and communi
 
 ## Exploring dependencies
 
-The Dependencies tab on a repo groups packages by name and shows all manifest files where each appears. The import button (arrow icon) next to a dependency resolves it to a repository URL via packages.ecosyste.ms and queues the full pipeline for it. Dependencies you've already imported show a link icon instead.
+The Dependencies tab on a repo groups packages by name and shows all manifest files where each appears. It shows runtime dependencies by default, with a toggle for test/build/dev rows. The import button (arrow icon) next to a dependency resolves it to a repository URL via packages.ecosyste.ms and queues the full pipeline for it. Dependencies you've already imported show a link icon instead.
 
 The same applies to the Dependents tab -- you can import any dependent's repository with one click.
 
--- a/docs/database.md
+++ b/docs/database.md
@@ -234,7 +234,7 @@ Package dependencies discovered by the `dependencies` skill. Replaced wholesale 
 | ecosystem | text | PURL type, e.g. `gem`, `npm`, `golang`. Derived from `p_url` (or the source ecosystem string when no PURL was recorded). Indexed. |
 | p_url | text | Package URL. |
 | requirement | text | Version constraint from the manifest. |
-| dependency_type | text | `runtime` or `development`. |
+| dependency_type | text | Normalised dependency phase: `runtime`, `dev`, `test`, `build`, or an unrecognised source value kept verbatim. |
 | manifest_path | text | Which file declared this dependency. |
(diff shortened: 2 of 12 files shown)


andrew (changes requested):
The normalization and UI work looks good, and classifying at read time so existing rows bucket correctly without a migration is a nice touch.

One concern before merging: `DependencyFindings` now excludes dev/test/build dependencies from dependency-finding reachability entirely, with no toggle. #44 only asked for the phase column and a UI filter on the Dependencies tab. For build dependencies especially this seems wrong - a vulnerable or compromised build dependency is a classic supply chain attack vector, and those findings now silently disappear from the dependency-findings view. Could you either keep build deps in reachability matching, or split the reachability exclusion out into its own PR so it can be discussed separately?

Minor: in `groupDeps`, a package appearing as both runtime and test now groups into one row labeled runtime, so the `deps=all` view can't show that a dep is also used in tests. Not blocking.

abhinavgautam01:
> The normalization and UI work looks good, and classifying at read time so existing rows bucket correctly without a migration is a nice touch.
> 
> One concern before merging: `DependencyFindings` now excludes dev/test/build dependencies from dependency-finding reachability entirely, with no toggle. #44 only asked for the phase column and a UI filter on the Dependencies tab. For build dependencies especially this seems wrong - a vulnerable or compromised build dependency is a classic supply chain attack vector, and those findings now silently disappear from the dependency-findings view. Could you either keep build deps in reachability matching, or split the reachability exclusion out into its own PR so it can be discussed separately?
> 
> Minor: in `groupDeps`, a package appearing as both runtime and test now groups into one row labeled runtime, so the `deps=all` view can't show that a dep is also used in tests. Not blocking.

Thanks, agreed. i removed the reachability-time exclusion so `DependencyFindings` includes all dependency phases again, including build/test/dev, while keeping the Dependencies tab UI filter behavior unchanged.

I also updated the DB test to assert that a test-only dependency still matches dependency findings and preserves its `dependency_type`. Full test/lint suite is passing...

andrew (approved):

PR #353, opened 2026-06-10
Resolve Maven dependency property requirements

### LINKED ISSUE #46: Maven variable interpolation leaks into dependency requirements
Surfaced while validating #29 (Java coverage).

\`dependencies\` stores raw manifest values for Maven pom.xml, which means variable expressions flow through verbatim:

\`\`\`
org.modelmapper:modelmapper    ${project.version}
org.openjdk.jmh:jmh-core        ${jmh.version}
\`\`\`

The downstream effect is that version-range advisory matching cannot resolve whether a dep is actually vulnerable — the requirement string doesn't name a concrete version. The same shape happens in parent POMs that centralise versions.

- [ ] Have the \`dependencies\` skill resolve \`${...}\` against the project's \`<properties>\` and \`parent\` chain before persisting
- [ ] Where resolution is not possible, store the unresolved expression but tag the row so UI/advisory code can flag it
- [ ] Equivalent story likely applies to Gradle \`buildSrc\` version catalogues (\`libs.versions.toml\`) and gradle.properties — verify once Gradle coverage lands

abhinavgautam01:
Closes #46

## Summary

Resolves Maven dependency requirement placeholders like `${project.version}` and `${jmh.version}` before dependency rows are persisted.

## Changes

- Resolve Maven `${...}` requirement values from local `pom.xml` properties
- Follow local parent POMs, including Maven's default `../pom.xml`
- Preserve unresolved placeholders and mark them with `requirement_unresolved`
- Persist unresolved requirement state on dependency rows
- Expose `requirement_unresolved` through the skill API/OpenAPI
- Show an `unresolved` badge in the repository Dependencies tab
- Add regression tests for direct properties, parent `project.version`, unresolved placeholders and parser persistence

## Testing

- `go test -count=1 ./internal/worker -run 'TestDependenciesScript|TestParseDependencies'`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -264,6 +264,7 @@ Package dependencies discovered by the `dependencies` skill. Replaced wholesale 
 | ecosystem | text | PURL type, e.g. `gem`, `npm`, `golang`. Derived from `p_url` (or the source ecosystem string when no PURL was recorded). Indexed. |
 | p_url | text | Package URL. |
 | requirement | text | Version constraint from the manifest. |
+| requirement_unresolved | boolean | True when `requirement` still contains an unresolved manifest expression such as `${project.version}`. |
 | dependency_type | text | Normalised dependency phase: `runtime`, `dev`, `test`, `build`, or an unrecognised source value kept verbatim. |
 | manifest_path | text | Which file declared this dependency. |
 | manifest_kind | text | `manifest` or `lockfile`. |
--- a/docs/database.md
+++ b/docs/database.md
@@ -265,6 +265,7 @@ Package dependencies discovered by the `dependencies` skill. Replaced wholesale 
 | p_url | text | Package URL. |
 | requirement | text | Version constraint from the manifest. |
 | requirement_unresolved | boolean | True when `requirement` still contains an unresolved manifest expression such as `${project.version}`. |
+| requirement_resolution | text | Resolver tag for `requirement`, e.g. `resolved`, `unresolved_property`, `unresolved_env`, `unresolved_parent`, `unresolved_profile_gated`, or `unresolved_missing`. |
 | dependency_type | text | Normalised dependency phase: `runtime`, `dev`, `test`, `build`, or an unrecognised source value kept verbatim. |
 | manifest_path | text | Which file declared this dependency. |
 | manifest_kind | text | `manifest` or `lockfile`. |
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.26
 toolchain go1.26.4
 
 require (
+	github.com/git-pkgs/pom v0.1.4
 	github.com/git-pkgs/purl v0.1.12
 	github.com/git-pkgs/sbom v0.1.2
 	github.com/glebarez/sqlite v1.11.0
--- a/go.sum
+++ b/go.sum
@@ -5,6 +5,8 @@ github.com/dlclark/regexp2 v1.11.0 h1:G/nrcoOa7ZXlpoa/91N3X7mM3r8eIlMBBJZvsz/mxK
 github.com/dlclark/regexp2 v1.11.0/go.mod h1:DHkYz0B9wPfa6wondMfaivmHpzrQ3v9q8cnmRbL6yW8=
 github.com/dustin/go-humanize v1.0.1 h1:GzkhY7T5VNhEkwH0PVJgjz+fX1rhBrR7pRT3mDkpeCY=
 github.com/dustin/go-humanize v1.0.1/go.mod h1:Mu1zIs6XwVuF/gI1OepvI0qD18qycQx+mFykh5fBlto=
+github.com/git-pkgs/pom v0.1.4 h1:C6st+XSbF75eKuwfdkDZZtYHoTcaWRIEQYar5VtszUo=
+github.com/git-pkgs/pom v0.1.4/go.mod h1:ufdMBe1lKzqOeP9IUb9NPZ458xKV8E8NvuyBMxOfwIk=
 github.com/git-pkgs/purl v0.1.12 h1:qCskrEU1LWQhCkIVZd992W5++Bsxazvx2Cx1/65qCvU=
 github.com/git-pkgs/purl v0.1.12/go.mod h1:ofp4mHsR0cUeVONQaf33n6Wxg2QTEvtUdRfCedI8ouA=
 github.com/git-pkgs/sbom v0.1.2 h1:18LkdFRBG8jkWf5cclGvwam5GwM3o3lK9/qUmup15+0=
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -336,16 +336,20 @@ type Dependent struct {
 // Dependency is one package dependency discovered by the git-pkgs job.
 // Rows are replaced wholesale each time the job runs for a repository.
 type Dependency struct {
-	ID             uint `gorm:"primarykey"`
-	RepositoryID   uint `gorm:"index;not null"`
-	Name           string
-	Ecosystem      string `gorm:"index"`
-	PURL           string
-	Requirement    string
-	DependencyType string
-	ManifestPath   string
-	ManifestKind   string
-	CreatedAt      time.Time
+	ID           uint `gorm:"primarykey"`
+	RepositoryID uint `gorm:"index;not null"`
+	Name         string
+	Ecosystem    string `gorm:"index"`
+	PURL         string
+	Requirement  string
+	// RequirementUnresolved is true when Requirement still contains a
+	// manifest-level expression such as ${project.version}. Advisory matching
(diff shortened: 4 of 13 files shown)


andrew (changes requested):
Good implementation and the test coverage is appreciated. One thing to fix before merging:

`relativePath` in `parse_pom` comes from untrusted repo content and is joined with `os.path.join` with no containment check, so a malicious POM can point at something like `../../../../path/to/host/file.xml`. Property values from whatever parses get substituted into `requirement` and persisted into the report - a limited file-content leak. The docker runner contains this, but the local runner executes on the host. Please resolve the parent path with `os.path.realpath` and skip it unless it stays under the src root.

Minor: the test stub heredoc picked up stray tab indentation on the `array)` and `*)` case arms - harmless in shell but looks like an editing accident.

abhinavgautam01:
> Good implementation and the test coverage is appreciated. One thing to fix before merging:
> 
> `relativePath` in `parse_pom` comes from untrusted repo content and is joined with `os.path.join` with no containment check, so a malicious POM can point at something like `../../../../path/to/host/file.xml`. Property values from whatever parses get substituted into `requirement` and persisted into the report - a limited file-content leak. The docker runner contains this, but the local runner executes on the host. Please resolve the parent path with `os.path.realpath` and skip it unless it stays under the src root.
> 
> Minor: the test stub heredoc picked up stray tab indentation on the `array)` and `*)` case arms - harmless in shell but looks like an editing accident.

thanks, i updated the Maven property resolver to canonicalize POM paths with `os.path.realpath` and skip parent `relativePath` targets unless they remain under the `./src` root.

I also added a regression test for a POM that points its parent outside `src`; the outside property is not read and the requirement stays unresolved. Cleaned up the stray tab indentation in the test stub as well. Full test/lint suite is passing.

andrew (changes requested):
The parent-path containment fix looks good, thanks. But before this lands I'd like to swap the resolver itself: rather than maintaining a bespoke Python interpolator in `index.sh`, use [git-pkgs/pom](https://github.com/git-pkgs/pom), which was written for exactly this problem (its README cites #46).

`pom.ResolveLocal(ctx, path, opts)` does what the Python script does — walks the `<parent><relativePath>` chain on disk with no network access and `Lstat` containment — plus the cases it doesn't: `dependencyManagement`, `<scope>import</scope>` BOMs, profile activation, and missing-version fill. It's golden-tested against `mvn help:effective-pom` output for 72 real POMs, including the artifacts from #46. Nested `${${a}}` style properties resolve instead of being flagged.

Two ways to wire it in, either is fine:

- import the library where the worker parses the skill output (it's stdlib-only Go, fits the existing `git-pkgs/purl`/`sbom`/`vers` deps), or
- install the `pom` CLI in the runner image and have `index.sh` pipe each `pom.xml` through `pom -f -` instead of the Python step

Please keep the `requirement_unresolved` plumbing from this PR — that part is good. One improvement to consider while you're there: `pom` returns a per-dependency `resolution` tag (`unresolved_property`, `unresolved_env`, `unresolved_profile_gated`, ...), which is more useful to surface than a bare boolean. Storing the tag would let the UI say why a requirement is unresolved.

abhinavgautam01:
> The parent-path containment fix looks good, thanks. But before this lands I'd like to swap the resolver itself: rather than maintaining a bespoke Python interpolator in `index.sh`, use [git-pkgs/pom](https://github.com/git-pkgs/pom), which was written for exactly this problem (its README cites #46).
> 
> `pom.ResolveLocal(ctx, path, opts)` does what the Python script does — walks the `<parent><relativePath>` chain on disk with no network access and `Lstat` containment — plus the cases it doesn't: `dependencyManagement`, `<scope>import</scope>` BOMs, profile activation, and missing-version fill. It's golden-tested against `mvn help:effective-pom` output for 72 real POMs, including the artifacts from #46. Nested `${${a}}` style properties resolve instead of being flagged.
> 
> Two ways to wire it in, either is fine:
> 
> * import the library where the worker parses the skill output (it's stdlib-only Go, fits the existing `git-pkgs/purl`/`sbom`/`vers` deps), or
> * install the `pom` CLI in the runner image and have `index.sh` pipe each `pom.xml` through `pom -f -` instead of the Python step
> 
> Please keep the `requirement_unresolved` plumbing from this PR — that part is good. One improvement to consider while you're there: `pom` returns a per-dependency `resolution` tag (`unresolved_property`, `unresolved_env`, `unresolved_profile_gated`, ...), which is more useful to surface than a bare boolean. Storing the tag would let the UI say why a requirement is unresolved.

thanks, updated this to use `github.com/git-pkgs/pom` in the worker dependency parser instead of the bespoke Python interpolator in `index.sh`.

The script now only wraps `git-pkgs list` output. Maven requirement enrichment happens while parsing the dependencies report, using `pom.ResolveLocal` against the scan workspace's local `pom.xml` files with no network access.

I kept the existing `requirement_unresolved` plumbing and added `requirement_resolution` so unresolved rows can explain why they were unresolved (`unresolved_property`, `unresolved_parent`, etc.). The dependencies API/schema/docs and repo UI tooltip now surface that tag.

kept regression coverage for parent paths escaping `src`, plus tests for resolved parent/project properties and unresolved property tags.

also rebased this branch on latest main and resolved the conflicts.

Tested with:

- `go test ./internal/worker -run 'TestDependenciesScript|TestParseDependencies'`
- `go test ./...`
- `go vet ./...`
- `env GOTOOLCHAIN=local /Users/gautam/go/bin/golangci-lint run ./...`

andrew (changes requested):
Thanks, the swap to `git-pkgs/pom` is exactly what I was after, and storing the resolution tag is a nice touch. The containment tests are good. A few things before this lands:

The `internal/worker/skill.go` refactor (`filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) is unrelated to this PR and conflicts with #387, which already landed the same extraction on main. Please drop that hunk and rebase. I left the same note on #359; #354 and #355 carry it too.

On `localPOMParentsStayUnderRoot`: this pre-walks the parent chain and parses every POM to check containment, then `ResolveLocal` parses them all again. The library already does `Lstat` containment internally; does `mavenpom.Options` expose a root/boundary field that would let it enforce the src-root limit itself? If so the pre-walk is redundant and we can drop ~40 lines. If not, what you have is fine as defence in depth.

Two bits of scope creep that I'd rather see reverted to keep the diff focused:

- `index.sh` now writes to a tmpfile and reads it back via `open(sys.argv[1])`. That was scaffolding for the Python resolver that's now gone; the original stdin pipe is simpler.
- The two `//nolint:dupl` directives in `api_reads.go`, one on `apiListDependents` which this PR doesn't otherwise touch. If `dupl` is firing on those handlers that's a separate conversation.

abhinavgautam01:
> Thanks, the swap to `git-pkgs/pom` is exactly what I was after, and storing the resolution tag is a nice touch. The containment tests are good. A few things before this lands:
> 
> The `internal/worker/skill.go` refactor (`filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) is unrelated to this PR and conflicts with #387, which already landed the same extraction on main. Please drop that hunk and rebase. I left the same note on #359; #354 and #355 carry it too.
> 
> On `localPOMParentsStayUnderRoot`: this pre-walks the parent chain and parses every POM to check containment, then `ResolveLocal` parses them all again. The library already does `Lstat` containment internally; does `mavenpom.Options` expose a root/boundary field that would let it enforce the src-root limit itself? If so the pre-walk is redundant and we can drop ~40 lines. If not, what you have is fine as defence in depth.
> 
> Two bits of scope creep that I'd rather see reverted to keep the diff focused:
> 
> * `index.sh` now writes to a tmpfile and reads it back via `open(sys.argv[1])`. That was scaffolding for the Python resolver that's now gone; the original stdin pipe is simpler.
> * The two `//nolint:dupl` directives in `api_reads.go`, one on `apiListDependents` which this PR doesn't otherwise touch. If `dupl` is firing on those handlers that's a separate conversation.

thanks, fixed...

I rebased this branch onto latest main and dropped the unrelated internal/worker/skill.go parser refactor, so this PR no longer conflicts with the #387/reobserveFinding work.

I checked github.com/git-pkgs/pom v0.1.4: mavenpom.Options only exposes Profiles and the local fetcher Lstats/rejects symlinks and absolute parent paths but does not expose a src-root boundary for preventing ../../ traversal outside Scrutineer’s checkout. I kept the localPOMParentsStayUnderRoot pre-walk as defence in depth for that case.

i also restored index.sh to the simpler stdin pipe and removed the added dupl suppressions. To keep CI green without nolint, apiListDependencies now uses a typed response struct for the dependency projection.

andrew (approved):

PR #354, opened 2026-06-10
Retry invalid skill reports with schema feedback

### LINKED ISSUE #261: Improve harness for ensuring that returned JSON blobs from jobs match schema
Right now, I have a lot of jobs (maintainers, metadata, occasionally others) that fail with errors like `parse repo_metadata: json: cannot unmarshal object into Go struct field .languages of type []string` or `parse maintainers report: json: cannot unmarshal array into Go struct field .maintainers.email of type string`.

It'd be good if we had a step where the JSON blob that the LLM emits is tested against the JSON Schema, and then immediately resumes the relevant session with feedback to the LLM about the JSON Schema discrepancy if there is any so it can fix its mistake.

abhinavgautam01:
## Summary

Fixes #261.

Adds a schema repair retry for skill runs when `report.json` does not validate against the skill's `schema.json`.

## Changes

- Validate skill output before parsing and detect schema mismatches early
- Resume the existing Claude session once with the schema validation errors
- Ask Claude to rewrite only `report.json` using the validator feedback
- Preserve existing behavior after the retry:
  - default mode logs the schema mismatch and continues parsing
  - `-schema-strict` still fails the scan if the repaired output is invalid
- Add tests for successful schema repair and strict-mode repair failure
- Document the schema repair behavior in `docs/skills.md`

## Testing

- `go test ./internal/worker`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `golangci-lint run ./...`

### DIFF
--- a/docs/skills.md
+++ b/docs/skills.md
@@ -193,7 +193,7 @@ The worker then runs `claude -p "Use the {name} skill in this workspace"` with t
 
 ## schema.json
 
-If a `schema.json` sits next to `SKILL.md`, scrutineer stages it into the workspace so the model can read the expected output shape, and after the run validates `report.json` against it with a draft 2020-12 validator. By default a mismatch is logged to the scan transcript and the parser still runs, so a stricter schema does not break ingestion. Start scrutineer with `-schema-strict` (or `schema_strict: true` in the config file) to turn that warning into a scan failure with the validator output in `Scan.Error`; useful while iterating on a skill locally.
+If a `schema.json` sits next to `SKILL.md`, scrutineer stages it into the workspace so the model can read the expected output shape, and after the run validates `report.json` against it with a draft 2020-12 validator. When validation fails and the Claude session is still available, scrutineer resumes that session once with the validator output and asks it to rewrite `report.json` before parsing. If the repaired report still fails validation, the mismatch is logged to the scan transcript and the parser still runs by default, so a stricter schema does not break ingestion. Start scrutineer with `-schema-strict` (or `schema_strict: true` in the config file) to turn the remaining warning into a scan failure with the validator output in `Scan.Error`; useful while iterating on a skill locally.
 
 Bundled skills with typed output kinds carry a schema; skills with `output_kind: freeform` generally do not.
 
--- a/internal/worker/claude.go
+++ b/internal/worker/claude.go
@@ -75,6 +75,10 @@ type SkillJob struct {
 	// conversation with full history instead of restarting from turn 0.
 	// The runner falls back to a fresh run if the session can't be found.
 	ResumeSessionID string
+	// ResumePrompt, when non-empty, replaces the default generic resume
+	// prompt. It lets callers resume the same conversation with targeted
+	// corrective instructions, such as rewriting an invalid report.json.
+	ResumePrompt string
 	// ClaudeConfigDir is a host directory the docker runner mounts as the
 	// container's CLAUDE_CONFIG_DIR so the resumable session store persists
 	// across container restarts. Empty disables the mount (the local runner
@@ -266,7 +270,11 @@ func buildClaudeArgs(sj SkillJob, effort string, globalMaxTurns int) []string {
 	}
 	args = append(args, "--max-turns", strconv.Itoa(effectiveMaxTurns(sj.MaxTurns, globalMaxTurns)))
 	if sj.ResumeSessionID != "" {
-		args = append(args, buildResumePrompt(sj.Name, sj.OutputFile))
+		if sj.ResumePrompt != "" {
+			args = append(args, sj.ResumePrompt)
+		} else {
+			args = append(args, buildResumePrompt(sj.Name, sj.OutputFile))
+		}
 	} else {
 		args = append(args, buildSkillPrompt(sj.Name, sj.OutputFile))
 	}
--- a/internal/worker/claude.go
+++ b/internal/worker/claude.go
@@ -17,6 +17,8 @@ import (
 // metadata nor the global config set a value.
 const DefaultSkillMaxTurns = 30
 
+const resumePromptNoFreshFallbackText = "not restarting repair prompt fresh"
+
 // MaxTurnsReachedError is returned when claude-code exits after hitting the
 // --max-turns cap. The caller should treat this as a soft completion.
 type MaxTurnsReachedError struct{}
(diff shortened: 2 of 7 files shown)


andrew (changes requested):
Nice approach and good test coverage, but there's one behavior regression to fix:

In `repairSchemaReport`, if the repair `RunSkill` call itself errors (CLI flake, timeout, max-turns), the error propagates and the whole scan fails. Before this PR, a non-strict scan with an invalid report still parsed with a warning - the docs say "a stricter schema does not break ingestion", but now the optional repair attempt can break it. The repair run should be best-effort: on error, log it and fall back to parsing the original report (at least when `-schema-strict` is off).

Two smaller things:

- The mismatch gets logged three times on a failed repair: once in `repairSchemaReport` before the attempt, again after, and a third time when `parseSkillOutput` re-validates. Worth deduplicating.
- The emitted messages hardcode `report.json` via `defaultSkillOutputFile`, while the prompt correctly uses `skill.OutputFile`. Skills with a different output file would log the wrong name.

abhinavgautam01:
> Nice approach and good test coverage, but there's one behavior regression to fix:
> 
> In `repairSchemaReport`, if the repair `RunSkill` call itself errors (CLI flake, timeout, max-turns), the error propagates and the whole scan fails. Before this PR, a non-strict scan with an invalid report still parsed with a warning - the docs say "a stricter schema does not break ingestion", but now the optional repair attempt can break it. The repair run should be best-effort: on error, log it and fall back to parsing the original report (at least when `-schema-strict` is off).
> 
> Two smaller things:
> 
> * The mismatch gets logged three times on a failed repair: once in `repairSchemaReport` before the attempt, again after, and a third time when `parseSkillOutput` re-validates. Worth deduplicating.
> * The emitted messages hardcode `report.json` via `defaultSkillOutputFile`, while the prompt correctly uses `skill.OutputFile`. Skills with a different output file would log the wrong name.

thanks, i updated the repair path so the schema repair run is best-effort: if the repair RunSkill call errors, it logs the repair failure and falls back to parsing the original report, preserving non-strict ingestion behavior.

I also deduplicated the schema validation logging so failed repairs only emit the validator detail once and changed the schema messages to use the skill's configured output file instead of hardcoded report.json. added tests for the repair-error fallback, deduped logging and custom output filenames.

andrew (changes requested):
The best-effort rework addresses my earlier points, thanks. Two issues with how the repair run interacts with existing worker machinery:

1. Dead-session fallback defeats the repair. In `claude.go` the existing fallback for an expired session clears `ResumeSessionID` and rebuilds the args, and that path ignores `ResumePrompt` and uses the full skill prompt. So when the saved session has expired, the repair attempt silently becomes a complete second skill run from turn 0 at full token cost. When `ResumePrompt` is set, skip the fallback (give up on repair and parse the original output) or carry the repair prompt into the rebuilt invocation.

2. Cost accounting drops the original run. The repair invocation emits a second `KindResult`, and `scanEmitter` assigns `CostUSD`, `Turns` and the token counts rather than accumulating them, so a repaired scan records only the repair invocation's numbers. The original run's cost disappears from the scan row and the Usage page. Accumulate when handling the repair result, with a test asserting the totals cover both invocations.

Two smaller things while you're in there, not blocking: `schemaRepairMaxAttempts = 1` as a loop bound reads oddly and the mid-loop "parsing original output" message would land in the wrong place if the constant ever grew, so a plain single attempt would be clearer; and the repair job inherits the skill's full `MaxTurns` when a report rewrite needs only a handful of turns.

abhinavgautam01:
> The best-effort rework addresses my earlier points, thanks. Two issues with how the repair run interacts with existing worker machinery:
> 
> 1. Dead-session fallback defeats the repair. In `claude.go` the existing fallback for an expired session clears `ResumeSessionID` and rebuilds the args, and that path ignores `ResumePrompt` and uses the full skill prompt. So when the saved session has expired, the repair attempt silently becomes a complete second skill run from turn 0 at full token cost. When `ResumePrompt` is set, skip the fallback (give up on repair and parse the original output) or carry the repair prompt into the rebuilt invocation.
> 2. Cost accounting drops the original run. The repair invocation emits a second `KindResult`, and `scanEmitter` assigns `CostUSD`, `Turns` and the token counts rather than accumulating them, so a repaired scan records only the repair invocation's numbers. The original run's cost disappears from the scan row and the Usage page. Accumulate when handling the repair result, with a test asserting the totals cover both invocations.
> 
> Two smaller things while you're in there, not blocking: `schemaRepairMaxAttempts = 1` as a loop bound reads oddly and the mid-loop "parsing original output" message would land in the wrong place if the constant ever grew, so a plain single attempt would be clearer; and the repair job inherits the skill's full `MaxTurns` when a report rewrite needs only a handful of turns.

fixed both issues.

When a repair resume uses ResumePrompt and the saved session is gone, the local and Docker runners now stop the repair attempt instead of falling back to a fresh full skill run. That lets the existing best-effort path parse the original output.

I also changed scan result handling to accumulate cost, turns, and token counts across the original run and repair invocation, and added a regression test covering the combined totals. While there, I simplified the repair path to one explicit attempt and capped repair runs with a smaller MaxTurns value.

Tested with:
- go test ./internal/worker -run 'TestLocalClaude_Resume|TestDoSkill_schema|TestScanEmitter'
- go test ./...
- go vet ./...
- env GOTOOLCHAIN=local /Users/gautam/go/bin/golangci-lint run ./...

andrew (changes requested):
All four points from the last round are addressed, thanks. The dead-session bail-out, cost accumulation, single-attempt shape and capped repair turns all look right and the tests cover them well.

One thing left: the `parseFindingsOutput` refactor at the bottom of `skill.go` (`filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) conflicts with #387, which already landed the same extraction on main. See my note on #359; same applies here. Drop just that hunk and rebase, the rest of `skill.go` (the repair plumbing) stays.

abhinavgautam01:
> All four points from the last round are addressed, thanks. The dead-session bail-out, cost accumulation, single-attempt shape and capped repair turns all look right and the tests cover them well.
> 
> One thing left: the `parseFindingsOutput` refactor at the bottom of `skill.go` (`filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) conflicts with #387, which already landed the same extraction on main. See my note on #359; same applies here. Drop just that hunk and rebase, the rest of `skill.go` (the repair plumbing) stays.

i rebased the branch onto latest main and dropped the unrelated parseFindingsOutput refactor hunk. The schema repair changes remain, while skill.go now keeps main’s #387/reobserveFinding structure instead of carrying the conflicting helper extraction.

andrew (approved):

PR #355, opened 2026-06-10
Hide closed findings by default

### LINKED ISSUE #244: Hide/separate fixed or rejected findings
Right now the findings tab can get pretty noisy if you run the scanner on a large repository where it may find various things. Once an issue is Fixed or Rejected explicitly, it'd be good to clean it out of the Findings tab, or at least put it in its own section near the bottom of the list. Otherwise it can be hard to figure out what findings still need investigation.

abhinavgautam01:
## Summary

Fixes #244.

Hides closed findings from the default Findings views so fixed/rejected/resolved noise does not crowd out items that
still need investigation.

## Changes

- Hide `fixed`, `published`, `rejected` and `duplicate` findings by default on the global Findings index
- Keep explicit status filters working, so closed findings can still be viewed with filters like `?status=fixed`
- Hide the same closed statuses from the repository Findings tab
- Update repository Findings tab copy to describe the hidden statuses
- Add tests for default hiding and explicit status filtering

## Testing

- `go test ./internal/web`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `golangci-lint run ./...`

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -296,6 +296,19 @@ var FindingLifecycles = []FindingLifecycle{
 	FindingAcknowledged, FindingFixed, FindingPublished, FindingRejected, FindingDuplicate,
 }
 
+// ClosedFindingLifecycles are terminal or hidden-by-default findings.
+var ClosedFindingLifecycles = []FindingLifecycle{
+	FindingFixed, FindingPublished, FindingRejected, FindingDuplicate,
+}
+
+func ClosedFindingLifecycleSQLValues() string {
+	values := make([]string, 0, len(ClosedFindingLifecycles))
+	for _, status := range ClosedFindingLifecycles {
+		values = append(values, "'"+strings.ReplaceAll(string(status), "'", "''")+"'")
+	}
+	return strings.Join(values, ", ")
+}
+
 // Advisory is a known security advisory from advisories.ecosyste.ms.
 type Advisory struct {
 	ID           uint `gorm:"primarykey"`
--- a/internal/db/ecosystem.go
+++ b/internal/db/ecosystem.go
@@ -206,7 +206,7 @@ func DependencyFindings(g *gorm.DB, appRepoID uint) ([]DependencyFinding, error)
 	}
 	var findings []Finding
 	if err := g.Where("repository_id IN ?", libIDs).
-		Where("status NOT IN ?", []FindingLifecycle{FindingFixed, FindingRejected, FindingDuplicate}).
+		Where("status NOT IN ?", ClosedFindingLifecycles).
 		Order("CASE severity WHEN 'Critical' THEN 0 WHEN 'High' THEN 1 WHEN 'Medium' THEN 2 ELSE 3 END, repository_id").
 		Find(&findings).Error; err != nil {
 		return nil, err
--- a/internal/web/sboms.go
+++ b/internal/web/sboms.go
@@ -136,8 +136,7 @@ func (s *Server) sbomShow(w http.ResponseWriter, r *http.Request) {
 		// Deep-dive findings only — the SBOM "Findings" tab is a
 		// downstream-impact view, where lint output from per-repo scanners
 		// (zizmor, semgrep) would be misleading at this level.
-		q := s.DB.Where("repository_id IN ? AND status NOT IN ?", repoIDs,
-			[]db.FindingLifecycle{db.FindingRejected, db.FindingDuplicate}).
+		q := s.DB.Where("repository_id IN ? AND status NOT IN ?", repoIDs, closedFindingStatuses).
 			Where("scan_id IN (?)", deepDiveScanIDs(s.DB))
 		if sev := r.URL.Query().Get("severity"); sev != "" {
 			q = q.Where("severity = ?", sev)
--- a/internal/web/sboms.go
+++ b/internal/web/sboms.go
@@ -136,7 +136,7 @@ func (s *Server) sbomShow(w http.ResponseWriter, r *http.Request) {
 		// Deep-dive findings only — the SBOM "Findings" tab is a
 		// downstream-impact view, where lint output from per-repo scanners
 		// (zizmor, semgrep) would be misleading at this level.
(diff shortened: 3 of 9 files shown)


andrew (changes requested):
This looks good overall, but the SBOM findings tab was missed: `internal/web/sboms.go` still filters with the inline `[]db.FindingLifecycle{db.FindingRejected, db.FindingDuplicate}` list, so fixed/published findings will still show up there. Could you switch it to `closedFindingStatuses` so all the findings views behave the same way?

abhinavgautam01:
> This looks good overall, but the SBOM findings tab was missed: `internal/web/sboms.go` still filters with the inline `[]db.FindingLifecycle{db.FindingRejected, db.FindingDuplicate}` list, so fixed/published findings will still show up there. Could you switch it to `closedFindingStatuses` so all the findings views behave the same way?

thanks, i switched the SBOM findings query to use `closedFindingStatuses` and expanded the SBOM aggregation test to cover fixed, published, rejected and duplicate findings. Also reran the full test/lint suite.

andrew (changes requested):
The sboms.go fix looks good. One gap left:

With the default now excluding closed findings, the "All statuses" dropdown option in `findings.html` still links to `/findings` with no `status` param, which now means "open only". The label is wrong, and there is no longer any way to list all findings (or all closed ones) in one view; closed findings are only reachable one status at a time. Add a `status=all` sentinel the handlers recognise and point the dropdown option at it, or relabel the default option "Open" and add an "All" option. #244 also floated a separate closed section near the bottom of the list as an alternative if that turns out nicer.

Two smaller things, not blocking:

- The `MissedTotal`/`ScannerTotal` toggle badges now exclude closed findings even when an explicit closed-status filter is active (`?status=fixed&missed=1`), so the badge count can disagree with the list it sits next to.
- The closed-status set is defined in `internal/web` while `internal/db/ecosystem.go` keeps its own inline list (which omits `published`) and `deepDiveFindingsCountSQL` hardcodes the four statuses in raw SQL. One shared set in `internal/db` would stop these drifting.

abhinavgautam01:
> The sboms.go fix looks good. One gap left:
> 
> With the default now excluding closed findings, the "All statuses" dropdown option in `findings.html` still links to `/findings` with no `status` param, which now means "open only". The label is wrong, and there is no longer any way to list all findings (or all closed ones) in one view; closed findings are only reachable one status at a time. Add a `status=all` sentinel the handlers recognise and point the dropdown option at it, or relabel the default option "Open" and add an "All" option. #244 also floated a separate closed section near the bottom of the list as an alternative if that turns out nicer.
> 
> Two smaller things, not blocking:
> 
> * The `MissedTotal`/`ScannerTotal` toggle badges now exclude closed findings even when an explicit closed-status filter is active (`?status=fixed&missed=1`), so the badge count can disagree with the list it sits next to.
> * The closed-status set is defined in `internal/web` while `internal/db/ecosystem.go` keeps its own inline list (which omits `published`) and `deepDiveFindingsCountSQL` hardcodes the four statuses in raw SQL. One shared set in `internal/db` would stop these drifting.

Thanks, fixed the remaining gap.

The findings index now treats the default status view as Open and adds a real `status=all` option for listing all statuses. The dropdown points “All statuses” at that sentinel instead of `/findings` with no status param.

I also made the missed/scanner badge counts reuse the same filtered query shape as the list, so explicit closed-status filters like `?status=fixed&missed=1` count the matching closed findings correctly. While there, I moved the closed finding lifecycle set into `internal/db` and reused it across the web views, dependency findings, SBOM findings, repo count SQL and worker missed-count logic to avoid drift.

Tested with:
- go test ./internal/web -run 'TestFindings_statusFilter|TestFindings_missedFilterAndBadge|TestFindings_scannerToggle|TestSBOMShow_aggregatesFindings|TestRepoList_batchedFindingsCountAcrossRepos|TestRepoList_findingsCountIgnoresClosedFindings'
- go test ./internal/db -run 'Test|Dependency'
- go test ./...
- go vet ./...
- env GOTOOLCHAIN=local /Users/gautam/go/bin/golangci-lint run ./...

andrew (changes requested):
Thanks, the `status=all` sentinel, badge-count fix and shared `db.ClosedFindingLifecycles` set all look right and the test coverage is good.

One thing left: the `parseFindingsOutput` refactor in `internal/worker/skill.go` (`filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) conflicts with #387, which already landed the same extraction on main. See my note on #359; same applies here. Drop just that hunk, the `markNotObserved` change in the same file is part of this PR and should stay.

Heads-up for the rebase: #392 just landed and added `ghsa_id` to the full-text search clause this PR moves into `findingsIndexQuery`. Make sure it survives the move; git can lose edits to relocated lines.

abhinavgautam01:
> Thanks, the `status=all` sentinel, badge-count fix and shared `db.ClosedFindingLifecycles` set all look right and the test coverage is good.
> 
> One thing left: the `parseFindingsOutput` refactor in `internal/worker/skill.go` (`filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) conflicts with #387, which already landed the same extraction on main. See my note on #359; same applies here. Drop just that hunk, the `markNotObserved` change in the same file is part of this PR and should stay.
> 
> Heads-up for the rebase: #392 just landed and added `ghsa_id` to the full-text search clause this PR moves into `findingsIndexQuery`. Make sure it survives the move; git can lose edits to relocated lines.

Thanks, fixed.

I rebased this branch onto latest main and dropped the unrelated parseFindingsOutput helper extraction from internal/worker/skill.go. The markNotObserved change remains.

i also resolved the #392 overlap by keeping ghsa_id in the moved findingsIndexQuery search clause. While testing, TestListPagesQueryCountsStayBounded caught an extra /findings query from the badge-count changes, so i combined the missed/scanner badge counts into one query while preserving the same filters.

andrew (approved):

PR #359, opened 2026-06-11
Add configurable model tiers

### LINKED ISSUE #330: Add a concept of a model "tiers" in Settings
I've been thinking about this for a little while and I think it's a decent enough idea to suggest for addition. If we intend to support tooling other than Claude, it'd probably make sense to have something like different "Tiers".

For example, a mid-tier model would be Sonnet 4.6 and a high-tier model would be Opus 4.8. Then we can more easily have the user swap Sonnet for GPT 5.3 (or whatever) and Opus for GPT 5.5, by setting the models to the specific tiers.

I'd also maybe argue for security-deep-dive getting its own model preference/tier (probably Max?). I don't care much if Opus 4.6 or 4.8 is used for most skills, but for security-deep-dive I want it to always use the best model available to me.

You could configure which is the default mid-tier and high-tier models in the Scrutineer settings, and then we can assign skills to default "tiers", e.g. metadata gets assigned mid-tier and other skills default to high-tier.

The effort level unfortunately might make this phrasing confusing, but 🤷

abhinavgautam01:
## Summary

Fixes #330.

This PR adds configurable model tiers so skills can target roles like `mid`, `high` or `max` instead of hardcoding a specific model ID.

Changes include:

- add `mid`, `high` and `max` model tiers
- add Settings controls to map each tier to a configured model
- resolve scan models by precedence:
  - explicit scan model or tier
  - skill model or tier
  - skill default tier
- default `metadata` to `mid`
- default `security-deep-dive` to `max`
- default other skills to `high`
- keep exact model IDs backwards-compatible
- update bundled Sonnet-pinned skills to use the `mid` tier
- update docs and tests

## Testing

- `go test ./internal/web ./internal/skills -run 'TestModelTiers|TestDefaultModelTierForSkill|TestSettingsShow_rendersRunnerControls|TestSettingsUpdateModelTier|TestEnqueueSkillWith_modelPrecedence|TestParseSkillModel|TestParseFile_model|TestParseFile_modelTier|TestParseFile_modelInvalidIgnoredWithWarning'`
- `go test ./...`
- `go vet ./...`
- `env GOTOOLCHAIN=local /Users/gautam/go/bin/golangci-lint run ./...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Every index page has a search box plus filter and sort dropdowns; the specifics 
 - **Scans** -- every scan that has run. Queued scans can be paused/resumed, running or queued scans can be cancelled and failed ones retried.
 - **Skills** -- installed skills from disk and from the UI; view, edit, or run any of them.
 - **Usage** -- token and cost totals across all scans, broken down by skill.
-- **Settings** -- theme, colour scheme, default model, runner concurrency (restarts the runner to apply, cancelling in-flight scans) and default turn cap (applied to the next scan), plus system stats (record counts, DB size, paths).
+- **Settings** -- theme, colour scheme, model tiers, runner concurrency (restarts the runner to apply, cancelling in-flight scans) and default turn cap (applied to the next scan), plus system stats (record counts, DB size, paths).
 
 ## Finding workflow
 
@@ -242,7 +242,7 @@ For deployments that treat skill prompts as untrusted, pass `--hardened` (or `ha
 
 Every flag above can be set in a YAML config file instead. The loader checks `./scrutineer.yaml` by default; override with `-config path/to/file`. Command-line flags always win. See [scrutineer.sample.yaml](scrutineer.sample.yaml) for the full shape.
 
-The config file can also replace the model pick list and pin the default model:
+The config file can also replace the model pick list and pin the fallback default model used by the high tier:
 
     default_model: claude-sonnet-4-6
     models:
@@ -250,6 +250,8 @@ The config file can also replace the model pick list and pin the default model:
         id:   claude-sonnet-4-6
       - name: Opus
         id:   claude-opus-4-7
+
+Scrutineer resolves skill models through tiers. `metadata` defaults to the `mid` tier, `security-deep-dive` defaults to the `max` tier, and other skills default to `high`. The Settings page lets you map each tier to any configured model. A skill can still pin `scrutineer.model` to an exact model id, or to a tier name (`mid`, `high`, `max`).
 
 ## Sandboxed Claude Code configs
 
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ The config file can also replace the model pick list and pin the fallback defaul
       - name: Opus
         id:   claude-opus-4-7
 
-Scrutineer resolves skill models through tiers. `metadata` defaults to the `mid` tier, `security-deep-dive` defaults to the `max` tier, and other skills default to `high`. The Settings page lets you map each tier to any configured model. A skill can still pin `scrutineer.model` to an exact model id, or to a tier name (`mid`, `high`, `max`).
+Scrutineer resolves skill models through tiers. Skills default to the `high` tier unless their `SKILL.md` metadata pins `scrutineer.model` to another tier or exact model id. Bundled lightweight skills such as `metadata` use `mid`, while `security-deep-dive` uses `max`. The Settings page lets you map each tier to any configured model.
 
 ## Sandboxed Claude Code configs
 
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -285,7 +285,7 @@ func run(log *slog.Logger) error {
 		return fmt.Errorf("queue: %w", err)
 	}
 
-	skills.ModelValidator = web.ValidModel
+	skills.ModelValidator = web.ValidModelPreference
 	skills.ProfileValidator = worker.IsNamedProfile
 	skillsRepoSHA, err := loadSkills(log, gdb, f.dataDir, f.skillLocal, f.skillsRepo, f.fullClone())
 	if err != nil {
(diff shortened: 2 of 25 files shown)


andrew (changes requested):
The tier resolution logic and test coverage look good. A few things before this lands:

- `scrutineer.sample.yaml` wasn't updated; its comment still describes the old precedence ("per-skill wins over this default, and an explicit per-scan model wins over both") while README and docs/skills.md got the new wording. The sample config ships, so it needs the same update.
- `defaultModelTierForSkill` hardcodes `"metadata"` as a magic string while `security-deep-dive` goes through the existing `deepDiveSkillName` constant, and it would also match any user-created skill that happens to be named "metadata". The existing convention (see f443c57) is to pin per-skill model defaults in SKILL.md metadata; I'd rather the default tier live there too (`scrutineer.model: mid` or a new `scrutineer.model_tier` key) than in a Go switch, so custom skills get the same mechanism and the worker doesn't special-case bundled skill names. Happy to discuss if you hit a reason that doesn't work.
- `builtinModelForTier` falls back by substring-matching "sonnet"/"opus" and depends on `Models` ordering; if the model list is ever config-replaced (the multi-vendor case in #330) all three tiers silently collapse to `DefaultModel()`. Fine for now, but add a comment noting the assumption so it doesn't bite later.
- `models_test.go` hardcodes `claude-sonnet-4-6`/`claude-opus-4-8`, so the next addition to `Models` breaks the test (we add models fairly often, e.g. ebe63a5). Assert tier-resolution behaviour against a test-local model list instead of the live one.

abhinavgautam01:
> The tier resolution logic and test coverage look good. A few things before this lands:
> 
> * `scrutineer.sample.yaml` wasn't updated; its comment still describes the old precedence ("per-skill wins over this default, and an explicit per-scan model wins over both") while README and docs/skills.md got the new wording. The sample config ships, so it needs the same update.
> * `defaultModelTierForSkill` hardcodes `"metadata"` as a magic string while `security-deep-dive` goes through the existing `deepDiveSkillName` constant, and it would also match any user-created skill that happens to be named "metadata". The existing convention (see [f443c57](https://github.com/alpha-omega-security/scrutineer/commit/f443c572f5ce6ccf020e3a1a84a284f9a42493e3)) is to pin per-skill model defaults in SKILL.md metadata; I'd rather the default tier live there too (`scrutineer.model: mid` or a new `scrutineer.model_tier` key) than in a Go switch, so custom skills get the same mechanism and the worker doesn't special-case bundled skill names. Happy to discuss if you hit a reason that doesn't work.
> * `builtinModelForTier` falls back by substring-matching "sonnet"/"opus" and depends on `Models` ordering; if the model list is ever config-replaced (the multi-vendor case in #330) all three tiers silently collapse to `DefaultModel()`. Fine for now, but add a comment noting the assumption so it doesn't bite later.
> * `models_test.go` hardcodes `claude-sonnet-4-6`/`claude-opus-4-8`, so the next addition to `Models` breaks the test (we add models fairly often, e.g. [ebe63a5](https://github.com/alpha-omega-security/scrutineer/commit/ebe63a591a3706b02df819e09f82b77d6fe4b9a8)). Assert tier-resolution behaviour against a test-local model list instead of the live one.

thanks, fixxed these.

- Updated scrutineer.sample.yaml to match the README/docs model precedence wording.
- Removed the hardcoded skill-name tier switch. Bundled defaults now live in SKILL.md metadata: metadata uses `scrutineer.model: mid` and security-deep-dive uses `scrutineer.model: max`.
- Added a comment documenting that builtin tier fallback assumes Anthropic-style model ids/order and custom multi-vendor lists should configure tiers explicitly.
- Reworked the model tier tests to use a test-local model list instead of hardcoding the live Models entries.
- Rebased onto latest main and fixed the lint failure from parseFindingsOutput cognitive complexity.

Tested with go test ./..., go vet ./... and golangci-lint run ./...

andrew (changes requested):
Thanks, all four points from the last review are addressed and the tier work itself looks ready.

One blocker: the `internal/worker/skill.go` refactor (extracting `filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) is unrelated to model tiers and conflicts with #387, which already landed the same extraction on main as `reobserveFinding`. Please drop that hunk and rebase on main.

I see the same "Reduce findings parser complexity" commit on #353, #354 and #355, so those will hit the same conflict. Worth dropping it from all four in one go. If golangci-lint's complexity check is what prompted it, main should already satisfy it after #387.

Minor while you're in there:
- `server_test.go` in `TestEnqueueSkillWith_modelPrecedence` still hardcodes `claude-sonnet-4-6` / `claude-opus-4-7` / `claude-opus-4-8`. Not as fragile as the case I flagged before since you pin via Settings, but it'll still break the day one of those drops off the list. Using `withTestModels` here too would make it self-contained.
- Test case name typo: `"metadata metadata uses mid tier"`.

abhinavgautam01:
> Thanks, all four points from the last review are addressed and the tier work itself looks ready.
> 
> One blocker: the `internal/worker/skill.go` refactor (extracting `filterFindingsByMinConfidence` / `saveOrObserveFinding` / `recordFindingObservedAgain`) is unrelated to model tiers and conflicts with #387, which already landed the same extraction on main as `reobserveFinding`. Please drop that hunk and rebase on main.
> 
> I see the same "Reduce findings parser complexity" commit on #353, #354 and #355, so those will hit the same conflict. Worth dropping it from all four in one go. If golangci-lint's complexity check is what prompted it, main should already satisfy it after #387.
> 
> Minor while you're in there:
> 
> * `server_test.go` in `TestEnqueueSkillWith_modelPrecedence` still hardcodes `claude-sonnet-4-6` / `claude-opus-4-7` / `claude-opus-4-8`. Not as fragile as the case I flagged before since you pin via Settings, but it'll still break the day one of those drops off the list. Using `withTestModels` here too would make it self-contained.
> * Test case name typo: `"metadata metadata uses mid tier"`.

Thanks, fixed.

I dropped the unrelated internal/worker/skill.go parser refactor from this branch so the PR now keeps the #387/reobserveFinding version from main. The branch is also current with main.

I also updated TestEnqueueSkillWith_modelPrecedence to use a test-local model list via withTestModels instead of hardcoded Claude model IDs and fixed the "metadata metadata uses mid tier" test name typo.

andrew (approved):

PR #394, opened 2026-06-14
feat(web): import all repositories from a GitHub org

### LINKED ISSUE #131: Import all repositories from a GitHub org
Add an org-level import that fetches all repositories from a GitHub organization and queues them for scanning.

By default, skip forks and archived repos. Provide toggles to include them if needed.

connorshea:
Implements #131 — an org-level import that fetches every repository in a GitHub organization and queues them for scanning.

Screenshot 2026-06-14 at 9 52 34 AM
Screenshot 2026-06-14 at 9 52 40 AM
Screenshot 2026-06-14 at 9 52 56 AM

## What it does
- New **Import a whole org** dialog in the primary UI (Add repository → Add multiple → Import a whole org), plus the no-JS fallback page.
- Fetches every repo via `GET /orgs/{org}/repos`, paginating to completion; falls back to `/users/{org}` when the login is a user account.
- **Skips forks and archived repos by default**, with `Include forks` / `Include archived` toggles to opt in (per the issue).
- Reuses the existing `createOrTriageRepo` spine, so URL dedup and default-skill (`triage`) enqueue behave exactly like a manual add. Duplicates already in the DB are skipped.
- Honors an optional `GITHUB_TOKEN` to raise GitHub's unauthenticated rate limit for large orgs.
- Redirects to the canonical `/orgs/{owner}` page on success (falls back to `/` when nothing was imported), with a flash toast summarizing added / already-present / forks-and-archived-skipped counts.

## Implementation notes
- `internal/web/org_import.go` — fetcher + `repoOrgImport` handler.
- `server.go` — `fetchOrgRepos` stubbable field, `POST /repositories/org` route, and `repoCreateError` refactored into `repoFormError(alertTmpl, …)` so the org dialog gets its own inline error alert.
- Imports run synchronously within the request (same shape as bulk paste); the per-repo work is a cheap insert + queue enqueue, not the scan itself.

## Tests
- Handler: default skip, include toggles, dedup, blank-org 422, fetch-error 502, `@org` prefix stripping, dialog reachability from the primary UI.
- Fetcher (`httptest`-backed): pagination to completion, org→user 404 fallback, unknown-owner friendly error.
- Full `go test ./...`, `go build`, `go vet`, and `gofmt` all clean.

Closes #131


connorshea:
Haven't had a chance to review or test this yet and I'm about to go to sleep, so feel free to ignore this PR until tomorrow

andrew (changes requested):
Nice work, a few things before this lands:

1. `type=all` on the user fallback (`org_import.go:105`) pulls in repos the user is a collaborator on, not just ones they own. For `/users/{user}/repos` GitHub defaults to `type=owner`, which is what we want here. Either drop the `type` param entirely or only set it on the `/orgs` path.

2. Mirrors should be skipped by default alongside forks and archived. Decode `mirror_url` into `OrgRepo` and skip when it's non-empty. I don't think it needs its own toggle, mirrors are never something we'd want to scan independently of the upstream.

3. The handler reads `r.FormValue("model")` and threads it through the confirm step, but the org dialog in `layout.html` has no model input, so it's always empty. Either add the selector to match the single/bulk add dialogs or drop the plumbing.

Minor: `orgImportUA` duplicates `userAgent` from `internal/worker/metadata.go`. Fine to leave as-is given the package split, just flagging.

connorshea:
> Nice work, a few things before this lands:
> 
>     1. `type=all` on the user fallback (`org_import.go:105`) pulls in repos the user is a collaborator on, not just ones they own. For `/users/{user}/repos` GitHub defaults to `type=owner`, which is what we want here. Either drop the `type` param entirely or only set it on the `/orgs` path.
> 
>     2. Mirrors should be skipped by default alongside forks and archived. Decode `mirror_url` into `OrgRepo` and skip when it's non-empty. I don't think it needs its own toggle, mirrors are never something we'd want to scan independently of the upstream.
> 
>     3. The handler reads `r.FormValue("model")` and threads it through the confirm step, but the org dialog in `layout.html` has no model input, so it's always empty. Either add the selector to match the single/bulk add dialogs or drop the plumbing.
> 
> 
> Minor: `orgImportUA` duplicates `userAgent` from `internal/worker/metadata.go`. Fine to leave as-is given the package split, just flagging.

Fixed all 3 of these concerns 👍

andrew (approved):

PR #417, opened 2026-06-17
skill: Capture the reproducer source, not just its output

edorian:
The verify and security-deep-dive reports recorded the output of a reproduction but not the script that produced it. A confirmed or fixed finding showed a result with no way to tell what was actually run or to re-run it, making the evidence trail useless to a later verify pass or an analyst.

---

When scanning i got annoyed that the output of the `Step 3: Validation` always includes a references to a file but never the actual file.

E.g.:

```
[...]

To reproduce run:

$ php /tmp/poc.php
This clearly show the issues exists: ...
```

This PR attempts to resolve this by including the reproducer script in the initial report instead of having to recreate it multiple times. 

I have no strong feelings on the matter (and I don't know if this is the correct solution). But it's an attempt , so let me know if you like it :)


andrew (changes requested):
Agreed this is a real gap, and a separate `reproducer` field is the right shape.

The `security-deep-dive` half works as-is since it's tightening the existing `validation` field. The `verify` half doesn't yet: `parseVerifyOutput` in `internal/worker/skill_parsers.go:598-602` only decodes `status`, `evidence`, and `notes`, so the new `reproducer` field will be populated by the model and then silently dropped before it reaches the FindingNote.

Add `Reproducer string` to that struct and write it into the note builder ahead of `evidence` (plus a test assertion in `skill_parsers_test.go`), and this is good to go.

Issue #315, opened 2026-06-02
Consider integrating more skills


miketheman:
Anthropic shared theirs here: https://github.com/anthropics/defending-code-reference-harness

andrew:
Thanks Mike, useful pointer.

Scrutineer already covers most of that harness's stages: `security-deep-dive` (find), `vuln-scan`, `verify`, `patch`, `finding-dedup`, `threat-model`, `repo-overview` (recon-ish). The pieces it has that we don't yet map onto existing backlog items:

- `novelty.py` → #112
- `grade.py` / `judge.py` → #111
- `recon.py` emitting focus areas → #108

I'll cite the reference harness as prior art on those three. Closing this one in favour of them; reopen if there's a specific skill from it I've missed.

PR #423, opened 2026-06-18
Add reference-harness inspired vuln-scan skill

abhinavgautam01:
Ref #315
## Summary

Adds a bundled `vuln-scan` skill adapted from Anthropic's defending-code reference harness. The skill performs a broad, read-only static source-code vulnerability scan, splits review by focus area and emits normal Scrutineer findings for later verification.

## Changes

- Add new `skills/vuln-scan` bundled skill with Scrutineer metadata
- Add a strict findings schema for `vuln-scan` reports
- Wire `vuln-scan` into the default `triage` code-scan path
- Document the new skill in README and skills docs
- Add bundled schema tests for valid and invalid `vuln-scan` output

## Testing

- `go test ./internal/worker -run TestBundledSchemas -count=1`
- `go test ./internal/skills -run TestLoadDirectory_bundledSkillsAreValid -count=1`
- `git diff --check`
- `go test ./...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ When a repo is added, the `triage` skill is enqueued. Its SKILL.md lists the ski
 | `subprojects` | Enumerates monorepo packages/workspaces so deep-dives can be scoped to a sub-path |
 | `threat-model` | Derives the project's security contract (components, entry-point trust table, claimed and disclaimed properties) for the deep-dive to load |
 | `semgrep` | Static analysis mapped into findings shape |
+| `vuln-scan` | High-recall model-backed static candidate scan adapted from Anthropic's defending-code reference harness |
 | `zizmor` | GitHub Actions workflow audit mapped into findings shape |
 | `ingest` | Normalizes external reports in arbitrary formats into findings when `/v1/import` cannot recognise the payload |
 | `security-deep-dive` | The model-backed audit producing structured findings |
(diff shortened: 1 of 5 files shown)


andrew (changes requested):
The skill itself is well done — the focus-area breakdown, the "do not report" list, and the schema are all solid and match the reference harness shape nicely.

I'd like to land it without the `skills/triage/SKILL.md` line for now though. With `scrutineer.model: max`, `max_turns: 90`, and up to 10 subagents, wiring it into the default `has_code` path means every code repo runs two full model-backed audits from the moment this merges. I'd rather have it available as an opt-in run from the UI first, compare its output and cost against `security-deep-dive` on a handful of real targets, and then enable it in triage as a follow-up once we know the overlap.

So: drop the triage change, keep everything else (skill, schema, docs, schema tests). `README.md` and `docs/skills.md` will also need a rebase onto main after today's merges.

abhinavgautam01:
thanks,

I dropped the `skills/triage/SKILL.md` change so `vuln-scan` is available as an opt-in skill only for now. The skill, schema, docs and schema tests are still included.

I also rebased onto latest `main`; the final diff no longer touches triage.

Local checks pass:
- `go test ./internal/worker -run TestBundledSchemas -count=1`
- `go test ./internal/skills -run TestLoadDirectory_bundledSkillsAreValid -count=1`
- `go test ./...`

andrew (approved):
Triage wiring dropped, rebased clean. We can revisit enabling it by default once we've compared output against security-deep-dive on a few real targets.

PR #429, opened 2026-06-18
ruby profile: ship PROFILE.md orientation guide

andrew:
The `ruby` profile builds a runner image but injects no per-container guidance at scan time, unlike `php`/`php-ext` which each ship a `PROFILE.md`. The scanning agent gets no Ruby-specific orientation as a result.

Add `docker/profiles/ruby/PROFILE.md`, modeled on the php guide:

- runtime summary (Ruby 3.4, `gem`, `bundle`, native-extension toolchain)
- install-from-lockfile step (`bundle install`) with an offline fallback note
- reproducer recipe (`ruby -e`, `bundle exec ruby /tmp/poc.rb`)
- out-of-scope note for installed gems

Extend the profile-guide test (renamed `TestProfileGuidesShipForPHP` -> `TestProfileGuidesShip`) to assert `ruby` ships a guide alongside `php`/`php-ext`.

### DIFF
--- a/docker/profiles/ruby/PROFILE.md
+++ b/docker/profiles/ruby/PROFILE.md
@@ -1,0 +1,43 @@
+# Ruby scanning container
+
+The repository under `./src` is a Ruby project.
+
+## Runtime
+
+- **Ruby 3.4** — `ruby`
+- **`gem`** on PATH for installing individual gems.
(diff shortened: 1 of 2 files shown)


connorshea (approved):
LGTM

PR #430, opened 2026-06-18
node profile: add Node.js ecosystem runner and guide

andrew:
Adds a `node` profile so scans of npm/pnpm/yarn projects can install their locked dependency set and reproduce findings in-place, matching `php`/`ruby`.

`docker/profiles/node/Dockerfile` extends the runner with:

- a sha256-verified Node 22 LTS tarball, fetched and checksummed the same way the runner installs `claude` (both arches listed), rather than from a moving apt repo
- `pnpm` and `yarn` (classic 1.x) pinned to exact versions; the npm registry keeps exact versions forever so the pin never 404s
- `build-essential` so `node-gyp` can compile native addons (`python3` is already in the runner)
- `corepack` left disabled so the pinned globals are what `pnpm`/`yarn` resolve to, instead of corepack reaching the network mid-scan

`docker/profiles/node/PROFILE.md` gives the scanning agent the per-manager install command (`npm ci` / `pnpm install --frozen-lockfile` / `yarn install --frozen-lockfile`), the reproducer recipe (`node -e`, `/tmp/poc.js`, ESM notes), and an out-of-scope note for `node_modules`.

Registers `node` (ecosystem `npm`) in `builtinProfiles` and covers detection, naming, and the shipped guide in the worker tests. Built locally on top of the runner image and smoke-tested: `node`, `npm`, `pnpm`, `yarn` and the node-gyp toolchain (`python3`/`make`/`cc`/`g++`) all resolve as the `runner` user.

### DIFF
--- a/docker/profiles/node/Dockerfile
+++ b/docker/profiles/node/Dockerfile
@@ -1,0 +1,80 @@
+# Node.js profile. Extends the scrutineer runner with a pinned Node.js
+# interpreter plus npm, pnpm, and yarn so scans of npm/pnpm/yarn projects
+# can install their locked dependency set and reproduce in-place.
+#
+# Node ships official, sha256-verifiable prebuilt binaries (glibc), so the
+# interpreter is fetched and checksummed the same way the runner installs
+# claude -- not from a moving apt repo. Bump NODE_VERSION and the matching
+# checksums together; both arches are listed because the worker may build
+# this on amd64 or arm64 hosts.
+#
+# At runtime the worker builds this on demand (see internal/worker/profile.go:
+# EnsureImage), tagging the result scrutineer-profile-node:<dockerfile-sha>
+# and passing the configured runner image via --build-arg RUNNER_IMAGE.
+#
+# To build it manually the way the project does, from the repo root:
+#
+#   # 1. Build the base runner image locally (Debian/glibc, ships brief etc.):
+#   docker build -t scrutineer-runner:local -f Dockerfile.runner .
+#
+#   # 2. Build this Node profile on top of that local runner:
+#   docker build -t scrutineer-profile-node \
+#       -f docker/profiles/node/Dockerfile \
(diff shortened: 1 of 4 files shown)


docker/profiles/node/PROFILE.md
```diff
@@ -0,0 +1,53 @@
...
+yarn install --frozen-lockfile  # yarn.lock
+```
+
+Pick one based on which lockfile exists; do not run more than one. If only `package.json` exists (no lock), call out the
+missing lock in the report and fall back to `npm install`. If install fails with `Could not resolve host` or a similar
+network error the scan is offline — proceed without installed packages and note which checks you had to skip.
+
+`corepack` is disabled, so the `pnpm`/`yarn` above are the image's pinned globals. If a project pins a different exact
+manager version via the `packageManager` field, just use the global one and note the version difference rather than
+trying to fetch the pinned version.
```
connorshea, line 31: I think corepack may be necessary here, as differences in pnpm/yarn versions _will_ result in differences in what is installed and whether it's successful for some projects.

docker/profiles/node/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+ARG NODE_SHA256_ARM64=4018815ac1bed4f18208901bbde524fee881253b591ee7bc952660e69bd057af
+
+# pnpm and yarn pinned to exact versions. Unlike sury's apt repo, the npm
+# registry keeps every published version forever, so an exact pin never
+# 404s. yarn stays on the 1.x (classic) line, which reads the yarn.lock v1
+# format most projects still ship. corepack is intentionally left disabled
+# so these globals -- present offline -- are what `pnpm`/`yarn` resolve to,
+# rather than corepack reaching out to the network mid-scan.
+ARG PNPM_VERSION=11.8.0
+ARG YARN_VERSION=1.22.22
```
connorshea, line 48: should there be SHA verification for these as well?

PR #432, opened 2026-06-18
python-ext profile: add ASan/UBSan CPython for native extensions

andrew:
Stacked on #431 (python profile) — review/merge that first; this branch's base will need retargeting to main after #431 merges.

Adds a `python-ext` profile, the native-code twin of `python` and the Python analogue of `php-ext`. It ships CPython built from source with `--with-pydebug` + AddressSanitizer + UndefinedBehaviorSanitizer, so C/C++ extension modules installed against it are instrumented and a reproducer that drives the extension surfaces memory-safety and UB bugs the static scan can only guess at.

Build notes:

- CPython is built with **gcc**. clang now defaults to `-Werror=implicit-function-declaration`, which breaks configure's `getaddrinfo` probe (`checking for getaddrinfo... no` then a hard error). `ac_cv_buggy_getaddrinfo=no` is kept as belt-and-suspenders for the ASan runtime probe.
- Extensions are built with the same gcc plus the exported sanitizer `CFLAGS`/`CXXFLAGS`/`LDFLAGS`, so they share CPython's gcc ASan runtime (a clang-built `.so` dlopened into gcc-ASan CPython aborts on a runtime mismatch). `-fno-sanitize-recover=undefined` makes UB in an extension abort rather than silently continue.
- `PROFILE.md` mirrors php-ext: why the container exists, the pre-exported sanitizer config, how to build the extension in place, how to work a sanitizer hit into a security finding, and reproducer rules.

Routing: auto-selected (ordered before `python`) for repos whose `setup.py` declares a C `Extension(`, and selectable by name via `?profile=python-ext`. The debug+sanitizer interpreter is far slower than the plain `python` profile, so it is scoped to repos that actually ship native code. One thing worth a sanity check in review: whether `setup.py` containing `Extension(` is the right auto-detect signal, or whether python-ext should be selection-only.

Covers marker detection, naming, and the shipped guide in the worker tests. Built locally on the runner image: CPython 3.13.14 debug+ASan compiles and runs, and a deliberately buggy C extension built in place (`build_ext --inplace`) trips ASan with a heap-buffer-overflow when called from Python.

### DIFF
--- a/docker/profiles/python-ext/Dockerfile
+++ b/docker/profiles/python-ext/Dockerfile
@@ -1,0 +1,103 @@
+# Python-extension profile.
+#
+# Ships CPython built from source with --with-pydebug + ASan + UBSan, for
+# scanning Python packages that include compiled C/C++ extension modules.
+# C extensions installed against this interpreter (CFLAGS/LDFLAGS wire the
+# sanitizers in) are instrumented, so a reproducer that drives the
+# extension surfaces memory-safety and UB bugs the static scan can only
+# guess at. This mirrors the php-ext profile.
+#
+# Auto-selected for repos whose setup.py declares a C Extension (see
+# internal/worker/profile.go); also selectable by name via ?profile=. The
+# debug+sanitizer interpreter is much slower than the plain python
+# profile, so it is scoped to repos that actually ship native code.
+#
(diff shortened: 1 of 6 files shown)


andrew:
Superseded by #459 (this was auto-closed when #431's merge deleted its base branch).

Issue #436, opened 2026-06-18
Pause scans when tokens are gone, not fail them


mechko:
This is about monitoring available credits on the used Claude account, and a "Pause" and "Resume Scan" function for scheduled projects. 

In a bulk scan of several projects (like using the "Scan all Subprojects" function), it would be nice if Scrutineer would not just fail them all if tokens ran out, but rather keep them in the queue and only "pause" the execution. Which could be then resumed if new tokens are available.

andrew:
PR over here: https://github.com/alpha-omega-security/scrutineer/pull/439 for pausing all when you run out of credit, but it doesn't automatically restart them yet

andrew:
Done in #439: hitting the account token limit now pauses the scan (and any others still queued for the same repo) instead of failing them, and #473 adds a per-repo Resume all button on the Scans tab.

PR #439, opened 2026-06-18
worker: pause scans on Claude token limit instead of failing them

andrew:
Addresses the core of #436: in a bulk run (e.g. Scan all Subprojects), running out of tokens should not fail every scan — they should be kept and resumable.

When a scan hits an account-wide Claude token/credit limit (already detected as `ClaudePlanLimitError` via the stderr/stdout phrase matching in `claude_limit.go`), it is now marked **paused** rather than **failed**, and every other still-queued scan is paused too. Otherwise the rest of the batch would each run straight into the same wall and fail one by one. Pausing keeps the batch intact so the operator hits the existing **Resume paused** action once the limit resets.

Most of the machinery already existed — the `ScanPaused` status, the pause/resume handlers (`scansResumePaused`/`resumeScan`), and the limit detection. This change wires the limit error to a pause:

- `finishErroredScan` sets `ScanPaused` (was leaving it failed)
- the worker loop bulk-pauses the rest of the queue via `pauseQueuedOnTokenLimit`; `wrap()` already drops stale queue jobs whose scan is no longer `queued`, so updating the rows is enough (same mechanism as a manual "Pause queued")
- paused scans keep the `Claude plan limit reached.` error prefix, so the scans page surfaces them: the stat and banner now key on paused-with-limit rather than failed-with-limit and point at Resume
- extracted `persistFinishedScan` from `wrap` to stay under the gocognit cap

Tests: a worker test confirms the triggering scan and a second queued scan both end paused with the limit prefix; the scans-stats and jobs-banner tests are updated for the paused semantics.

Deliberately **not** included, and worth a separate issue/PR: the proactive half of #436 — monitoring the account's credit balance and auto-resuming when tokens return. Anthropic doesn't expose a simple per-account credit-balance API for all plan types, so 'know when credits are back' realistically means a timed retry (plan-limit windows) or operator action; I'd scope that on its own rather than bolt it on here.

### DIFF
--- a/internal/web/list_performance_test.go
+++ b/internal/web/list_performance_test.go
@@ -47,11 +47,14 @@ func TestScanListStatsAggregatesCounts(t *testing.T) {
 	for _, status := range []db.ScanStatus{db.ScanQueued, db.ScanQueued, db.ScanPaused, db.ScanRunning} {
 		s.DB.Create(&db.Scan{RepositoryID: repo.ID, Kind: "skill", Status: status})
 	}
+	// A scan paused because the account hit the Claude token/credit limit
+	// (auto-paused by the worker), plus an unrelated failure that must not
+	// be counted.
 	s.DB.Create(&db.Scan{
 		RepositoryID: repo.ID,
 		Kind:         "skill",
-		Status:       db.ScanFailed,
-		Error:        "Claude plan limit reached. Pause queued scans and retry after your limit resets.",
+		Status:       db.ScanPaused,
+		Error:        "Claude plan limit reached. Queued scan paused automatically; resume after the limit resets.",
 	})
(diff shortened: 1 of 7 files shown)


docker/profiles/beam/Dockerfile
connorshea, line 1: Looks like this accidentally included changes from the elixir branch

connorshea (changes requested):

andrew:
Dropped the stray beam-profile commits and rebased onto main (which also folded persistFinishedScan into the new finalizeScan from #446). Down to +87/-26 now.

connorshea (approved):
Works well for me! :)

Image

PR #442, opened 2026-06-19
fix: gate patch diffs against the resume-aware workspace

connorshea:
Generated with Claude Code, reviewed and tested by me. This should avoid some potential problems when resuming a scan.

`parsePatchOutput` resolved the diff-staging dir with `workRoot(scan.ID)`, the only staging call site not going through `scanWorkRoot`. Every other site (`skill.go`, `skill_parsers.go`, `exposure.go`) resolves through the resume lineage, which maps a resumed scan to its lineage-root workspace where `src/` is actually staged.

On a resumed patch run (`ResumedFromScanID` set) `scan.ID != workspace id`, so `gatePatch` stat'd `scan-<retryID>/src` which never existed, failed every `os.Stat`, and silently rejected valid diffs with "diff targets missing file". So `suggested_fix` was never written.

### DIFF
--- a/internal/worker/patch_gate.go
+++ b/internal/worker/patch_gate.go
@@ -52,7 +52,7 @@ func (w *Worker) parsePatchOutput(scan *db.Scan, report string, emit func(Event)
 		return fmt.Errorf("load finding %d: %w", *scan.FindingID, err)
 	}
 
-	srcDir := filepath.Join(w.workRoot(scan.ID), "src")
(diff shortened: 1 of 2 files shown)


connorshea:
And of course I have since hit this exact problem while attempting to generate a patch in scrutineer that hit max-turns and needed to be retried lol (not running the code changes from this PR, this PR should indeed fix the problem)

andrew (approved):
Confirmed this was the only staging call site bypassing `scanWorkRoot`. Good precondition assert in the test.

PR #443, opened 2026-06-19
fix(security-deep-dive): stop subagents clobbering report.json

connorshea:
This fix was generated by Claude Code, reviewed by me but not deeply tested. We may want to test this further before merging it, but I believe it should resolve the issue I've been seeing.

On a large repo (e.g. a production Rails app with >10k files), the skill often fans out to subagents to parallelize the audit. Claude Code subagents share the parent's working directory and do not inherit SKILL.md, so each infers the deliverable from the cwd and writes to the same `./report.json` file, overwriting the last and silently dropping the others' findings. A clobbered-but-partial report is still schema-valid, so the harness doesn't always realize that this has occurred and - in my experience - can lose report context as a result. This wastes tokens and is just generally Not A Good Time.

Add an orchestrator-framed fan-out section: delegate slices to subagents, tell each not to touch `./report.json`, give each a distinct scratch file (`inventory-<area>.json` / `dispositions-<area>.json`) so single-writer is mechanical, and have the orchestrator be the sole writer. Strengthen Output so consolidation is a union, not a copy of the last slice, dropping no sink - mirroring vuln-scan's fan-out then consolidate pattern.


andrew (approved):
The failure mode is real and the single-writer + per-slice scratch file pattern is the right fix. Additive instruction only, so small-repo runs are unaffected.

andrew:
vuln-scan has the same exposure (subagents, shared cwd, single `report.json`, no clobber guard) — opened #450 to port this pattern there.

PR #446, opened 2026-06-19
worker: auto-enqueue finding-dedup after deep-dive

connorshea:

## What

Auto-runs the `finding-dedup` skill after a `security-deep-dive` scan completes, and hardens the concurrent revalidate/dedup interaction.

### Commit 1 — auto-enqueue finding-dedup after deep-dive
Wires `autoEnqueueFindingDedupAfterDeepDive` onto `Worker.OnScanFinalized`. After a deep-dive scan finalizes (and its findings are committed), enqueue a repo-scoped dedup run **only when both** hold:

1. The deep-dive produced at least one *new* finding (re-observed findings keep their original `scan_id`, so `scan_id == this scan` counts exactly the new rows).
2. The repo already held another *open non-scanner* finding to compare against — matching the Findings-tab toggle exactly (`nonScannerScanFilter`): deep-dive/legacy/manual only; the cheap scanners (semgrep, zizmor) and tool imports (CodeQL, Snyk) do **not** count.

Both counts read committed DB state rather than threading a tally through the callback, so the decision is independent of parse-time races. A dedup run already queued/in-flight for the repo is a no-op. Enqueue errors are logged and swallowed so they never fail the upstream scan.

`OnScanFinalized` fires on `ScanDone` **or** a `FailOnThresholdError` (fail_on still commits findings).

### Commit 2 — skip revalidate on findings that have left the active funnel
A finding-scoped revalidate and the repo-scoped dedup can run concurrently (worker concurrency = 4). If dedup marks a finding `duplicate` first, a later revalidate would promote it `new->enriched`, cache a verdict, and chain a verify run on a finding nobody will look at — clobbering the duplicate status and wasting spend.

`parseRevalidateOutput` now bails when the loaded finding is closed. Adds `db.FindingLifecycle.Closed()` (the in-memory counterpart to the `status NOT IN ClosedFindingLifecycles` SQL filter) and refactors the dedup-side `dedupCandidateOpen` to its complement, so both skills share one definition of "open".

## Race notes
No data-corruption risk: SQLite (WAL + `busy_timeout`) serializes writers, and both skills write through `WriteFindingField` (targeted single-column `UPDATE`, not whole-row `Save`). Commit 2 closes the more damaging half of the lost-update window on the shared `status` column. The symmetric ordering (revalidate's `enriched` landing after dedup's `duplicate`) remains last-writer-wins; fully closing it would need a status-guarded CAS in `WriteFindingField`, left as a follow-up since it touches shared infrastructure.

## Tests
- `internal/web`: table test covering both gating conditions, scanner/import/closed exclusion, non-deep-dive skip, double-queue guard, absent-skill graceful, nil scan.
- `internal/worker`: `TestParseRevalidate_skipsClosedFinding`.
- Full `db` / `worker` / `web` suites, `go build`, and `go vet` all pass.

### DIFF
--- a/internal/web/finding_dedup_enqueue.go
+++ b/internal/web/finding_dedup_enqueue.go
@@ -11,8 +11,8 @@ import (
 // See skills/finding-dedup/SKILL.md.
 const findingDedupSkillName = "finding-dedup"
 
-// autoEnqueueFindingDedupAfterDeepDive is wired onto Worker.OnScanDone. The
-// worker calls it once after a scan completes and its findings are
+// autoEnqueueFindingDedupAfterDeepDive is wired onto Worker.OnScanFinalized.
+// The worker calls it once after a scan completes and its findings are
 // committed. We enqueue a repository-scoped finding-dedup run only when both
 // conditions the dedup pass needs to be worth its model spend hold:
 //
@@ -20,10 +20,11 @@ const findingDedupSkillName = "finding-dedup"
 //     finding. Re-observed findings keep the scan_id of the run that first
 //     created them, so counting findings with scan_id == this scan counts
 //     exactly the new rows. Nothing new means nothing fresh to dedup.
-//  2. The repository already held other non-scanner findings before this
-//     scan — open deep-dive/legacy/manual rows from a prior scan, an import
-//     of that kind, or a manual entry. Without something else to compare
-//     the new findings against, dedup has no pairs to consider.
+//  2. The repository now holds at least two open non-scanner findings in
+//     total (the new rows count toward this). Dedup needs a pair to compare,
+//     but the pair need not predate this scan: a first-ever deep-dive that
+//     emits several findings describing the same bug from different subagent
+//     angles is exactly what dedup exists to collapse.
 //
 // "Non-scanner" matches the Findings-tab toggle exactly (nonScannerScanFilter):
 // the cheap tool scanners (semgrep, zizmor) and tool imports (CodeQL, Snyk)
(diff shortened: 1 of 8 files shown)


internal/web/finding_dedup_enqueue.go
```diff
@@ -0,0 +1,95 @@
...
+	if err := s.DB.Model(&db.Finding{}).
+		Where("repository_id = ? AND scan_id <> ?", scan.RepositoryID, scan.ID).
+		Where(nonScannerScanFilter, deepDiveSkillName).
+		Where("status NOT IN (" + db.ClosedFindingLifecycleSQLValues() + ")").
+		Count(&otherNonScanner).Error; err != nil {
+		s.Log.Warn("auto-enqueue finding-dedup: count existing findings",
+			"scan", scan.ID, "repo", scan.RepositoryID, "err", err)
+		return
+	}
+	if otherNonScanner == 0 {
```
andrew: Requiring a *prior* finding (`scan_id <> ?`) means a first-ever deep-dive never triggers dedup, even when it emits several findings that describe the same bug from different subagent angles — which #443 makes more likely now each subagent contributes its own rows. Please change the gate to fire when the repo has ≥2 open non-scanner findings in total with ≥1 of them new: drop the `scan_id <> ?` from this query, require the count to be ≥2, and add a table case for `newFinding`×2 with `hasPrior=false → wantQueued=true`.

internal/web/finding_dedup_enqueue.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"scrutineer/internal/db"
+)
+
+// findingDedupSkillName is the repository-scoped pass that compares open
+// findings and marks ones describing the same vulnerability as duplicates.
+// See skills/finding-dedup/SKILL.md.
+const findingDedupSkillName = "finding-dedup"
+
+// autoEnqueueFindingDedupAfterDeepDive is wired onto Worker.OnScanDone. The
```
andrew: `OnScanDone` → `OnScanFinalized`

andrew (changes requested):
The hook placement, race handling and test matrix are all good. One gap in the gating logic below.

internal/web/finding_dedup_enqueue.go
connorshea: Updated

andrew (approved):

PR #447, opened 2026-06-19
feat: Skip `dependents` job when `packages` comes back with no published packages

connorshea:

The `dependents` skill only has work when the repository actually publishes a package. Today `triage`'s `has_packages` gate is true for any repo with a package manager (e.g. a `Gemfile` counts), so it cannot tell "publishes" from "consumes": a Rails app gets `dependents` enqueued, runs a full agent, queries `ecosyste.ms`, finds nothing, and exits. This adds a content-aware gate in the worker prereq layer, where `dependents` already waits on `packages`, to skip that wasted agent run.

This should eliminate unnecessary token spend when scanning something like a Rails app.

This may not be the right place to put this guard, I can try other options if preferred, but the initial attempt by Claude required an additional database column on skills, which seemed quite excessive for this change.

### DIFF
--- a/internal/worker/preflight.go
+++ b/internal/worker/preflight.go
@@ -19,6 +19,14 @@
 // reached a terminal failed/cancelled state the dependent fails
 // immediately rather than burning the retry budget waiting on
 // something that will not recover on its own.
+//
+// One content-aware special case sits on top of the generic rule: the
+// dependents skill is meaningful only when the repository publishes a
+// package. Once its packages prereq is satisfied, if packages ran but
+// produced no Package rows the dependents scan is marked a no-op done
+// rather than dispatched, sparing an agent run that could only emit an
+// empty result. dependents is the sole skill that requires packages, so
+// this stays a name-keyed check here instead of generic machinery.
 
 package worker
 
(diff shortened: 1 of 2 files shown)


internal/worker/preflight.go
```diff
@@ -156,6 +177,72 @@ func (w *Worker) unsatisfiedPrereqs(repoID uint, names []string) (pending, dead
...
+	if done == 0 {
+		return false
+	}
+	var live int64
+	w.DB.Model(&db.Scan{}).
+		Where("repository_id = ? AND skill_name = ? AND status IN ?",
+			repoID, "packages", []db.ScanStatus{db.ScanQueued, db.ScanRunning, db.ScanPaused}).
+		Count(&live)
+	if live > 0 {
+		return false
```
andrew, line 231: Returning `false` here makes the caller dispatch dependents while the packages re-scan is mid delete-then-recreate, so it runs against an empty `Package` table — the wasted agent run this PR is meant to prevent, just on the rescan path. The comment already has it right: this case should defer. Please make `live > 0` re-enqueue with `prereqRetryDelay` (the machinery is a few lines up in `preflightSkill`) and flip `TestPreflightSkill_dependentsDispatchesWhenPackagesRescanInFlight` to assert `deferred == true` with the scan still queued. A bool return can't express skip/defer/proceed, so this probably wants either a tristate return or the live-check lifted into the caller.

andrew (changes requested):
Right place for this, and the test matrix is good. One behaviour to fix in the rescan path below.

internal/worker/preflight.go
connorshea: Updated 👍

andrew (approved):

PR #451, opened 2026-06-20
Increase patch skill max turns

### LINKED ISSUE #448: Increase max turns default for `patch` skill
Should be more like 50, at the default 30 I frequently hit max turns when generating a patch.

abhinavgautam01:
closes #448

## Summary

Increases the bundled `patch` skill's per-skill max turns from the global default to 50, so patch generation has more room before hitting the turn cap.

## Changes

- Add `scrutineer.max_turns: 50` to `skills/patch/SKILL.md`
- Add regression coverage that the bundled `patch` skill loads with `MaxTurns == 50`

## Testing

- `go test ./internal/skills -run TestLoadDirectory_bundledSkillsAreValid -count=1`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/skills/parse_test.go
+++ b/internal/skills/parse_test.go
@@ -581,6 +581,13 @@ func TestLoadDirectory_bundledSkillsAreValid(t *testing.T) {
 	if n == 0 {
 		t.Fatal("no skills loaded from ../../skills")
 	}
+	var patch db.Skill
+	if err := gdb.Where("name = ?", "patch").First(&patch).Error; err != nil {
+		t.Fatalf("patch skill not loaded: %v", err)
+	}
+	if patch.MaxTurns != 50 {
+		t.Errorf("patch max_turns = %d, want 50", patch.MaxTurns)
+	}
 }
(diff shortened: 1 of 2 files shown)


internal/skills/parse_test.go
```diff
@@ -581,6 +581,13 @@ func TestLoadDirectory_bundledSkillsAreValid(t *testing.T) {
 	if n == 0 {
 		t.Fatal("no skills loaded from ../../skills")
 	}
+	var patch db.Skill
+	if err := gdb.Where("name = ?", "patch").First(&patch).Error; err != nil {
+		t.Fatalf("patch skill not loaded: %v", err)
+	}
+	if patch.MaxTurns != 50 {
+		t.Errorf("patch max_turns = %d, want 50", patch.MaxTurns)
+	}
```
connorshea, line 590: having a test specifically for this is probably excessive, but it's fine

connorshea (approved):

andrew (approved):

PR #453, opened 2026-06-20
fix(web): delete SBOM packages explicitly on upload delete

connorshea:
## Problem

`sbomDelete` relied on the `SBOMUpload.Packages` `ON DELETE CASCADE` constraint to clean up child `SBOMPackage` rows. But sqlite's `foreign_keys` pragma is **per-connection**, and `db.Open` enables it with a single `Exec` that lands on only one connection in the pool. In production the pool is uncapped and runs in WAL mode, so most serving connections have `foreign_keys=OFF` — the cascade silently no-ops and every package row for the deleted upload is orphaned.

This is the same hazard `repoDelete` already works around by deleting its children by hand.

### Empirical reproduction

A throwaway probe against the real `db.Open` path showed:
- **3 of 8** pooled connections reported `foreign_keys=OFF` despite `Open` setting it `ON`.
- **20 of 20** cascade-delete attempts left an orphaned child row.

## Fix

Delete the upload's packages and the upload itself explicitly, inside one transaction.

Parsing the id with `strconv.Atoi` for the child `WHERE` clause also closes the GORM inline-condition SQL-injection sink on this route as a side effect: a path like `/sboms/1 OR 1=1/delete` now 404s instead of reaching raw SQL. (The other inline-condition sinks elsewhere in the app are out of scope here.)

## Test

`TestSBOMDelete` now pins `SetMaxOpenConns(1)` + `PRAGMA foreign_keys=OFF` to reproduce the production hazard deterministically, then asserts zero orphaned packages remain. It **fails on the old handler** and **passes on the new one**.

Full `go test ./...` passes; `golangci-lint run` reports 0 issues.

### DIFF
--- a/internal/web/sboms.go
+++ b/internal/web/sboms.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/git-pkgs/sbom"
+	"gorm.io/gorm"
 
 	"scrutineer/internal/db"
 )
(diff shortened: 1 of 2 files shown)


connorshea:
This feels a bit goofy and it seems like we should maybe update the way the database connection works so foreign keys always work, but I don't know if that's possible / if SQLite really lets us do that, so.

andrew (approved):

Issue #305, opened 2026-06-01
Copy skill scripts up to repo root


connorshea:
Is there a reason not to also copy the relevant python scripts for a given skill in the same way? I noticed the sbom skill gets caught up trying to find git-pkgs and it seems like it may be fixable in the same way as this PR.

_Originally posted by @connorshea in https://github.com/alpha-omega-security/scrutineer/issues/229#issuecomment-4589156911_

PR #470, opened 2026-06-21
fix(worker): mirror context.json into skill dir so ./context.json resolves

connorshea:
## Summary

- `stageContext` writes `context.json` to `workRoot` (`/work/context.json`), but skill agents sometimes try to read it relative to the skill directory (`.claude/skills/{name}/context.json`) — the same failure mode that hit `schema.json` (#221) and `scripts/` (#305).
- Fix: `stageSkill` now reads `context.json` from `workRoot` and copies it into the skill directory (`dst`) alongside `SKILL.md` and `schema.json`.
- Added `TestStageSkill_mirrorsContextJSONToSkillDir` to cover the new behaviour.

## Test plan

- [ ] `go test ./internal/worker/ -run TestStageSkill` — all pass including new test
- [ ] `go test ./internal/worker/` — full suite passes
- [ ] `golangci-lint run ./internal/worker/...` — 0 issues

### DIFF
--- a/internal/worker/exposure.go
+++ b/internal/worker/exposure.go
@@ -163,11 +163,11 @@ func (w *Worker) doExposure(ctx context.Context, scan *db.Scan, emit func(Event)
 	}
 
 	skillDir := filepath.Join(workRoot, ".claude", "skills", skill.Name)
-	if err := stageSkill(&skill, workRoot, skillDir); err != nil {
-		return "", fmt.Errorf("stage skill: %w", err)
-	}
 	if err := stageContext(workRoot, w.APIBase, w.ForkOrg, w.metadataDir(), scan, &scan.Repository); err != nil {
 		return "", fmt.Errorf("stage context: %w", err)
+	}
+	if err := stageSkill(&skill, workRoot, skillDir); err != nil {
+		return "", fmt.Errorf("stage skill: %w", err)
 	}
 
 	depRepo := db.Repository{URL: dep.RepositoryURL, Name: dep.Name}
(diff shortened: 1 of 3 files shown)


andrew (changes requested):
This works for the main skill path but misses the exposure runner.

`stageSkill` reads `workRoot/context.json`, so it only mirrors anything if `stageContext` has already run. In `internal/worker/skill.go` the order is `stageContext` then `stageSkill`, so that's fine. But `internal/worker/exposure.go:166-169` calls them the other way round:

```go
if err := stageSkill(&skill, workRoot, skillDir); err != nil {
    ...
}
if err := stageContext(workRoot, w.APIBase, w.ForkOrg, w.metadataDir(), scan, &scan.Repository); err != nil {
```

So when the exposure skill runs, `context.json` doesn't exist yet at the point `stageSkill` tries to read it, the `os.ReadFile` fails silently, and nothing is mirrored.

Swapping the two calls in `exposure.go` to match `skill.go` would fix it and is probably the right answer regardless of this PR.

andrew:
Pushed 9653cdd with two small follow-ups:

- The `os.ReadFile` on `workRoot/context.json` now propagates any error other than `ErrNotExist`, matching the `mirrorScripts` pattern from #305. The previous `if err == nil` swallowed permission and I/O errors, which would silently leave the skill dir unmirrored.
- The `stageSkill` doc comment now notes it must run after `stageContext`, since that ordering is load-bearing as of this PR and the failure when they're swapped is silent.

Thanks for the `exposure.go` swap in 727377f, that was the right fix.

andrew (approved):
exposure.go ordering fixed and the read error path now matches mirrorScripts. Thanks @connorshea.

PR #472, opened 2026-06-21
feat(api): add validate-report endpoint for in-container schema checks

### LINKED ISSUE #471: Add `POST /api/scans/{id}/validate-report` endpoint
## Problem

When a skill runs inside a Docker container, Claude oftentimes attempts to self-validate its `report.json` against `./schema.json` before finishing. It does this by attempting to install a JSON Schema library (`pip install jsonschema`, `gem install json_schemer`, etc.), which wastes turns and is unreliable — in particular, `json_schemer` (Ruby) chokes on `contentMediaType: "text/markdown"` annotations in the schema, causing false failures.

The current setup is especially problematic with the many newly-added profiles, because Claude wastes multiple turns right now in a Ruby project attempting to `pip install` packages that it can never install because it does not have pip. Then it tried to create a schema validator in ruby, failed, and basically gave up on that. So we can save 3-4 turns in cases like that by just having Claude defer to Scrutineer on doing the validation itself.

The harness already validates server-side in `internal/worker/schema_validate.go` using `santhosh-tekuri/jsonschema/v6` and sends a repair prompt if validation fails (`repairSchemaReport` in `internal/worker/skill.go:218`). Claude's in-container self-validation is therefore redundant and broken.

## Fix

Add a lightweight `POST /api/scans/{id}/validate-report` endpoint that lets Claude validate its report via a curl call rather than installing packages. The endpoint reuses the existing `validateReportSchema` function from `internal/worker/schema_validate.go`, so the in-container check uses the exact same validator as the harness.

The container already has curl and the API token available in `./context.json` at `scrutineer.api_base` and `scrutineer.token`.

### 1. `internal/web/api.go`

Register the route in `apiHandler()` alongside the existing scan routes:

```go
mux.HandleFunc("POST /scans/{id}/validate-report", s.apiValidateReport)
```

Add the handler. The handler reads the request body as the candidate report JSON, looks
up the running scan to get the skill's `SchemaJSON`, and calls `validateReportSchema`.
The endpoint is protected by the existing `apiAuth` middleware, which restricts access
to the scan that owns the token — no additional auth logic needed.

```go
func (s *Server) apiValidateReport(w http.ResponseWriter, r *http.Request) {
    id, _ := strconv.Atoi(r.PathValue("id"))
    scan := r.Context().Value(ctxKeyScan).(*db.Scan)
    if scan.ID != uint(id) {
        writeAPIError(w, http.StatusForbidden, "scan may only validate its own report")
        return
    }
    if scan.SkillID == nil {
        writeAPIError(w, http.StatusBadRequest, "scan has no skill")
        return
    }
    var skill db.Skill
    if err := s.DB.First(&skill, *scan.SkillID).Error; err != nil {
        writeAPIError(w, http.StatusNotFound, "skill not found")
        return
    }
    if skill.SchemaJSON == "" {
        writeJSON(w, http.StatusOK, map[string]any{"valid": true, "note": "skill has no schema"})
        return
    }
    body, err := io.ReadAll(io.LimitReader(r.Body, 4<<20))
    if err != nil {
        writeAPIError(w, http.StatusBadRequest, "could not read body")
        return
    }
    detail := worker.ValidateReportSchema(skill.SchemaJSON, string(body))
    if detail == "" {
        writeJSON(w, http.StatusOK, map[string]any{"valid": true})
        return
    }
    writeJSON(w, http.StatusOK, map[string]any{"valid": false, "errors": detail})
}
```

Note: `validateReportSchema` is currently unexported. Either export it as
`ValidateReportSchema` in `internal/worker/schema_validate.go`, or move it to a shared
internal package. Exporting from the `worker` package and importing it in `web` is the
simplest path — the `web` package already imports `worker` indirectly via the server
struct, but check for import cycles first. If there is a cycle, extract the function to
a new `internal/schemavalidate` package and import it from both `worker` and `web`.

### 2. `internal/worker/schema_validate.go`

Export `validateReportSchema` → `ValidateReportSchema` (or move to a shared package as
noted above).

### 3. `skills/threat-model/SKILL.md` (and any other skill that self-validates)

In the self-check or output section, replace the package-install validation pattern with:

```
To validate report.json against the schema before finishing:
curl -s -X POST \
  -H "Authorization: Bearer {token}" \
  -H "Content-Type: application/json" \
  --data-binary @./report.json \
  "{api_base}/scans/{scan_id}/validate-report"
```

Where `token`, `api_base`, and `scan_id` come from `./context.json`. A `{"valid":true}`
response means the report matches the schema. A `{"valid":false,"errors":"..."}` response
lists the validation failures.

## What to test

- `POST /api/scans/{id}/validate-report` with a valid report returns `{"valid": true}`
- `POST /api/scans/{id}/validate-report` with an invalid report returns errors matching
  what `validateReportSchema` would produce for the same input
- A scan token cannot validate reports for a different scan ID (403)
- A skill with no `schema_json` returns `{"valid": true, "note": "skill has no schema"}`
- Body larger than 4 MB is rejected

There is an existing test file at `internal/web/api_test.go` — add cases there.
`internal/worker/schema_validate.go` already has `schema_bundled_test.go` for the
validate function itself; those tests do not need to change if the function is only
renamed/exported.

connorshea:
Add `POST /api/scans/{id}/validate-report` so a skill running in the runner container can validate its `report.json` against the skill's schema via curl, reusing the harness's own validator (ValidateReportSchema). 

This should improve the token usage and unnecessary additional turns required for Claude to install a schema validation package, write the validator, run the validator, and then get back a loud validation response. 

Changes:
- Export `worker.ValidateReportSchema` for reuse by the web package.
- Inject a concise schema-validation hint into the centralized skill activation prompt (`buildSkillPrompt`) for all JSON-output skills.
- Document the endpoint in `openapi.yaml`.

This was generated with Claude Code, fixes #471. 

I've tested this myself locally on two full repo scans and a few minor retries, I'm pretty confident in the change now.

Example logs after this change:

```
All 10 skills enqueued successfully (201). No 404s. Now let me write the report.
[write] /work/report.json
Now let me validate the report against the schema.
[bash] curl -s -X POST -H "Authorization: Bearer token" -H "Content-Type: application/json" -d @/work/report.json "http://host.docker.internal:8080/api/scans/3025/validate-report"
The schema doesn't allow `release_watch` as a field. Let me remove it.
[edit] /work/report.json
[bash] curl -s -X POST -H "Authorization: Bearer token" -H "Content-Type: application/json" -d @/work/report.json "http://host.docker.internal:8080/api/scans/3025/validate-report"
Report validated successfully. Here's what happened:
```

This should also reduce the likelihood of needing a re-run of the job due to schema errors.

### DIFF
--- a/internal/web/api.go
+++ b/internal/web/api.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
@@ -98,6 +99,7 @@ func (s *Server) apiHandler() http.Handler {
 	mux.HandleFunc("POST /repositories/{id}/skills/{name}/run", s.apiRunSkill)
 	mux.HandleFunc("POST /findings/{id}/skills/{name}/run", s.apiRunFindingSkill)
 	mux.HandleFunc("GET /scans/{id}", s.apiGetScan)
+	mux.HandleFunc("POST /scans/{id}/validate-report", s.apiValidateReport)
 	mux.HandleFunc("GET /findings/{id}", s.apiGetFinding)
 	mux.HandleFunc("PATCH /findings/{id}", s.apiPatchFinding)
 	mux.HandleFunc("GET /findings/{id}/notes", s.apiListFindingNotes)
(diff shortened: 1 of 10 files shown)


andrew (approved):
LGTM. Holding off merging per your note about further testing — go ahead when you're happy.

connorshea:
ah I tested this on 2 more full repo runs after opening it yesterday and forgot to update that part of the PR description, I'll update the description and merge.

PR #461, opened 2026-06-21
refactor: rename deepDiveScanIDs/autoEnqueueFindingDedup helpers

connorshea:
## Summary

Pure identifier renames in `internal/web`, with no behavior change:

- `deepDiveScanIDs` → `findingsScanIDs` (the GORM subquery helper + all call sites)
- `autoEnqueueFindingDedupAfterDeepDive` → `autoEnqueueFindingDedup` (the `OnScanFinalized` worker hook + tests)

The function bodies are untouched — `findingsScanIDs` still filters on the security-deep-dive skill name.

## Why

I wanted to land the various method renames as the first PR so it's easier to understand what's happening. I'll update the code in a subsequent PR to start including vuln-scan results as proper Findings instead of Scanner results.

## Test plan

- `go build ./internal/web/` passes
- `go vet ./internal/web/` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #478, opened 2026-06-22
feat(ecosystems): pre-fetch and cache ecosyste.ms data per repository

### LINKED ISSUE #322: Pre-fetch ecosyste.ms data server-side, cache per repository
## Problem

Five skills call ecosyste.ms during scan execution (`metadata`, `packages`, `advisories`, `dependents`, `maintainers`), and each fetch returns the same response on every scan of a given repo. The model still issues the WebFetch every time, reads the entire upstream payload into its context, and carries that payload across every subsequent turn of the conversation, which dominates the per-scan cost on these skills.

`maintainers` makes three of these fetches against commits.ecosyste.ms, issues.ecosyste.ms, and packages.ecosyste.ms; the responses then sit in the per-turn cached prefix across the remaining ~22 turns. 2693 completed runs in the dataset average 3.9 minutes and $0.65 each, working out to roughly 175 hours and $1750 of model time spent on JSON that didn't need a model to read it. The four single-endpoint skills run 2400-2700 times each at 1-3 minute averages, so the aggregate is meaningful even though the per-scan share is smaller.

## Proposal

Move the mechanical fetches into Go, cache the raw payloads on the Repository row, and expose them through the scrutineer API so the skills can read them without the WebFetch turn. Each skill keeps its existing body for the classification, normalisation, and SECURITY.md-shaped interpretation that the LLM is useful for; only the "fetch a known URL" step gets replaced with a "try cache first, fall back to WebFetch if 404" preamble.

## Storage

Text + datetime column pairs on `repositories`, mirroring the existing `metadata` blob pattern. Six sources:

- `ecosystems_repo_data` / `_fetched_at` (repos.ecosyste.ms lookup)
- `ecosystems_packages_data` / `_fetched_at` (packages.ecosyste.ms lookup)
- `ecosystems_advisories_data` / `_fetched_at` (advisories.ecosyste.ms)
- `ecosystems_commits_data` / `_fetched_at` (commits.ecosyste.ms)
- `ecosystems_issues_data` / `_fetched_at` (issues.ecosyste.ms)
- `ecosystems_dependents_data` / `_fetched_at` (chained off the packages response)

10 MB cap per column, matching the existing `maxResponseBody` constant.

## Trigger

Eager on repo-add via a goroutine that runs in parallel with the triage enqueue. A TTL-driven refresh runs in the worker at scan start so rescans see fresh-enough data without re-fetching what's still current.

## Refresh policy

TTL per source, checked at scan start. The same coordinator used for the eager prefetch is called with a stale-only filter, so a scan with everything fresh is a no-op.

| Source | TTL | Why |
|---|---|---|
| commits | 7 days | Lead-maintainer turnover; `report-upstream` correctness depends on this being current. |
| issues | 7 days | Secondary signal for lead identification. |
| advisories | 7 days | CVEs publish constantly; missing prior art during triage is the failure mode. |
| packages | 30 days | Registry ownership moves slowly; latest version drifts but isn't security-critical. |
| dependents | 30 days | Top-N dependents list moves slowly. |
| repo | 30 days | Description, language, stars. Mostly cosmetic. |

For disclosure-critical scans (`verify`, `disclose`, `report-upstream`) where 6-day-old data may still be too stale, a `?fresh=1` query param on each API endpoint forces an inline refresh. The disclosure-flow skills opt in; routine scans don't.

## API

The cache backs a small set of purpose-built endpoints rather than a single per-source raw-retrieval URL. Each endpoint reads the cached payload (refreshing if past TTL, or returning a 404 if there's nothing cached and `fresh=1` wasn't requested), filters and reshapes the data into the few-KB structure the calling skill actually needs, and returns it. All endpoints are gated by the existing per-scan bearer.

For `maintainers`, three endpoints cover what the current skill body looks for in the three upstream payloads:

- `GET /api/v1/repositories/{id}/maintainers/commits` returns an array of `{login, name, email, commits_total, commits_past_year, last_commit_at}`, ordered by past-year activity, with an optional `limit`.
- `GET /api/v1/repositories/{id}/maintainers/issues` returns `{login, name, prs_merged_past_year, issues_closed_past_year, reviews_past_year, last_active_at}` in the same shape.
- `GET /api/v1/repositories/{id}/maintainers/registry-owners` returns the distinct owner identities across the repo's published packages.

For `metadata`, `packages`, `advisories`, and `dependents`, the digested endpoints can map closely to the existing parser output shape, since each skill is already producing structured rows from the raw upstream response (and in several cases similar endpoints already exist for the UI).

A diagnostic endpoint `GET /api/v1/repositories/{id}/ecosystems/{source}/raw` returns the cached payload verbatim for operator use and to give a skill an escape hatch when it hits an edge case the digested endpoint doesn't cover; under normal operation the skill consumes only the digested form.

A 404 on any digested endpoint signals fall-back to WebFetch in the skill body, so the skill still works if the prefetch failed, hasn't run, or was deleted.

## Sequencing

Three PRs. The first introduces the Go fetcher, the schema columns, the API endpoints, the eager-on-add hook and the TTL refresh; no skill changes, so the cache populates passively while existing behaviour is unaffected. The second rewrites `maintainers` to consume the cache, which is where most of the per-scan saving lands because of the three-fetch shape. The third covers `metadata`, `packages`, `advisories`, and `dependents`, one skill per PR, in whichever cost-savings order is preferred since they're independent of each other.

## Open questions

The digested endpoint shapes for `maintainers` are the design choice with the most room for getting wrong, since omitting a field the current skill body uses for classification drops accuracy. The three sketches above are derived from the existing skill's data-source notes at `skills/maintainers/SKILL.md` lines 35-37 but want a closer reading before they're finalised, and the equivalent exercise has to happen for each skill before its PR.

TTL values are gut-feel from "lead turnover within a week is unusual" and worth a sanity-check from anyone with stronger intuition on actual disclosure cadence; the 7-day commits/issues number is the one most worth scrutiny since it's where disclosure routing correctness lands.

`?fresh=1` opt-in versus tighter TTLs everywhere is still open. Opt-in puts the cost where the correctness need is, but each disclosure-flow skill body has to remember to set the flag and forgetting it silently reads stale data.

alexandre-daubois:
Fix #322

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -383,6 +383,9 @@ func run(log *slog.Logger) error {
 			broker.Publish(web.Event{Name: name, Data: data, ScanID: scanID, RepoID: repoID})
 		},
 	}
+	w.RefreshEcosystemsCache = func(ctx context.Context, repoID uint) error {
+		return worker.RefreshEcosystems(ctx, gdb, repoID, true, log)
+	}
 	w.Register(q)
 
 	srv, err := web.New(gdb, q, log, broker, w)
--- a/docs/database.md
+++ b/docs/database.md
@@ -27,6 +27,18 @@ The central entity. One row per git URL.
 | icon_url | text | Avatar/icon URL. |
 | metadata | text | Full ecosyste.ms JSON response. Queryable with `json_extract`. |
 | fetched_at | datetime | When the metadata skill last ran. |
+| ecosystems_repo_data | text | Cached raw `repos.ecosyste.ms` lookup payload, pre-fetched server-side. |
+| ecosystems_repo_fetched_at | datetime | When `ecosystems_repo_data` was last refreshed. TTL 30 days. |
+| ecosystems_packages_data | text | Cached raw `packages.ecosyste.ms` lookup payload. |
+| ecosystems_packages_fetched_at | datetime | When `ecosystems_packages_data` was last refreshed. TTL 30 days. |
+| ecosystems_advisories_data | text | Cached `advisories.ecosyste.ms` payload, paginated and concatenated. |
+| ecosystems_advisories_fetched_at | datetime | When `ecosystems_advisories_data` was last refreshed. TTL 7 days. |
+| ecosystems_commits_data | text | Cached raw `commits.ecosyste.ms` lookup payload. |
+| ecosystems_commits_fetched_at | datetime | When `ecosystems_commits_data` was last refreshed. TTL 7 days. |
+| ecosystems_issues_data | text | Cached raw `issues.ecosyste.ms` lookup payload. |
+| ecosystems_issues_fetched_at | datetime | When `ecosystems_issues_data` was last refreshed. TTL 7 days. |
+| ecosystems_dependents_data | text | Cached dependents, chained off the packages lookup (per-package top dependents, capped). |
+| ecosystems_dependents_fetched_at | datetime | When `ecosystems_dependents_data` was last refreshed. TTL 30 days. |
 | disclosure_channel | text | Preferred reporting vector (email, GHSA URL, registry owner handle, SECURITY.md URL). Written by `maintainers`/`cna-match`; analyst-editable. |
 | posture | text | Disclosure-readiness tier from the `posture` skill: `ready`, `partial`, `unprepared`. |
 | posture_summary | text | One-line explanation that goes with `posture`. |
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -42,6 +42,25 @@ type Repository struct {
 	Metadata      string `gorm:"type:text"`
 	FetchedAt     *time.Time
 
+	// Ecosystems* hold raw ecosyste.ms payloads pre-fetched server-side so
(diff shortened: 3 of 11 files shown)


andrew (approved):
Looks good for step 1 of #322. Two things to keep an eye on, neither blocking:

Six text blobs on `Repository` means any unscoped `Find(&repos)` now pulls up to six payloads per row. The existing `Metadata` column already has this shape so the list handlers may already `Select()` around it, but worth checking the repo list and org pages and adding column lists if not. Fine as a follow-up.

The scan-start refresh in `wrap()` is synchronous and the sources loop is sequential. If the eager prefetch failed and everything is stale, the scan waits on 6+ round-trips (more with the dependents fan-out) before it starts. Bounded by `httpTimeout` per request so it's latency not a hang, but a repo with 25 packages chasing dependents could sit there for a bit. Parallelising the sources or moving the refresh off the scan's critical path would help, but can wait until it bites.

Needs a rebase onto main — `server.go`, `worker.go`, `server_test.go` and `db.go` have all moved (#439, #461, #473, #475 landed). Happy to merge once that's done.

PR #479, opened 2026-06-22
worker: pull runner base image before profile build

andrew:
`EnsureImage` ran `docker build` without refreshing the `FROM` base, so a stale locally cached `scrutineer-runner:latest` was reused. In #477 that stale image was the pre-#271 Alpine / pre-#304 amd64-only build, which on an arm64 host produced `apt-get: not found` plus an `InvalidBaseImagePlatform` warning.

This adds a best-effort `docker pull` of the configured runner image before the profile build. A failed pull (offline, or `--runner-image` pointing at a local-only tag like `scrutineer-runner:local`) is logged and the build carries on with whatever is cached, so the documented local-dev flow in the profile Dockerfile headers keeps working. `docker build --pull` was the obvious alternative but it hard-fails when the base only exists locally.

The pull only fires on a profile-image cache miss, which is already the slow first-build path.

Also lifts the build-arg assembly into `profileBuildArgs` so it can be unit tested without shelling out.

Not addressed here: `imageTag()` hashes the runner image ref string rather than its resolved digest, so a moved `:latest` won't invalidate an already-built `scrutineer-profile-*` image. Will file separately.

Fixes #477

Issue #477, opened 2026-06-22
Docker build failing when building locally


jamietanna:
I.e. when running `triage` on an M4 Pro MacBook, I see:

```
profile: building scrutineer-profile-python:9dcb5e673bd6 (first build can take several minutes)
profile: python build failed, using default: docker build scrutineer-profile-python:9dcb5e673bd6: exit status 1
#0 building with "desktop-linux" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 4.32kB done
#1 DONE 0.0s

#2 [internal] load metadata for ghcr.io/alpha-omega-security/scrutineer-runner:latest
#2 DONE 0.1s

#3 [internal] load .dockerignore
#3 transferring context: 2B done
#3 DONE 0.0s

#4 [1/3] FROM ghcr.io/alpha-omega-security/scrutineer-runner:latest@sha256:c57819ab132f98975c24e1b2b741ab31beb8e7f13cdaa12d09c6dbeec88560a5
#4 resolve ghcr.io/alpha-omega-security/scrutineer-runner:latest@sha256:c57819ab132f98975c24e1b2b741ab31beb8e7f13cdaa12d09c6dbeec88560a5 0.0s done
#4 DONE 0.2s

[...]

PR #481, opened 2026-06-22
fix(worker): fold runner base digest into profile image tag

### LINKED ISSUE #480: Profile images never rebuild when runner :latest moves
`imageTag()` in `internal/worker/profile.go` derives the profile image tag from a hash of the Dockerfile bytes plus the runner image ref string:

```go
func imageTag(profileName string, dockerfile []byte, runnerImage string) string {
	h := sha256.New()
	h.Write(dockerfile)
	h.Write([]byte{0})
	h.Write([]byte(runnerImage))
	...
}
```

When `runnerImage` is a moving tag like the default `ghcr.io/alpha-omega-security/scrutineer-runner:latest`, the hash never changes even though the underlying image does. So once `scrutineer-profile-python:9dcb5e673bd6` is in the local cache, `imageExistsLocally` short-circuits in `EnsureImage` and the profile is never rebuilt against a newer base. A user can sit on a months-old runner indefinitely.

#479 added a best-effort `docker pull` before the build, which fixes the case in #477 where the first build used a stale base. It doesn't help once the profile image already exists, because the pull is behind the `imageExistsLocally` check.

Options:

- Fold the resolved base digest into the hash. Resolve via `docker manifest inspect` (registry round-trip) or `docker image inspect --format {{.Id}}` after the pull (local, but only reflects what's pulled). A new digest produces a new tag, so the old profile image stays cached and the new one builds alongside it.
- Or drop the local short-circuit and always `docker build`; BuildKit's layer cache makes a no-op rebuild cheap, and combined with the `docker pull` from #479 the FROM would track upstream. Loses the per-tag mutex's "skip entirely" fast path though.
- Or pin `DefaultRunnerImage` to a `sha-<commit>` tag from the runner-image workflow rather than `:latest`, and bump it explicitly. The ref string then changes whenever the base does and the existing hash is sufficient. More release-process overhead.

The first option is probably the right one since it keeps `:latest` working for people who don't pin.

alexandre-daubois:
Fix #480

### DIFF
--- a/internal/worker/profile.go
+++ b/internal/worker/profile.go
@@ -310,18 +310,48 @@ func lockForTag(tag string) *sync.Mutex {
 }
 
 // imageTag returns the content-addressed tag for a profile's Dockerfile.
-// The runner image ref is folded into the hash so a `--runner-image`
-// bump invalidates profile images whose FROM resolved to the old base.
-// Editing the Dockerfile produces a new tag, so the local cache is
-// invalidated transparently. Old tags stay cached until the operator
-// prunes them.
-func imageTag(profileName string, dockerfile []byte, runnerImage string) string {
+// The runner image ref and its resolved registry digest are both folded
+// into the hash: editing the Dockerfile, pointing --runner-image at a
+// different ref, or a moved tag (the default :latest resolving to a new
+// digest) each yield a new tag, so the local cache is invalidated
+// transparently and the new image builds alongside the old. baseDigest is
+// empty when the digest can't be resolved (offline, or a local-only ref);
+// the tag then keys on the ref string alone, the behaviour before the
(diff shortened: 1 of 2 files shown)


andrew (approved):
Digest-keyed tag is the right fix for #480 — the ref-string hash never moved when `:latest` did, so cached profile images lived forever.

I pushed one commit on top to also cover #477 (closed #479 in favour of this): when `resolveBaseDigest` finds the runner in a registry, the build now passes `--pull` so BuildKit fetches the base matching that digest rather than reusing a stale locally cached `:latest`. Skipped when the digest didn't resolve (offline / local-only ref) so `scrutineer-runner:local` still builds from the local cache.

PR #483, opened 2026-06-22
disclose: post a summary note on the finding

### LINKED ISSUE #482: disclosure: "The draft is prepared. Review it in the notes..." but there's nothing in the notes
When generating a draft disclosure, it appears to not get added to notes - I only see `revalidate` and `verify`

andrew:
The disclose skill PATCHes the draft onto `Finding.DisclosureDraft` via the API, but unlike `verify`/`revalidate`/`patch`/`mitigate`/`release-watch` it had no parser writing a `FindingNote`, so the Notes panel showed nothing for the run and the model's "review it in the notes" closing message pointed at an empty list.

This adds a `disclose` `output_kind` whose parser posts a note summarising the run: the GHSA `summary`, which fields were patched and which were preserved, the reference counts, and the report's `notes` prose. An error-only report records `disclose: refused` with the reason.

Fixes #482.

### DIFF
--- a/internal/worker/skill_parsers.go
+++ b/internal/worker/skill_parsers.go
@@ -804,6 +804,63 @@ func (w *Worker) parseMitigationOutput(scan *db.Scan, report string, emit func(E
 	return nil
 }
 
+// parseDiscloseOutput posts a FindingNote summarising a disclose run so the
+// finding's Notes panel records that a draft was prepared, alongside the
+// verify/revalidate/patch entries (#482). The draft itself is on
+// Finding.DisclosureDraft (PATCHed by the skill via the API); this note is
+// the audit trail pointing at it: the GHSA summary, which fields were
+// patched/preserved, references added, and the report's notes prose. An
+// error-only report records why the skill refused to draft.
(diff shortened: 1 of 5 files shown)


internal/worker/skill_parsers.go
```diff
@@ -804,6 +804,63 @@ func (w *Worker) parseMitigationOutput(scan *db.Scan, report string, emit func(E
...
 }
 
+// parseDiscloseOutput posts a FindingNote summarising a disclose run so the
+// finding's Notes panel records that a draft was prepared, alongside the
+// verify/revalidate/patch entries (#482). The draft itself is on
+// Finding.DisclosureDraft (PATCHed by the skill via the API); this note is
+// the audit trail pointing at it: the GHSA summary, which fields were
+// patched/preserved, references added, and the report's notes prose. An
+// error-only report records why the skill refused to draft.
+func (w *Worker) parseDiscloseOutput(scan *db.Scan, report string, emit func(Event)) error {
```
jamietanna, line 814: Would it be straightforward to add a link to the drafted Disclosure? Thinking about the case where I've found a few things, triggered the Disclosure creations, and want to come back to the Finding and follow through to the resulting scan/Disclosure contents?

PR #489, opened 2026-06-23
fix: detect subproject-specific profiles during scans

### LINKED ISSUE #460: Subproject scans inherit the repo-root profile instead of detecting their own
## Summary

When a repo contains a subproject in a different language than the main project (e.g. a Ruby on Rails app with a `frontend/` directory in JS/Node), the subproject's scans run under the **main project's profile/container image**, not one matching the subproject's own language. There is no per-subproject profile detection or override mechanism.

## Current behavior

Profile (container image) selection happens once, against the **repo root**, and the resulting image is reused for every scan in the repo — root and subprojects alike:

- `RunSkill` resolves the profile using `src = <WorkRoot>/src` (the full clone root); the subproject's `SubPath` is never passed into profile selection. (`internal/worker/docker.go`)
- With no explicit profile, `DetectProfile` runs `brief` against the **whole repo root** and `matchProfile` returns the **first match** from `builtinProfiles`, ordered `php-ext → php → ruby → node`. (`internal/worker/profile.go`)
- Subprojects store a `Kind` (e.g. `npm-workspace`, `go-module`), but `Kind` is purely UI metadata and is never consulted during profile resolution. Languages are stored at the repository level only, not per-subproject.

**Example:** A Rails app with a Node `frontend/` — `brief` detects both Bundler and npm; ruby is listed before node, so ruby wins, and the Ruby image runs the Node subproject's scans too.

## Existing escape hatch (not automatic)

An operator can force a profile per-scan via `?profile=node` (`ScanOpts.Profile` → `Scan.Profile`). However, the subproject enqueue path only passes `SubPath` and `Model` (`internal/web/server.go`), so subproject scans fall through to the same root-based auto-detection. Nothing derives a profile from the subproject's language/`Kind`.

## Possible fix

In the subproject enqueue/detection path, either:

- pass `DetectProfile` the `SubPath`-joined directory so detection runs against the subproject, or
- map `Subproject.Kind` → profile name and set `ScanOpts.Profile` accordingly.

abhinavgautam01:
Fixes #460

### Summary
This PR fixes an issue where subproject scans incorrectly inherited the root repository's profile/language instead of detecting their own. 

### Changes
- Added `SubPath` to the `SkillJob` struct so the subproject path isn't lost during enqueue.
- Populated `SubPath` for both standard skill and exposure jobs (`internal/worker/skill.go`, `internal/worker/exposure.go`).
- Updated `DockerRunner.resolveProfile` in `internal/worker/docker.go` to append `SubPath` to the `src` directory when calling `DetectProfile`. This ensures `brief` detects the language ecosystem of the specific subproject folder rather than the repo root.

### Testing
- Verified that `go test ./...` passes cleanly and logic conforms to expected subproject boundaries.

### DIFF
--- a/internal/worker/claude.go
+++ b/internal/worker/claude.go
@@ -51,6 +51,7 @@ type SkillJob struct {
 	// the docker runner refuses to start hardened with ScanID == 0.
 	ScanID       uint
 	WorkRoot     string
+	SubPath      string
 	Model        string
 	Name         string
 	SkillDir     string // host absolute path to the staged skill directory
--- a/internal/worker/docker.go
+++ b/internal/worker/docker.go
@@ -108,7 +108,7 @@ func (d DockerRunner) RunSkill(ctx context.Context, sj SkillJob, emit func(Event
 	work := sj.WorkRoot
 	absWork, _ := filepath.Abs(work)
 
-	profile, image := d.resolveProfile(ctx, sj.Profile, src, emit)
+	profile, image := d.resolveProfile(ctx, sj.Profile, src, sj.SubPath, emit)
 	if sj.RequiresProfile != "" && profile != sj.RequiresProfile {
 		got := profile
 		if got == "" {
@@ -322,7 +322,7 @@ func (d DockerRunner) buildDockerArgs(absWork, image, perScanNetwork, claudeConf
 // default image); when empty, scrutineer probes the clone with `brief`
 // to auto-select. Any failure along the way falls back to the default
 // image with a log line so a missing profile never blocks a scan.
-func (d DockerRunner) resolveProfile(ctx context.Context, requested, src string, emit func(Event)) (string, string) {
+func (d DockerRunner) resolveProfile(ctx context.Context, requested, src, subPath string, emit func(Event)) (string, string) {
 	defaultImg := d.image()
 	if d.ProfilesDir == "" {
 		return "", defaultImg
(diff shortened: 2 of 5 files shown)


andrew (changes requested):
The `skill.go` and `docker.go` changes look right.

`exposure.go:182` is off though. `doExposure` clones the dependent repo (`prepareDependentSrc` at line 150 pulls `dep.RepositoryURL` into `workRoot/src`), but `scan.SubPath` is a path inside the upstream repo being scanned. Joining them points `DetectProfile` at `<dependent-clone>/<upstream-subpath>`, which won't exist, so exposure jobs would silently fall back to the default profile. Drop the `SubPath` line from the exposure `SkillJob` and leave it empty there.

A test for `resolveProfile` exercising a non-empty `subPath` would also be good so this doesn't regress.

abhinavgautam01:
thanks @andrew, 
I've removed the  SubPath  mapping from the  SkillJob  initialization in  exposure.go so it no longer breaks  DetectProfile by appending an upstream subpath to a dependent's root.

i also added a test case `( TestResolveProfile_SubPath  in  docker_test.go )` to ensure that  resolveProfile  properly respects and concatenates a non-empty  subPath  when delegating to  DetectProfile .

just pushed the fixes, ready for another look..

andrew (approved):
Thanks for the quick turnaround. Both review points addressed: the exposure `SubPath` assignment is gone, and `TestResolveProfile_SubPath` covers the new param. LGTM.

(PR description still mentions `exposure.go` but the diff no longer touches it — no need to fix, just noting for the record.)

Issue #365, opened 2026-06-11
Add a fix-validation pipeline


andrew:
Validating that a fix actually fixes a finding takes three pieces that all exist but are not composed: scans can target a branch, tag, or commit (#303); verify re-runs a finding's reproduction; fingerprints correlate findings across scans.

Compose them into one operation that takes a repo, a fix ref, and the finding id(s):

- [ ] Scan the fix ref with the default skill set
- [ ] Run finding-scoped verify against the fix ref expecting the reproduction to fail (the fix worked)
- [ ] Diff findings against the baseline scan via fingerprints: resolved, surviving, new (a fix that introduces new findings should say so loudly)
- [ ] Emit a single validation report covering all three

Related: #111 (tightening verify into a scored rubric strengthens the middle step), #13 (the same shape applied to already-published advisories).

PR #490, opened 2026-06-23
feat(web): add fix-validation pipeline

alexandre-daubois:
Part of #365

`POST /repositories/{id}/validate-fix` (repo, fix ref, finding ids) verifies each finding against the fix ref, re-scans with its producing skill, and writes one report on the anchor scan: resolved/surviving/new by fingerprint, plus the verify verdicts.

Follow-ups: no UI trigger yet, it's HTTP only; "default skill set" means the finding's producing skill, not `triage`, so the diff stays same-skill; and verify on an unmerged ref marks the finding `fixed`, which a follow-up could limit to the default branch.


andrew (approved):
Pushed a small follow-up to log the `.Find()` error in `fixValidationVerdicts` so it matches the sibling loaders.

Issue #464, opened 2026-06-21
Reconsider how SBOM import works


connorshea:
When importing an SBOM for a project, even a relatively simple Rails app, it imports _hundreds_ of gems and npm packages for scanning. For a complex production application, we are likely talking about thousands of dependencies.

A few options we should consider:

- At the least, warn the user that the SBOM contains X number of packages before we import the repos. Allow them a chance to cancel before they flood their Scrutineer instance.
- Rather than importing each immediately, have a "staging" step where the operator chooses which repos to import? This may not be practical due to the possible size of the dataset.
- Import all of the repos, but do not trigger triage automatically. Let the operator decide which repos they care about and trigger triage/scanning themselves.

I'm not sure which of these options are the right solution, but I think the current way that SBOM import works is not ideal for the operator.

abhinavgautam01:
hii @connorshea, i think the best approach is Option 3 + Option 1 combined:

1. Pre-import summary: after parsing, show the package count and let the operator confirm before committing.
2. Import all repos but defer triage scans: create the repo rows (cheap) so the SBOM view shows the full dependency tree, but don't auto-enqueue triage scans (expensive: clone + LLM per repo). The Findings/Advisories tabs will still surface existing data from already-tracked repos.
3. Add a "Trigger Scan" button on the SBOM show page: let operators selectively kick off triage, optionally scoped by the existing Direct/Transitive filter.

As a smart default, we could auto-triage only  direct  deps (typically 20-50 packages) since  classifyScope()  already distinguishes them, leaving the transitive bulk (500+) for on-demand scanning.

Option 2 (staging/cherry-pick) feels impractical at scale — the SBOM show page with scope filters already serves as the "review and select" step post-import.

The main code change is splitting  createOrTriageRepo  so the SBOM resolve path can create repos without auto-enqueueing triage. Happy to work on a PR if this direction sounds right.

andrew:
Agreed, auto-triage on direct deps only is the right default. The resolve loop in `internal/web/sboms.go` already has `p.Scope` at the point it calls `createOrTriageRepo`, so the simplest cut is a `triage bool` (or similar) argument on `createOrTriageRepo` that the SBOM path sets from `p.Scope == scopeDirect`. Transitives still get a repo row so the SBOM page is complete, they just don't enqueue.

The pre-import count + confirm is a nice addition but can be a separate PR; the triage gating is the bit that actually stops the flood.

Happy for you to pick this up.

[...]

PR #491, opened 2026-06-23
feat: Only auto-triage direct dependencies on SBOM import

abhinavgautam01:
Part of #464

### Description
This PR updates the SBOM import flow to prevent massive scan queue flooding when importing large SBOMs.

Previously, importing an SBOM would automatically enqueue triage scans for *every* parsed package, which could overwhelm the system for repos with hundreds of transitive dependencies.

Now, `createOrTriageRepo` accepts a `triage bool` parameter. During an SBOM import, we check the package scope and only pass `triage = true` for **direct dependencies** (`p.Scope == scopeDirect`). Transitive dependencies still get their repository rows created (so the SBOM view remains complete), but they skip the expensive auto-enqueuing step.

### Changes
*   Added `triage bool` to `createOrTriageRepo` signature.
*   Updated SBOM resolution to conditionally triage based on package scope.
*   Updated all other callers to default to `true`.
*   Added tests verifying that direct dependencies enqueue a scan while transitive ones do not.

### DIFF
--- a/internal/web/sboms.go
+++ b/internal/web/sboms.go
@@ -263,7 +263,7 @@ func (s *Server) resolveSBOMPackages(uploadID uint) {
 			s.DB.Model(p).Update("resolve_error", err.Error())
 			continue
 		}
-		repo, _, err := s.createOrTriageRepo(ctx, input, "")
+		repo, _, err := s.createOrTriageRepo(ctx, input, "", p.Scope == scopeDirect)
 		if err != nil {
 			s.DB.Model(p).Update("resolve_error", err.Error())
 			continue
--- a/internal/web/sboms.go
+++ b/internal/web/sboms.go
@@ -263,7 +263,7 @@ func (s *Server) resolveSBOMPackages(uploadID uint) {
 			s.DB.Model(p).Update("resolve_error", err.Error())
 			continue
 		}
-		repo, _, err := s.createOrTriageRepo(ctx, input, "", p.Scope == scopeDirect)
+		repo, _, err := s.createOrTriageRepo(ctx, input, "", p.Scope != scopeTransitive)
 		if err != nil {
 			s.DB.Model(p).Update("resolve_error", err.Error())
 			continue
--- a/internal/web/api_ecosystems_test.go
+++ b/internal/web/api_ecosystems_test.go
@@ -87,7 +87,7 @@ func TestCreateOrTriageRepo_prefetchesNewRemoteRepo(t *testing.T) {
 
 	repo, _, err := s.createOrTriageRepo(context.Background(), RepoInput{
 		CloneURL: "https://github.com/acme/widget", Owner: "acme", Name: "widget",
-	}, "")
+	}, "", true)
 	if err != nil {
 		t.Fatalf("createOrTriageRepo: %v", err)
 	}
@@ -104,7 +104,7 @@ func TestCreateOrTriageRepo_skipsPrefetchForLocalRepo(t *testing.T) {
 
 	if _, _, err := s.createOrTriageRepo(context.Background(), RepoInput{
 		CloneURL: LocalScheme + t.TempDir(), Name: "local", Local: true,
-	}, ""); err != nil {
+	}, "", true); err != nil {
 		t.Fatalf("createOrTriageRepo: %v", err)
 	}
 	if called {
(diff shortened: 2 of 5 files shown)


andrew:
Tests and lint is failing here

abhinavgautam01:
> Tests and lint is failing here

fixed them

internal/web/sboms.go
```diff
@@ -263,7 +263,7 @@ func (s *Server) resolveSBOMPackages(uploadID uint) {
 			s.DB.Model(p).Update("resolve_error", err.Error())
 			continue
 		}
-		repo, _, err := s.createOrTriageRepo(ctx, input, "")
+		repo, _, err := s.createOrTriageRepo(ctx, input, "", p.Scope == scopeDirect)
```
andrew: `classifyScope` only populates `Scope` when the SBOM carries a relationship graph. A flat-list SBOM — bare CycloneDX `components` with no `dependencies`, or SPDX without `DEPENDS_ON` edges — leaves every package's `Scope` as `""` (the `hasScope` guard in `sbomShow` already exists for exactly this case).

With `p.Scope == scopeDirect` that means a flat SBOM goes from "scan everything" on main to "scan nothing" here, silently. The operator uploads, gets repo rows, and nothing is queued.

Is that intended? Two options if not:

```go
// treat unknown scope as direct (only suppress confirmed-transitive)
repo, _, err := s.createOrTriageRepo(ctx, input, "", p.Scope != scopeTransitive)
```

or keep the strict check but flash a note on the upload page when no package has a scope ("this SBOM has no dependency graph; nothing was queued, trigger scans manually").

Either way, a test case with `Scope: ""` would pin the chosen behaviour.

andrew (commented):
The mechanics look right and the test covers both directions (direct gets a scan, transitive gets a repo row but no scan). One behavioural question inline about flat SBOMs that I'd like resolved before merging; the rest is minor.

A couple of non-blocking notes:

- `internal/web/server.go:1689` — the `triage` check lands after the default-skill lookup and the `RequiresRemote` check. Both are cheap, but `if !triage { return repo, isNew, nil }` right after the prefetch would skip the unused query and read more directly.
- The PR body says `Resolves #464`. The issue proposes three UX directions (pre-import count/confirm, staging, import-without-triage); this implements a fourth variant that fixes the queue-flooding symptom but doesn't give the operator the choice the issue is asking about. Suggest `Part of #464` so the UX discussion stays open rather than auto-closing on merge.

internal/web/sboms.go
abhinavgautam01: i switched this to treat unknown scope as direct and only skip triage for packages explicitly classified as transitive:

`p.Scope != scopeTransitive`

I also extended the resolve test with a `Scope: ""` package to assert that flat SBOM entries are linked and queued for triage..

abhinavgautam01:
> The mechanics look right and the test covers both directions (direct gets a scan, transitive gets a repo row but no scan). One behavioural question inline about flat SBOMs that I'd like resolved before merging; the rest is minor.
> 
> A couple of non-blocking notes:
> 
> * `internal/web/server.go:1689` — the `triage` check lands after the default-skill lookup and the `RequiresRemote` check. Both are cheap, but `if !triage { return repo, isNew, nil }` right after the prefetch would skip the unused query and read more directly.
> * The PR body says `Resolves #464`. The issue proposes three UX directions (pre-import count/confirm, staging, import-without-triage); this implements a fourth variant that fixes the queue-flooding symptom but doesn't give the operator the choice the issue is asking about. Suggest `Part of #464` so the UX discussion stays open rather than auto-closing on merge.

thankss, fixed

I changed SBOM package resolution to only suppress triage for confirmed transitive packages. Flat-list SBOM packages with blank scope now behave like direct dependencies, so they get repo rows and triage scans instead of silently linking without queueing work.

I also moved the early `!triage` return in `createOrTriageRepo` before the default-skill lookup as suggested.

Tests pass locally:
- `go test ./internal/web -run TestSBOMResolve_linksRepoAndEnqueuesTriage -count=1`
- `go test ./internal/web -count=1`
- `go test ./...`

andrew (approved):
All three points addressed: `p.Scope != scopeTransitive` keeps flat-list SBOMs queueing, the early `!triage` return now sits before the skill lookup, and the test pins direct / blank / transitive behaviour. Thanks!

PR #492, opened 2026-06-24
Rootless podman and SELinux support

dkw-oss:
Support podman (incl. rootless) as a container runtime, with SELinux

Adds podman alongside docker as a selectable per-scan container runtime, with
first-class rootless support and SELinux bind-mount relabeling. Rootless podman is
the recommended posture: runtime access is not root-equivalent, so a repo that
escapes the scan sandbox lands as an unprivileged sub-uid.

Runtime
- `--runtime docker|podman` (config `runtime:`); default docker, explicit opt-in,
  no auto-detection. DetectRuntime confirms the engine is reachable and probes
  rootless-ness.

Rootless
- `--userns=keep-id` maps the container user back to the host uid so the /work
  output and /claude-config resume store stay operator-owned, not a remapped
  sub-uid. A startup keep-id smoke test fails fast with a hint if /etc/subuid is
  misconfigured.
- host-gateway egress needs podman >= 4.7 (`--add-host ...:host-gateway`); startup
  warns if the version looks too old or the gateway doesn't resolve. IPv4 is
  resolved explicitly to dodge dual-stack fetch failures.

Hardened mode
- Supported under podman and verified fail-closed per scan: rootless `--internal`
  semantics vary across backends (pasta/slirp4netns/netavark), so each scan first
  proves its per-scan network blocks external egress yet still reaches the host
  proxy, and refuses to run otherwise.

SELinux
- On enforcing hosts (Fedora/RHEL/Rocky/Alma) container_t can't access the
  bind-mounted clone/output, so scans failed with EACCES. Adds `:z` relabeling to
  the runner mounts, gated by `--selinux auto|on|off` — auto detects SELinux via
  /sys/fs/selinux (engine-agnostic), so non-SELinux hosts are byte-for-byte
  unchanged — plus a fail-closed startup mount probe. `:z` over `:Z` keeps the
  report host-readable; see docs/podman.md.

Docs/tests: docs/podman.md (security model + residual gaps), threatmodel.md,
README, sample config; podman integration tests behind a build tag, vetted in CI.


andrew:
Now that `DockerRunner` drives both engines, the name (and `docker.go`, `buildDockerArgs`, `runDockerOnce`) reads a bit oddly — `--runtime podman` configuring a `DockerRunner` is going to confuse someone eventually.

Not asking for it in this PR since the rename would double the diff and obscure the actual change, but would you be up for a follow-up that moves to `ContainerRunner` / `container.go` (or `runner.go`)? `--no-docker` and `no_docker:` probably want a `--no-container` alias at the same time, keeping the old names accepted for compatibility.

Happy to do it myself after this lands if you'd rather not.

internal/worker/docker.go
```diff
@@ -482,34 +491,152 @@ func dirSize(root string) (int64, error) {
...
+		return "", noop, fmt.Errorf("hardened mode requires SkillJob.ScanID; refusing to share %s0 across scans", hardenedNetworkPrefix)
+	}
+	network := hardenedNetworkName(sj.ScanID)
+	if err := EnsureHardenedNetwork(d.Runtime, network); err != nil {
+		return "", noop, fmt.Errorf("create hardened network: %w", err)
+	}
+	cleanup := func() { _ = exec.Command(d.Runtime.bin(), "network", "rm", "--", network).Run() }
+	// docker's bridge --internal is already trusted by this path; podman's
+	// rootless network backends need the isolation proven per scan.
+	if d.Runtime.Bin == "podman" {
```
andrew: This gates on `Bin == "podman"` so rootful podman also pays the per-scan verification cost (four extra container launches). The PR body and `docs/podman.md` frame the need as rootless `--internal` varying across pasta/slirp4netns/netavark; rootful podman uses netavark+bridge much like docker.

Is running it for rootful podman deliberate belt-and-braces, or would `d.Runtime.needsKeepID()` (i.e. rootless only) be the tighter gate? Not blocking either way — fail-closed on both is defensible, just checking it's intentional.

internal/worker/docker.go
```diff
@@ -482,34 +491,152 @@ func dirSize(root string) (int64, error) {
...
+//
+//	(a) a container with no proxy env must FAIL to reach a routable public IP
+//	    (a literal address, so a pass means no IP-level egress rather than
+//	    merely blocked DNS); and
+//	(b) a container must still reach the host egress proxy through the gateway.
+//
+// Any probe that cannot even run (image won't start, curl missing) is treated
+// as a failure: the runner must never fall back to a weaker sandbox silently.
+func (d DockerRunner) verifyHardenedNetwork(network, image string) error {
+	gwTarget := "host-gateway"
```
andrew, line 596: `ResolveHostGatewayIPv4` runs here against `network`, and then `buildDockerArgs` calls it again against the same `perScanNetwork` (the `d.Hardened` branch around line 254). That's two container launches resolving the same value per hardened scan.

Returning the resolved IP from `setupHardenedNetwork` — `(network, gatewayIP, cleanup, err)` — and threading it into `buildDockerArgs` would drop one launch per scan. Small win, happy for it to land in the follow-up rename PR instead.

andrew (approved):
Read every file. The fail-closed posture is consistent throughout (keep-id smoke test, SELinux mount probe, per-scan hardened verification all refuse rather than degrade), the docker path is provably unchanged — `TestBuildDockerArgs_KeepIDGating` and `TestBuildDockerArgs_SELinuxRelabel` assert the arg vector is byte-identical with the zero-value runtime — and `docs/podman.md` writes down the residual gaps rather than glossing over them. The `:z` over `:Z` reasoning (host read-back of `/work`, overlapping `/work`+`/src` mounts, isolation resting on per-scan work roots not MCS) is correct.

Two non-blocking notes inline. Both are fine to defer to the rename follow-up if you'd rather keep this diff as-is.

dkw-oss:
@andrew will review your comments, also confirming that I have ran end-to-end test with rootless podman which produce results.

Keen to get your thoughts on this tradeoff too

### Rootless podman: what you trade off vs full --hardened

With `--hardened-rootless-runtime` you keep the container hardening (read-only
rootfs + no-new-privileges) on top of the always-on baseline (--cap-drop ALL,
non-root user, /tmp tmpfs, SELinux :z). The thing rootless can't have is full
--hardened's NETWORK isolation:

  - Enforced egress    : the proxy stays cooperative — a workload that ignores
                         HTTPS_PROXY can dial out directly (runner image is still
                         pinned/audited, so the practical exposure is bounded).
  - Per-scan isolation : concurrent scans share the default network instead of
                         each getting its own --internal network.

Why: --hardened's per-scan --internal network severs the route to the host, and
rootless puts the egress proxy across that namespace boundary, so the container
can't reach it. Scrutineer detects this and refuses the scan (fail-closed).

Options
  1. Rootless + --hardened-rootless-runtime   (recommended) — accept cooperative
     egress; keep everything else.
  2. Rootful podman / docker + --hardened — full enforcement, but runtime access
     is root-equivalent again (loses the main rootless win).
  3. Egress-gateway sidecar (not built yet) — proxy in a container on the
     --internal network → enforced egress under rootless. (@andrew thoughts?)

dkw-oss:
> Now that `DockerRunner` drives both engines, the name (and `docker.go`, `buildDockerArgs`, `runDockerOnce`) reads a bit oddly — `--runtime podman` configuring a `DockerRunner` is going to confuse someone eventually.
> 
> Not asking for it in this PR since the rename would double the diff and obscure the actual change, but would you be up for a follow-up that moves to `ContainerRunner` / `container.go` (or `runner.go`)? `--no-docker` and `no_docker:` probably want a `--no-container` alias at the same time, keeping the old names accepted for compatibility.
> 
> Happy to do it myself after this lands if you'd rather not.

Yep happy to do it and agree it should be in a separate PR

internal/worker/docker.go
dkw-oss: e552e17

internal/worker/docker.go
dkw-oss: 376d06a

andrew (approved):
Re-read everything since 2e6f52e.

e552e17 and 376d06a address both inline notes: verification is now gated on rootless only via `needsHardenedNetVerify()`, and the gateway is resolved once and threaded through `hardenedNet`, which also makes `buildDockerArgs` pure again. Nice.

`--hardened-rootless-runtime` (fa20232, a089c8c) is the right shape, and bumping the smoke timeout to a configurable 5 min default is the correct fix for the first-run keep-id remap. The ruby `GEM_HOME=/work/.gem` change (45f19a5) is a direct consequence of the read-only rootfs and belongs here.

On the tradeoff question: ship option 1 as the documented recommendation for rootless hosts, which this PR already does. The cooperative-egress gap is real but bounded (pinned/audited runner image, workload is model + cloned repo), and `docs/podman.md` says so plainly.

Option 3 is the right follow-up. After 376d06a most of the scaffolding is in place: `setupHardenedNetwork` already creates the per-scan network and returns a struct, so "start the proxy as a container attached to {internal, default} and return its in-network address" slots into the same shape. The runner's `HTTPS_PROXY` would point at the sidecar instead of `host.docker.internal`, and the existing fail-closed verification (can't reach the world, can reach the proxy) covers the sidecar case unchanged. I'll open an issue for it alongside the `DockerRunner` → `ContainerRunner` rename so neither gets lost.

a819b53 and 449fb34 (repo-id in heading, findings count) are unrelated to podman/SELinux. Not blocking on two lines of HTML, but worth keeping rider commits off large feature branches in future since the squash blame line will be misleading.

Approval stands; merging.

PR #493, opened 2026-06-24
feat(web): add fix-validation trigger to the finding page

### LINKED ISSUE #365: Add a fix-validation pipeline

alexandre-daubois:
Fix #365 

One design choice to confirm: "scan the fix ref with the default skill set" is realized by re-running the finding's producing skill, not `triage`, so the fingerprint diff compares same-skill against same-skill.

### DIFF
--- a/internal/web/server_test.go
+++ b/internal/web/server_test.go
@@ -1320,6 +1320,34 @@ func TestFindingShow_disablesVerifyActionWhenVerifyInFlight(t *testing.T) {
 	}
 }
 
+func TestFindingShow_rendersValidateFixForm(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+
+	repo := db.Repository{URL: "https://example.com/r", Name: "r"}
+	s.DB.Create(&repo)
+	scan := db.Scan{RepositoryID: repo.ID, Kind: "skill", Status: db.ScanDone, SkillName: "security-deep-dive"}
+	s.DB.Create(&scan)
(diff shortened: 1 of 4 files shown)


andrew (approved):
Wires the #490 endpoint into the finding page and closes the follow-up it flagged: a `fixed` verify verdict against a candidate ref no longer flips the finding's lifecycle, since the fix isn't on the default branch yet. The note and the validation report still capture the per-ref verdict.

Two non-blocking notes for later:
- Validation errors redirect to the repo page rather than back to the finding page; fine for now since it's the existing handler behaviour.
- The form renders even when the baseline scan has no `SkillID` (imported findings); submitting flashes the error, which works, but could be disabled up front.

Re-running the producing skill rather than `triage` keeps the fingerprint diff same-skill vs same-skill, which is the right call.

PR #494, opened 2026-06-24
feat: bucket vuln-scan findings into Findings instead of Scanners

### LINKED ISSUE #458: Should findings discovered via vuln-scan be bucketed into Findings instead of Scanners?
Right now the vuln-scan skill results in its findings going into the Scanners bucket. This doesn't really feel correct, as this is using an LLM to find vulnerabilities in a similar way to security-deep-dive. I've found 1 or 2 legitimate vulnerabilities with it from a few tries, so I think its findings probably belong in the "proper" Finding bucket?

Otherwise it can be very easy to lose the findings in the scanner noise from semgrep/zizmor, and the dedupe skill won't be able to dedupe them.

connorshea:
The `vuln-scan` skill's output should be in the Findings bucket like `security-deep-dive`, but its findings were filed under the Scanners tab alongside semgrep/zizmor noise.

This PR promotes vuln-scan into the proper Findings bucket so we can handle them like proper findings from other skills.

Fixes #458.

Generated with Claude Code, tested and reviewed by me.

### DIFF
--- a/internal/web/finding_dedup_enqueue.go
+++ b/internal/web/finding_dedup_enqueue.go
@@ -62,7 +62,7 @@ func (s *Server) autoEnqueueFindingDedup(scan *db.Scan) {
 	var openNonScanner int64
 	if err := s.DB.Model(&db.Finding{}).
 		Where("repository_id = ?", scan.RepositoryID).
-		Where(nonScannerScanFilter, deepDiveSkillName).
+		Where(nonScannerScanFilter).
 		Where("status NOT IN (" + db.ClosedFindingLifecycleSQLValues() + ")").
 		Count(&openNonScanner).Error; err != nil {
 		s.Log.Warn("auto-enqueue finding-dedup: count open findings",
--- a/internal/web/finding_dedup_enqueue.go
+++ b/internal/web/finding_dedup_enqueue.go
@@ -21,15 +21,16 @@ const dedupMinFindings = 2
 // committed. We enqueue a repository-scoped finding-dedup run only when both
 // conditions the dedup pass needs to be worth its model spend hold:
 //
-//  1. The scan is a security-deep-dive that produced at least one *new*
-//     finding. Re-observed findings keep the scan_id of the run that first
-//     created them, so counting findings with scan_id == this scan counts
-//     exactly the new rows. Nothing new means nothing fresh to dedup.
+//  1. The scan is a curated LLM audit (security-deep-dive or vuln-scan) that
+//     produced at least one *new* finding. Re-observed findings keep the
+//     scan_id of the run that first created them, so counting findings with
+//     scan_id == this scan counts exactly the new rows. Nothing new means
+//     nothing fresh to dedup.
 //  2. The repository now holds at least two open non-scanner findings in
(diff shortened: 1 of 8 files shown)


andrew (changes requested):
The `findingsBucketSkillSQL` consolidation is the right shape, and `TestVulnScanBucketedAsFinding` pins both the repo-tab path (`loadRepoFindings`) and the cross-repo index (`findingToggleCounts`) so the two definitions of "scanner" can't drift apart again. All three callers of `nonScannerScanFilter`/`scannerScanFilter` have had the now-dead bound arg dropped, and it merges cleanly with what's just landed on main.

One thing I'd like fixed before this goes in.

`internal/web/revalidate_enqueue.go:37` still gates on deep-dive only:

```go
if scan.SkillName != deepDiveSkillName || scan.BaselineScanID != nil {
```

This is the same pattern `autoEnqueueFindingDedup` had before this PR, and the PR moves dedup to `isLLMAuditSkill()` but leaves revalidate behind. If vuln-scan findings now land in the curated Findings bucket they'll sit in the human queue without the cheap revalidate pre-sort that deep-dive findings get. vuln-scan is the high-recall skill, so its High/Critical output arguably needs the classifier more, not less; without it the analyst sees untriaged candidates mixed in with triaged deep-dive rows, which is the opposite of what #458 is after.

I'd switch this to `!isLLMAuditSkill(scan.SkillName)` and add a vuln-scan case to `TestAutoEnqueueRevalidate_*` (or whatever covers it). If it's deliberately left out, the comment at lines 20-25 should say why, because as written it reads as an oversight.

Two smaller things, neither blocking:

`findingsBucketSkillSQL` can be built from the constants since Go const string concatenation works (the old `deepDiveFindingsCountSQL` already did this):

```go
findingsBucketSkillSQL = "(skill_name IN ('" + deepDiveSkillName + "', '" + vulnScanSkillName + "') OR skill_name = '' OR skill_name IS NULL)"
```

Then the "the listed names mirror the constants above" caveat in the comment goes away and a skill rename only touches one line.

And for the record: `findingsBucketSkillSQL`, `isLLMAuditSkill`, and the two raw queries in `maintainers.go`/`orgs.go` each list the skills separately. The comments cross-reference, which is fine for two entries; if a third audit skill lands, a `var llmAuditSkills = []string{...}` that the raw queries bind with `IN ?` and `isLLMAuditSkill` ranges over would collapse it to one place. Not needed here.

andrew (approved):
Thanks for the follow-up commit.

`autoEnqueueRevalidate` now gates on `isLLMAuditSkill` and the renamed `TestAutoEnqueueRevalidate_onlyHighAndCriticalFromLLMAudits` covers vuln-scan at Critical/High/Medium, so the high-recall output gets the same pre-sort as deep-dive. `findingsBucketSkillSQL` is built from the skill-name constants now too. That's everything I flagged.

Two cosmetic leftovers, not worth another push: `deepDiveFindingsCountSQL` is a slight misnomer now that it counts both audit skills, and the `isLLMAuditSkill` doc comment has an odd mid-sentence wrap after "Unlike findingsBucketSkillSQL this". Can tidy on the next pass through `server.go`.

PR #495, opened 2026-06-25
Extra fields in encrypted export

dkw-oss:
Hi @andrew 

We discussed this on the PR that originally implemented the encrypted import/export feature. We've since found it's useful to enrich the data with some additional fields and this PR reflects this change. Also make the revalidate stage optional, but default to true (to preserve the existing interface), so those without API keys can still import findings.

The change is fully backwards compatible with any exports made prior.

NOTE: This PR was branched from an unmerged feature branch covering https://github.com/alpha-omega-security/scrutineer/pull/492

~~Once the above is merged, the diff on this PR shall be much smaller~~ rebased

PR #500, opened 2026-06-26
chore: rename docker* to container*

### LINKED ISSUE #497: Rename DockerRunner to ContainerRunner now it drives podman too
Since #492, `DockerRunner` drives both docker and podman, so `--runtime podman` configures a `DockerRunner` and the helpers live in `docker.go` (`buildDockerArgs`, `runDockerOnce`, `DockerRunner.RunSkill`). That's going to confuse someone reading the code without the history.

Proposed rename, kept out of #492 to avoid doubling that diff:

- `DockerRunner` → `ContainerRunner`
- `internal/worker/docker.go` → `internal/worker/container.go` (and `docker_test.go` → `container_test.go`)
- `buildDockerArgs` → `buildRunArgs`, `runDockerOnce` → `runContainerOnce`
- `--no-docker` / `no_docker:` gain `--no-container` / `no_container:` aliases; the old names stay accepted so existing configs keep working
- `worker.HostGatewayAlias` keeps the `host.docker.internal` value (it's what both engines resolve), just the Go identifier is fine as-is

No behaviour change. @dkw-oss offered to take this in https://github.com/alpha-omega-security/scrutineer/pull/492#issuecomment refs.

dkw-oss:
Make container-runner naming runtime-agnostic

The per-scan runner drives both docker and podman, so DockerRunner is renamed to ContainerRunner (along with docker.go and the build/run helpers). 

Adds `--no-container` / `no_container` as the canonical switch, keeping `--no-docker` / `no_docker` as backwards compatible aliases, so existing invocations and config files are unaffected. 

Also reconciles the threatmodel T1/T12 statuses.

Fixes: https://github.com/alpha-omega-security/scrutineer/issues/497


andrew (approved):
Rename is complete — no `DockerRunner`/`buildDockerArgs`/`runDockerOnce` stragglers, and `git log --follow` on `container.go` reaches back through #492 so blame survives.

The back-compat is the part I checked most carefully and it holds at all three layers: the double-`BoolVar` binding parses `--no-docker` off argv into the same field, `merge` checks both set-keys so the legacy flag still beats a conflicting config value, and `Load` coalesces `no_docker` into `NoContainer` with the canonical key winning. Each path has a test. Pulling the flag wiring out into `registerFlags(fs, f)` is a nice side effect — first time that's been testable without touching `flag.CommandLine`.

The threatmodel changes go a bit past a rename but the file referenced a `ClaudeRunner` that never existed and a "planned" `DockerRunner`, so it had to move anyway. T1's container blast-radius now matches what `buildRunArgs` actually does, T12 "avoided" is accurate, and the T13 credential-in-sandbox paragraph (plus the matching comment at the `-e ANTHROPIC_API_KEY` line) is a useful design note for whoever picks up proxy-side injection.

This and #495 both touch `docs/encrypted-sharing.md` but in different hunks, so whichever lands second should rebase clean.

PR #502, opened 2026-06-26
chore(skills): use Sonnet-level agent for sbom and repo-overview

### LINKED ISSUE #501: Move `sbom` and `repo-overview` skills to Sonnet-level agent
No need to waste the tokens for the `sbom` and `repo-overview` skills, since they're each mostly just running a script.

abhinavgautam01:
Fixes #501

### Description
This PR addresses issue #501 by configuring the `sbom` and `repo-overview` skills to use the Sonnet-level agent (`claude-sonnet-4-6`).

Since both of these skills primarily execute predefined scripts under the hood (`git-pkgs sbom` and `brief --json`), there is no need to waste tokens by allocating them to a more capable/expensive model.

### Changes Made
- Added `scrutineer.model: claude-sonnet-4-6` to the metadata of `skills/repo-overview/SKILL.md`.
- Added `scrutineer.model: claude-sonnet-4-6` to the metadata of `skills/sbom/SKILL.md`.

### Testing
- Code formatting and the full test suite (`go test ./...`) have run and passed successfully.

### DIFF
--- a/skills/repo-overview/SKILL.md
+++ b/skills/repo-overview/SKILL.md
@@ -4,7 +4,7 @@ description: Run `brief --json` to produce a structured overview of the reposito
 license: MIT
 compatibility: Requires the `brief` CLI (https://github.com/git-pkgs/brief) on PATH.
 metadata:
-  scrutineer.model: claude-sonnet-4-6
+  scrutineer.model: mid
   scrutineer.version: 1
   scrutineer.output_file: report.json
   scrutineer.output_kind: repo_overview
--- a/skills/repo-overview/SKILL.md
+++ b/skills/repo-overview/SKILL.md
@@ -4,6 +4,7 @@ description: Run `brief --json` to produce a structured overview of the reposito
 license: MIT
 compatibility: Requires the `brief` CLI (https://github.com/git-pkgs/brief) on PATH.
 metadata:
+  scrutineer.model: claude-sonnet-4-6
   scrutineer.version: 1
   scrutineer.output_file: report.json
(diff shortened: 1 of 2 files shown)


andrew (changes requested):
Thanks for picking this up. One small change: use the tier name rather than a concrete model id.

```yaml
scrutineer.model: mid
```

`sbom` and `repo-overview` are script wrappers in the same bucket as `metadata`, `semgrep`, `zizmor` and `maintainers`, and those four all declare `mid`. The README documents this as the convention for lightweight skills:

> Bundled lightweight skills such as `metadata` use `mid`, while `security-deep-dive` uses `max`. The Settings page lets you map each tier to any configured model.

Pinning `claude-sonnet-4-6` directly works today but bypasses the tier mapping — an operator who points `mid` at Haiku (or a newer Sonnet) on the Settings page wouldn't get these two following along, and the id goes stale when the model is superseded.

(There are five existing skills that hardcode `claude-sonnet-4-6` the same way — that's drift from after #359 landed the tier system and worth a separate sweep, not this PR.)

abhinavgautam01:
thanks for catching this, you're absolutely right about using the tier mapping so it doesn't bypass local settings or go stale. I've updated both skills to use `scrutineer.model: mid` instead of the concrete model ID.

connorshea (approved):

connorshea:
@andrew oop, I just noticed you've been squash merging PRs on main thus far, do you want to update the repo settings so I can't mess it up in the future? 😅

PR #503, opened 2026-06-26
feat(web): add Gmail-ready disclosure HTML view for findings

### LINKED ISSUE #462: Gmail-formatted disclosure report option
This may be too much of a pain in the butt to figure out, but when a report needs to be sent over email (e.g. on a project which lacks PVR) it would be nice to not need to do as much work rewriting the Markdown into a format Gmail actually handles correctly (with bolding, monospaced text in relevant places, etc).

I am, admittedly, not totally sure what that would look like unfortunately. I've been trying my best to do that manually when converting a markdown report, but it's all manual at the moment.

I am also open to other options for solving this outside the scope of scrutineer.

alexandre-daubois:
Fix #462

### DIFF
--- a/docs/disclosure-fallback.md
+++ b/docs/disclosure-fallback.md
@@ -29,7 +29,7 @@ If the CNA wants the report submitted through a GitHub PVR on a different repo (
 
 The upstream is reachable but the maintainer has not turned PVR on. Two paths, depending on the project's apparent readiness signals from `posture`:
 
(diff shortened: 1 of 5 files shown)


andrew (approved):
This is exactly the shape I suggested on #462, thanks.

`inlineGmailStyles` gets the order right: the `<pre><code...>` block pass runs before the inline `<code>` replacer so only standalone spans pick up the inline-code style, and the regex handles both bare and `class="language-..."` fenced blocks. Body goes through `renderMarkdown`, which is goldmark without `WithUnsafe`, so any raw HTML in the draft is already escaped before the post-process touches it; the title is `HTMLEscapeString`'d separately.

Tests cover the bits that matter (no `<style>` block, all five tag types carry `style=`, the nested `</code></pre>` is collapsed, both 404 paths). The `disclosure-fallback.md` mention lands at the right spot in the email-fallback flow.

One trivial edge I'm not asking you to change: the template gates the link on the untrimmed `DisclosureDraft` while the handler trims before the 404, so a whitespace-only draft would show a link to a 404. Not a real-world case.

PR #505, opened 2026-06-26
feat: concurrent-finding log for parallel audit runs

### LINKED ISSUE #116: Concurrent-finding log for parallel audit runs
Once recon partitions a repo into focus areas and we run N deep-dives in parallel, they'll converge on the same bugs where focus areas overlap. Before the dedup judge sees anything, each in-flight agent should be able to check what its siblings have already filed.

Write each finding to `GET /repositories/{id}/findings?scan_group={uuid}` as soon as it's emitted (before the scan completes), and require the audit skills to fetch that list and include a `dup_check` field on every finding: one sentence stating which existing findings were compared against and why this one is distinct. The judge then has the agent's own reasoning to weigh, and an agent that finds bug X ten minutes after a sibling already filed it can move on instead of writing it up.

alexandre-daubois:
Fix #116

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -72,7 +72,7 @@ One row per skill execution or external import. `skill_name` / `skill_version` p
 | ref | text | Git ref to checkout after cloning. Empty means the default branch. |
 | skills_repo_sha | text | Commit of `-skills-repo` resolved at startup and stamped on every skill scan. Empty when `-skills-repo` is unset or for `import` scans. |
 | sub_path | text | Scopes code analysis to a sub-folder of the clone (monorepo packages). Empty means repo root. |
-| scan_group | text | Groups a cohort of deep-dive scans launched as one parallel batch (Scan-all-subprojects, or a single New-scan run). An in-flight skill lists its siblings' findings via `/repositories/{id}/findings?scan_group=...` to avoid re-filing what a sibling already reported. Empty when not part of a batch. |
+| scan_group | text | Groups a cohort of deep-dive scans launched as one parallel batch (Scan-all-subprojects, or a single New-scan run). Each sibling streams a finding to `POST /repositories/{id}/findings` the moment it confirms it and reads `GET /repositories/{id}/findings?scan_group=...` before reporting, so an in-flight skill sees what a sibling has filed so far — not only after that sibling finishes — and can avoid re-filing it. Empty when not part of a batch. |
 | profile | text | Runner profile that ran the scan (e.g. `php`). Empty = the default runner image. Set explicitly via `?profile=` or auto-detected from the clone by `brief` before launch; persisted so retries reuse the choice. |
 | commit | text | Git HEAD at scan time. |
 | started_at | datetime | |
--- a/docs/database.md
+++ b/docs/database.md
@@ -72,6 +72,7 @@ One row per skill execution or external import. `skill_name` / `skill_version` p
 | ref | text | Git ref to checkout after cloning. Empty means the default branch. |
 | skills_repo_sha | text | Commit of `-skills-repo` resolved at startup and stamped on every skill scan. Empty when `-skills-repo` is unset or for `import` scans. |
 | sub_path | text | Scopes code analysis to a sub-folder of the clone (monorepo packages). Empty means repo root. |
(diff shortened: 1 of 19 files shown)


andrew (changes requested):
Looks good overall, a couple of gaps before this can land:

- **Retries drop the group.** `scanRetry` and `scansRetryFailed` in `internal/web/scans.go` build `ScanOpts` without `ScanGroup`, and the bulk-retry's `Select` doesn't include `scan_group`. A failed sibling that's retried falls out of the cohort. Please thread `ScanGroup: scan.ScanGroup` through both retry paths and add `scan_group` to the Select column list, with a test that a retried scan keeps its group.
- **openapi.yaml.** `/repositories/{id}/findings` needs the new `scan_group` query param, and `FindingSummary` needs `dup_check`.

Two questions:

- The PR body says `Fix #116`, but the issue's "write each finding as soon as it's emitted (before the scan completes)" isn't here — siblings only see each other after one finishes. Is the streaming side a follow-up (in which case `Refs #116` so the issue stays open), or are we calling this done?
- `dup_check` lands in `findingSummary` but `skills/finding-dedup/SKILL.md` isn't told to read it. Intentional follow-up?

Minor: the deep-dive SKILL.md says "When `scan_group` is absent (a one-off run)" but `repoScan` always stamps one, so absent only applies to API/chained enqueues. And `spec_version` wasn't bumped — let me know if that's deliberate.

andrew (changes requested):
Thanks — the streaming side answers the #116 question and the retry/openapi/SKILL.md gaps are all closed.

One thing before this lands: reconciliation only handles streamed ∩ report, not streamed ∖ report. A finding that's POSTed mid-scan and then absent from the final `report.json` stays in the database indistinguishable from a confirmed one — `seen_count=1`, `status=open`, no marker. Don't delete it (a sibling may have stood down citing it, and deleting would lose the bug from both scans), but at the end of `parseFindingsOutput` flag it: for each finding where `scan_id = ? AND last_seen_scan_id = ?` and `fingerprint NOT IN seenThisScan`, write a `FindingHistory{Field: "retracted"}` row. Plus a test that streams F, ingests a report without F, asserts one `retracted` history row and the finding still present.

Two non-blocking notes:
- `FindingInput.severity` in openapi.yaml declares an enum the handler doesn't enforce; `PersistStreamedFinding` only checks non-empty. Same gap exists on the report path so not new, just visible now.
- `TestScansRetryFailed_preservesScanGroup` sits next to `TestScansRetryFailed_preservesEffort` with near-identical setup; 8400a045 table-ified the single-retry tests but left the bulk pair as copies.

`spec_version` in `skills/security-deep-dive/schema.json` still wasn't bumped — fine if that's only for breaking changes, just confirm.

alexandre-daubois:
`dup_check` is an optional additive field, no existing consumer breaks on it. So keeping it at 12 is deliberate, if that's fine for you as well.

andrew (changes requested):
Requesting changes for one remaining correctness issue. internal/worker/skill.go:478 only refreshes location/locations plus vid/snippet/references when a streamed finding is reconciled with the final report from the same scan. Because POST /repositories/{id}/findings accepts minimal streamed bodies and the skill may post as soon as a finding is confirmed, the final report can contain richer fields for the same fingerprint, but this same-scan path leaves the DB row stuck with the sparse streamed values. Please refresh the full final-report finding content on same-scan reconciliation (for example confidence, cwe, affected, reachability, quality_tier, trace, boundary, validation, prior_art, reach, rating, dup_check, and other parser-owned finding fields), and add a regression test where a minimal streamed finding is followed by a richer final report and the persisted row contains the final details.

alexandre-daubois:
It should be good now!

andrew (approved):
Thanks for sticking with this. `22dbb2f` closes the last gap: the `sameScan` branch in `reobserveFinding` now refreshes every field `scanFinding.toFinding` produces, and `TestPersistStreamedFinding_sameScanReconcileRefreshesFinalReportContent` table-asserts all of them on the persisted row. Cross-scan re-observation is untouched. `spec_version` staying at 12 for the additive `dup_check` is fine.

PR #506, opened 2026-06-26
chore(skills): move hardcoded sonnet pins to model tiers

### LINKED ISSUE #504: Move hardcoded claude-sonnet-4-6 skill pins to model tiers
Five bundled skills pin `scrutineer.model: claude-sonnet-4-6` directly rather than declaring a tier:

- `skills/breaking-change/SKILL.md`
- `skills/ingest/SKILL.md`
- `skills/mitigate/SKILL.md`
- `skills/release-watch/SKILL.md`
- `skills/revalidate/SKILL.md`

These all landed after #359 introduced the tier system (`mid` / `high` / `max`) but before anyone noticed the convention. The script-wrapper skills from the same era (`metadata`, `semgrep`, `zizmor`, `maintainers`) use `mid`, and #502 is moving `sbom` / `repo-overview` there too.

A concrete id bypasses the Settings-page tier mapping, so an operator who points `mid` at Haiku or a newer Sonnet doesn't get these five following along, and the id goes stale when `claude-sonnet-4-6` is superseded.

Most of these probably want `mid` (they were pinned to Sonnet to be cheaper than the `high` default), but `ingest` and `mitigate` do more reasoning than a script wrapper so worth a per-skill look rather than a blanket replace.

Spotted while reviewing #502.

abhinavgautam01:
Fixes #504

### Description
This PR addresses issue #504 by migrating five bundled skills that were still using hardcoded concrete model IDs (`claude-sonnet-4-6`) to the designated tier system introduced in #359.

This prevents the model pins from going stale and ensures they follow the proper tier mappings configured on the Settings page.

### Changes Made
- Updated `scrutineer.model` to `mid` for `breaking-change`, `release-watch`, and `revalidate`, as these are lightweight script wrappers.
- Removed the hardcoded model pin entirely for `ingest` and `mitigate`. This causes them to correctly default to the `high` tier, honoring the fact that they perform more complex reasoning (normalizing text reports and drafting mitigations/semgrep rules).

### Testing
- `go fmt ./...` and `go test ./...` have both successfully run and passed.

### DIFF
--- a/skills/breaking-change/SKILL.md
+++ b/skills/breaking-change/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   scrutineer.version: 1
   scrutineer.output_file: report.json
   scrutineer.output_kind: breaking_change
-  scrutineer.model: claude-sonnet-4-6
+  scrutineer.model: mid
   scrutineer.requires_remote: true
 ---
 
--- a/skills/ingest/SKILL.md
+++ b/skills/ingest/SKILL.md
@@ -7,7 +7,6 @@ metadata:
   scrutineer.version: 1
   scrutineer.output_file: report.json
   scrutineer.output_kind: findings
-  scrutineer.model: claude-sonnet-4-6
 ---
 
 # ingest
--- a/skills/mitigate/SKILL.md
+++ b/skills/mitigate/SKILL.md
@@ -7,7 +7,6 @@ metadata:
   scrutineer.version: 1
   scrutineer.output_file: report.json
   scrutineer.output_kind: mitigation
-  scrutineer.model: claude-sonnet-4-6
   scrutineer.requires_remote: true
 ---
 
--- a/skills/release-watch/SKILL.md
+++ b/skills/release-watch/SKILL.md
@@ -8,7 +8,7 @@ metadata:
   scrutineer.output_file: report.json
   scrutineer.output_kind: release_watch
   scrutineer.requires_remote: true
-  scrutineer.model: claude-sonnet-4-6
+  scrutineer.model: mid
 ---
 
 # release-watch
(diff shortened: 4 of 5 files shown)


connorshea (dismissed):
This LGTM but I'll defer to andrew on whether the ingest/mitigate skills are fine to have set to high

Dockerfile.runner
```diff
@@ -11,6 +11,7 @@ RUN GOBIN=/out go install github.com/git-pkgs/git-pkgs@v0.15.3 && \
     GOBIN=/out go install github.com/git-pkgs/brief/cmd/brief@v0.6.0
 
 FROM rust:1.96-trixie@sha256:c6811167278337db5f3b0234964ced5f538f154a2a20f09ec03721d7411c933d AS zizmor-build
+ENV CARGO_HTTP_MULTIPLEXING=false
```
connorshea: Is there a reason for this change?

abhinavgautam01: there was one CI check which was failing on the `cargo install --locked zizmor` step with the following error:
`curl failed [16] Error in the HTTP2 framing layer`

after some findings, i get to know that, this is a known bug where Cargo's underlying `libcurl` struggles with HTTP/2 multiplexing when downloading crates in certain CI environments. 

Adding `ENV CARGO_HTTP_MULTIPLEXING=false` disables HTTP/2 multiplexing for the build, which acts as a reliable workaround and allows the image to build successfully without dropping the connection...

connorshea: 👀 interesting, I think this should probably go in its own PR, could you move this to its own branch? 🙏

abhinavgautam01: okay i will create a separate branch for it...

connorshea (approved):

andrew (approved):
ingest and mitigate at `high` is fine.

PR #508, opened 2026-06-26
feat(web): warn when the scrutineer runner image is >=7 days old

### LINKED ISSUE #337: Should we automatically pull in new releases of the Docker image when booting?
Right now, we end up with a stale version of the Docker image unless the user explicitly runs a `docker pull`, and so I'm on a version of the image that's a few weeks out of date. Should we consider auto-pulling new versions on/before boot somehow?

connorshea:

The runner image carries the analysis toolchain (claude/semgrep/zizmor) and is pulled once, then never refreshed, so it silently drifts behind the published `:latest`. Before this, the user was not made aware of this staleness or how to resolve it.

Check on boot (never pull): in a background goroutine that can't delay startup and fails soft to silence if the registry is unreachable, compare the local image's registry digest against :latest and, when a newer build exists and the local one is >= 7 days old, log a one-line nag and show a banner on the Settings page. Auto-update is left to the operator (watchtower or --pull=always), as discussed on the issue; a README note points there.

A few screenshots (after manually modifying the stale threshold to 1 day instead of 7 to trigger the behavior):

Screenshot 2026-06-26 at 8 16 53 AM

Screenshot 2026-06-26 at 8 25 19 AM

I've also updated the "Runner image" field in the Settings page to say the specific commit SHA instead of just "latest", tell me if we want to make that part a separate PR.

Fix #337.


andrew (changes requested):
This is well done. Two test gaps before it lands:

- A Settings banner test in `settings_handlers_test.go`: call `s.SetRunnerImageStatus(worker.RunnerImageStatus{Stale: true, AgeDays: 10, PullCommand: "docker pull x"})`, render `/settings`, and assert the banner text and `data-copy` attribute appear. That covers both the template and the overlay in `settingsShow`.
- A small table test for `RunnerImageRef()` covering the `:latest` strip, the 12-char truncation, and the empty-revision fallback.

Minor: `remoteRunnerDigest` is nearly identical to `resolveBaseDigest` in `profile.go` (only the `sha256:` prefix and return signature differ). Worth having one call the other so #513 only has to extend one place.

On your question: keep the `RunnerImageRef` change in this PR, it's the same concern.

andrew (approved):
All review points addressed. The `resolveBaseDigest` dedup also picks up the Apple-runtime skopeo path from #513, which the original `remoteRunnerDigest` would have missed. LGTM.

PR #511, opened 2026-06-27
test: make git-backed tests hermetic

p-linnane:
Several test helpers shell out to git to build fixture repos but didn't isolate git's configuration, so they inherited the host's global and system config. On a machine configured to sign commits, the `git commit`/`git tag` calls trigger an interactive signing prompt partway through the test run.

This routes every git invocation in the affected tests through a small per-package `testGitEnv()` helper that pins `GIT_CONFIG_GLOBAL` to `os.DevNull`, sets `GIT_CONFIG_NOSYSTEM=1`, and supplies a fixed author/committer identity. Host and system config (including signing) are never read, and commits still succeed without relying on host config.

`clone_test.go` already neutralized global config inline; this folds that into the shared helper and applies the same treatment to the skills, web, and worker suites.

No production code changes. `go test -race`, `go vet`, and `golangci-lint run` all pass locally.

### DIFF
--- a/internal/skills/git_test.go
+++ b/internal/skills/git_test.go
@@ -79,6 +79,7 @@ func mustRun(t *testing.T, dir string, args ...string) string {
 	t.Helper()
 	cmd := exec.Command("git", args...)
 	cmd.Dir = dir
+	cmd.Env = testGitEnv()
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		t.Fatalf("git %v: %s: %v", args, out, err)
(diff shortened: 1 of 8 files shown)


connorshea:
I think one in `internal/worker/patch_gate_test.go` was missed?

p-linnane:
Thanks for the review @connorshea! I left that one bare on purpose since `git apply` never signs, so it wasn't part of the signing-prompt problem this PR set out to fix. I agree that it's probably better to add it for consistency though so bbf6612 takes care of that.

andrew (approved):

PR #513, opened 2026-06-27
feat: support Apple's container runtime (experimental)

p-linnane:
Running scrutineer's container scans on macOS currently requires running Docker Desktop or podman. Apple's container is a native, Apache-2.0 licensed alternative for macOS 26 (Tahoe) on Apple silicon, with no daemon and no Docker Desktop license. Apple supports it only on macOS 26 and won't address issues that can't be reproduced there, so older macOS is out of scope. This adds it as a third runtime so Mac operators don't need either.

Apple also runs each container in its own lightweight VM, so a guest escape is contained by the VM rather than the shared host kernel docker and podman use. `--hardened` is supported through a per-scan `--internal` vmnet host-only network, proven fail-closed before each scan as on docker/podman; the one flag Apple's CLI lacks, `--security-opt no-new-privileges`, is covered by the VM boundary.

Strictly opt-in and additive: the default stays docker, there is no auto-detection, and docker/podman arg generation is unchanged (the runtime-specific flags are gated). Experimental while it gets wider testing. See docs/apple.md for the full parity matrix and security model.

Note: Apple's container is the virtualization method I prefer to use, so I figured it couldn't hurt to offer up this PR 🙂


connorshea:
Asking mostly out of curiosity because I've wanted to see if this was better than docker for a while, what are the benefits of Apple's container system over Docker?

p-linnane:
Honestly, mostly that it feels lighter. Docker Desktop can run on Apple's Virtualization framework too, so it's not the VM tech, it's that there's no Docker Desktop app or always-on daemon, just the CLI. The other real difference is isolation. I like how `container` gives each container its own lightweight VM (isolated from each other by default), which is a nicer boundary for running untrusted code. Main catch is that it only runs on macOS 26 + Apple silicon.

andrew (changes requested):
This is very thorough; the security reasoning in `docs/apple.md` and the `threatmodel.md` updates are careful. A few small things:

- The `"con"+"tainer"` split in `runtime_test.go` (×2) and `skill_parsers_test.go` is unnecessary: `.golangci.yml` has `goconst: ignore-tests: true`. Please just write `"container"`.
- `HardeningSupportError(hardened, hardenedRootless bool)` never reads `hardened`. Drop the param.
- `ensureProfileImage` passes `--pull` to `<bin> build` whenever skopeo resolved a base digest, with no `supports*` gate, and `DetectProfile` passes `--network none` to `run`. Neither is tested for apple. Do both work on `container` 1.0.0? If yes, a note in `docs/apple.md` that profile builds were verified would do; if not, gate them.

Minor:
- Under hardened apple, `apiBase` in context.json advertises `host.docker.internal`, which doesn't resolve in the VM (works only because hardened forces everything through the proxy and the proxy recognises the alias). Non-hardened apple advertises the gateway IP. Worth a sentence in `docs/apple.md` so nobody trips over the inconsistency later.
- `routeGatewayIPv4` handles a full `/proc/net/route` line shape the awk filter never emits; harmless but the test exercises an input the production path can't produce.

Heads up on #508: its `remoteRunnerDigest` would fall into the docker branch on apple. I've asked Connor to wrap it around `resolveBaseDigest` (which you do extend here), so whichever lands second should pick that up.

p-linnane:
Thanks, this is great feedback. Addressed all of it in a follow-up commit on top of the feature commit, so the changes are easy to see on their own:

- Dropped the `"con"+"tainer"` splits for plain `"container"` (you're right, `ignore-tests` makes them pointless).
- `HardeningSupportError` now takes just `hardenedRootless`; collapsed the redundant test cases.
- Profile builds: verified on `container` 1.0.0 — both `container build --pull` and the `--network none` detect path work, so I added a note in `docs/apple.md` rather than gating.
- Documented the hardened `apiBase` / `host.docker.internal` quirk in `docs/apple.md`.
- Simplified `routeGatewayIPv4` to the single-field shape the awk emits, and the test with it.

On #508: makes sense. If it lands first I'll rebase onto the wrapped `resolveBaseDigest`.

andrew (approved):
All addressed, thanks. The capability-method gating keeps the apple divergence tidy and the docker/podman arg paths provably unchanged; docs/apple.md and threatmodel.md line up with what the code actually does. Tests and lint pass locally.

#508 is still open; `resolveBaseDigest` here already routes apple through the skopeo branch so whichever lands second should slot in.

PR #515, opened 2026-06-28
feat: auto-reject findings after N consecutive misses

### LINKED ISSUE #514: Auto-reject findings after N consecutive misses
Follow-up to #62.

`markNotObserved` in `internal/worker/skill.go` increments `Finding.MissedCount` when a rescan of the same scope doesn't re-observe a previously-open finding. Per `internal/db/db.go` the count is currently "a hint the finding may have been fixed upstream" and nothing acts on it.

Add a threshold (configurable, default off or ~3) at which a finding auto-transitions to `rejected` with a `FindingHistory` note ("not observed in N consecutive rescans"). The transition should:

- only fire when `MissedCount >= threshold` and the finding is still in an open lifecycle (`new`/`enriched`/`triaged`)
- skip findings that have ever been `reported` or `acknowledged` (a maintainer has seen it; don't auto-close on them)
- record `Source: db.SourceSystem` on the history row so it's distinguishable from an analyst rejection
- be reversible: a later rescan that re-observes the fingerprint should reopen via `reobserveFinding` (which already resets `MissedCount` to 0)

Threshold goes on `config.Config` / `scrutineer.yaml` with 0 meaning disabled.

abhinavgautam01:
Resolves #514

This PR implements an automated lifecycle transition for findings that disappear upstream. Currently, `markNotObserved` increments `Finding. MissedCount` but takes no further action, leaving stale findings open indefinitely.

This introduces a configurable threshold (via `scrutineer.yaml` or the CLI flag `-auto-reject-missed-count`) where an open finding is automatically marked as `rejected` once it is missed across `N` consecutive rescans.

### Key Changes
- **Configuration:** Added `auto_reject_missed_count` to `config.Config` and `scrutineer.sample.yaml`, alongside a corresponding CLI flag. It defaults to `0` (disabled).
- **Auto-Rejection Logic:** When an open finding (in the `new`, `enriched`, `triaged` or `ready` lifecycle state) reaches the consecutive miss threshold, its status automatically transitions to `rejected`.
- **Maintainer Safeguard:** Findings that have *ever* reached the `reported` or `acknowledged` status are proactively skipped. This ensures the system does not auto-close issues actively seen or tracked by human maintainers.
- **History Traceability:** Auto-rejections append a `FindingHistory` note stating `"not observed in <N> consecutive rescans"`. A new `db. SourceSystem` source was added to cleanly distinguish these automated transitions from human analyst reviews.
- **Reversibility:** If a later rescan observes the same fingerprint, it triggers the existing `reobserveFinding` logic, cleanly updating the finding, writing an `observed` history row and resetting the `MissedCount` to 0.

### Testing Done
- [x] Compiled and passed all formatting checks (`go fmt ./...`)
- [x] Passed all unit test suites (`go test ./...`)

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -87,31 +87,32 @@ func main() {
 // parseFlags fills defaults and CLI overrides; merge layers the config
 // file underneath any flag the user set explicitly.
 type flags struct {
-	configPath       string
-	addr             string
-	dataDir          string
-	effort           string
-	defaultModel     string
-	noContainer      bool
-	runtime          string
-	selinux          string
-	hardened         bool
-	hardenedRootless bool
-	runnerImage      string
-	profilesDir      string
-	skillsRepo       string
-	concurrency      int
-	cloneMode        string
-	scanTimeout      time.Duration
-	smokeTimeout     time.Duration
-	maxTurns         int
-	anthropicBaseURL string
-	forkOrg          string
-	metadataDir      string
-	schemaStrict     bool
-	recipientsFile   string
-	identityFile     string
-	skillLocal       skillDirs
+	configPath            string
+	addr                  string
+	dataDir               string
+	effort                string
+	defaultModel          string
+	noContainer           bool
+	runtime               string
+	selinux               string
+	hardened              bool
+	hardenedRootless      bool
+	runnerImage           string
+	profilesDir           string
+	skillsRepo            string
+	concurrency           int
+	cloneMode             string
+	scanTimeout           time.Duration
+	smokeTimeout          time.Duration
+	maxTurns              int
+	anthropicBaseURL      string
+	forkOrg               string
+	metadataDir           string
+	schemaStrict          bool
+	recipientsFile        string
+	identityFile          string
+	autoRejectMissedCount int
+	skillLocal            skillDirs
 
 	// set records which flags were passed on the command line so merge
 	// knows not to let the config file override them.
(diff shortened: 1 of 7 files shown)


andrew (changes requested):
Thanks for picking this up. One gap before this can close #514: the reversibility requirement is not implemented.

The PR description says a later rescan that re-observes the fingerprint "triggers the existing `reobserveFinding` logic, cleanly updating the finding". `reobserveFinding` does run (the fingerprint lookup at `internal/worker/skill.go:408` has no status filter), but it only resets `missed_count`, refreshes location/VID/snippet, and writes an `observed` history row. It never touches `status`. So an auto-rejected finding stays `rejected`, which is in `ClosedFindingLifecycles` and hidden from the triage funnel.

Net effect: once auto-rejected, a finding is permanently closed even if the bug reappears upstream.

A blanket `status = new` in `reobserveFinding` would be wrong though: analyst-rejected false positives would reopen on every rescan that still reports them. The reopen needs to be scoped to system-rejected findings.

Suggested approach: in `reobserveFinding`, when `existing.Status == db.FindingRejected`, look up the most recent `FindingHistory` row with `field = "status"`. If its `Source == db.SourceSystem`, restore `status` to that row's `OldValue` (so a `triaged` finding comes back as `triaged`, not demoted to `new`) and write a `SourceSystem` history row for the reopen. If the last status change was analyst or model, leave it rejected. This mirrors the `hasEverBeenReportedOrAcknowledged` pattern already in this PR.

Tests: `findings_dedup_test.go` is the place. With `w.AutoRejectMissedCount = 2`, two empty rescans, assert `rejected`; then the original report again, assert status is restored and a `SourceSystem` reopen history row exists. A second case should confirm an analyst-rejected finding does not reopen on re-observation.

abhinavgautam01:
Thanks for catching that!

I've updated `reobserveFinding` to check if a finding is `rejected` and its most recent `FindingHistory` transition was `SourceSystem`. If so, it restores the `status` to what it was before being auto-rejected (the `OldValue`) and writes a new `SourceSystem` history row for the reopen. If it was rejected by an analyst or model, it correctly leaves it as `rejected`.

I also added `TestParseFindingsOutput_autoRejectReversibility` in `findings_dedup_test.go` to cover the full lifecycle (auto-rejection -> restoration and ensuring analyst-rejections don't reopen).

andrew (approved):
Nice, that's exactly the shape I was after. Verified the round-trip restores to the pre-reject status (not `new`) and analyst rejections stay closed. Thanks for the thorough test.

PR #517, opened 2026-06-29
Fix #466: Simplify maintainers skill using summarization script

### LINKED ISSUE #466: Simplify/reduce "weight" of maintainers skill
I don't think that the `maintainers` skill is really worth the current runtime/token usage it has, to the extent that we should strongly consider slimming it down to as minimal a size as we can?

We generally don't need all the data on the project's committers from the last year, we only need the actual, "central" maintainers.

A few suggestions:

- Can we reduce the amount of work it's doing?
- Can we reduce how often it needs/tries to fallback to other sources after finding its initial set of data?
- For consideration: Do we even need to run it at all if we do not have any findings on the given repo? When mass-scanning repos, I don't necessarily want to run the maintainers skill until I know there's something worth reporting to them.
- Can we split the maintainers skill out from defining the disclosure channel? They're somewhat separate goals and I wonder if that'd ultimately save usage by reducing context window size when determining the disclosure channel.
- Reduce effort level for maintainers skill.

It is currently the third-heaviest skill in my instance, in terms of token usage and cost:

Image

abhinavgautam01:
Resolves #466.

### Summary of Changes

- **Added `summarise.py` Script**: Shifted the heavy lifting of data aggregation to a new Python script (`skills/maintainers/scripts/summarise.py`). This script queries the cached commits, issues and packages data via the Scrutineer API (introduced in PR #478), parses local `SECURITY.md`/`CODEOWNERS` files and checks the GitHub private vulnerability reporting (PVR) endpoint.
- **Updated `maintainers` Skill Instructions**: Modified `skills/maintainers/SKILL.md` to instruct the LLM to run the new script instead of making direct HTTP tool calls. The LLM now only needs to read the generated `summary.json`, which drastically reduces context size, token usage and runtime.
- The maintainers skill is now entirely manual-only and is no longer automatically enqueued by triage or security-deep-dive.

This ensures the `maintainers` skill is significantly cheaper, faster to run and won't consume unnecessary resources during standard triage runs.

### DIFF
--- a/skills/maintainers/SKILL.md
+++ b/skills/maintainers/SKILL.md
@@ -2,7 +2,7 @@
 name: maintainers
 description: Identify the real maintainers of a repository and the best way to contact them about a security issue. Distinguishes active leads from occasional contributors and bots, using commit history, issue activity, and registry ownership. Use when preparing a disclosure and needing to know who to reach.
 license: MIT
-compatibility: Needs network access to commits.ecosyste.ms, issues.ecosyste.ms, packages.ecosyste.ms, and api.github.com (to verify private vulnerability reporting before proposing the GitHub advisories channel).
+compatibility: Uses a python script (scripts/summarise.py) to aggregate cached ecosyste.ms data, parse local security files, and check PVR status, saving LLM tokens.
 metadata:
   scrutineer.version: 1
   scrutineer.output_file: report.json
@@ -29,17 +29,11 @@ You are identifying who maintains a repository so a security disclosure can reac
 
 ## Data sources
 
-Fetch each of these using your fetch tool. Each returns JSON. Include the repository URL from `context.json` as the query parameter.
+Run `python3 .claude/skills/maintainers/scripts/summarise.py`
+The script uses `context.json` to query the Scrutineer API for cached commits, issues, and packages data, parses `SECURITY.md` and `CODEOWNERS`, and hits the GitHub PVR endpoint if applicable.
 
-1. **Commits**: `https://commits.ecosyste.ms/api/v1/repositories/lookup?url={repo_url}` — who has written code, how much, past-year activity. Follow redirects.
-2. **Issues and PRs**: `https://issues.ecosyste.ms/api/v1/repositories/lookup?url={repo_url}` — who reviews, responds, closes issues. Follow redirects.
-3. **Packages**: `https://packages.ecosyste.ms/api/v1/packages/lookup?repository_url={repo_url}` — registry owners and publishers.
-
-URL-encode the repository URL before substituting it into the query string.
-
-If all three lookups return empty or 404, fall back to `git -C ./src shortlog -sne --since='1 year ago'` plus `git -C ./src log --no-merges -20 --format='%aN <%aE>'` and classify from that alone; say so in `notes`.
-
-Also read `SECURITY.md`, `.github/SECURITY.md`, `CODEOWNERS`, and `README.md` in `./src` if they exist. These often name a security contact directly.
+After running the script, read `summary.json`. It contains all the necessary data to classify the maintainers and pick a disclosure channel.
+If the API data is missing, the script will fall back to git logs, which will be present in `summary.json`.
 
 ## How to classify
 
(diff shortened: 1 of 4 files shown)


andrew (changes requested):
Thanks for picking this up. The direction is right (collapse the tool-call round-trips into one script run), but a few things need fixing before this can land.

**Blocking**

1. **`git shortlog` fallback is broken.** `subprocess.check_output(["git", "-C", "./src", "shortlog", "-sne", "--since=1 year ago"])` has no revision argument. When stdin is not a tty — which is the case under `subprocess` in the runner container — `git shortlog` reads log input from stdin instead of defaulting to `HEAD`, so it either hangs or returns empty. Append `"HEAD"`. The old SKILL.md ran this via the model's Bash tool where stdin is a tty, which is why it worked before.

2. **The script doesn't actually summarise.** It dumps the raw cached `commits_data` / `issues_data` / `packages_data` payloads verbatim into `summary.json`, then has the model read the whole file. Those payloads are the same size as what the old skill fetched directly. Collapsing ~8 tool calls into 2 is a real win, but #466's complaint is token cost, and the PR description claims this "drastically reduces context size". To actually close #466 the script should reduce `commits_data` to top-N committers by past-year count and `issues_data` to top-N closers/reviewers before writing, rather than passing the blobs through.

3. **No timeouts.** None of the `urllib.request.urlopen` calls (three API reads + PVR check) set `timeout=`. A stalled scrutineer API or api.github.com hangs the whole skill run. Add a timeout (10s is fine).

**Smaller**

- `compatibility:` documents runtime requirements (see other skills: "Needs network access to…", "Requires X on PATH"). The new value describes implementation. Suggest: `Requires python3 on PATH; needs network access to the scrutineer API and api.github.com.`
- CODEOWNERS is only checked at `./src/CODEOWNERS`; GitHub also reads `.github/CODEOWNERS` and `docs/CODEOWNERS`.
- `README.md` was in the old read-list and is now dropped. Plenty of small projects only name a contact there.
- The PR body says the skill is "gated so it only runs when there is an actual finding to report during the disclosure path" — the diff doesn't do that. It's removed from `triage` but still listed in `security-deep-dive`'s `scrutineer.requires` (skills/security-deep-dive/SKILL.md:18), so it runs on every deep-dive regardless of findings. Either the description needs correcting or there's missing wiring.
- `except urllib.error.HTTPError as e:` — `e` is unused, and `urllib.error` is only reachable because `urllib.request` happens to import it internally; add `import urllib.error` explicitly.

No external dependencies in the script, which is good — please keep it stdlib-only.

abhinavgautam01:
Thanks for the detailed review @andrew 

I've just pushed a commit that addresses all of your points:
- Added `HEAD` to the `git shortlog` fallback so it won't hang.
- The `summarise.py` script now actually parses the JSON payloads to extract and aggregate the `commits` and `issues` arrays, truncating them to just the top 20 committers and top 20 issue participants. The raw payloads are no longer passed through to the LLM.
- Added `timeout=10` to all `urllib.request.urlopen` calls.
- Explicitly added `import urllib.error` in the script.
- Updated the script to check for `CODEOWNERS` in `.github/` and `docs/` as well.
- Restored the logic to read `README.md` (and case variations) and pass it into the summary.
- Fixed the `compatibility` definition in `SKILL.md`.
- Removed `maintainers` from `skills/security-deep-dive/SKILL.md`'s `scrutineer.requires` block to ensure it is actually gated.

Let me know if this looks good to land or needs some modifications.

andrew (changes requested):
Thanks, nearly there. Most points are addressed: `HEAD` on shortlog, timeouts, explicit `urllib.error` import, `compatibility` field, CODEOWNERS paths, README restored, `maintainers` dropped from `security-deep-dive`'s `requires`. Still stdlib-only.

**One blocker remains — the truncation is dead code.**

The script reads `summary["commits_data"].get("commits", [])` and `summary["issues_data"].get("issues", [])`, but the cached payloads (verbatim `commits.ecosyste.ms` / `issues.ecosyste.ms` lookup responses, per `internal/worker/ecosystems.go:35-36,81-82`) have no `commits` or `issues` keys. Checked against the live API:

- commits lookup returns `committers` and `past_year_committers` — arrays of `{name, email, login, count}`, already aggregated and sorted by count
- issues lookup returns `maintainers`, `active_maintainers`, `issue_authors`, `past_year_issue_authors`, `past_year_pull_request_authors` — also already aggregated

So both `if commits:` / `if issues:` branches are skipped and the full raw payloads still land in `summary.json`. (The `{"commits":[{"login":"alice"}]}` fixture in `ecosystems_test.go:47` is a synthetic stub that doesn't match the real shape — likely where the confusion came from.)

Since ecosyste.ms already aggregates, no `Counter` needed; just slice the right fields:

```python
raw = get_json(f"{base_url}/commits/raw", token)
if isinstance(raw, dict):
    summary["commits_data"] = {
        "past_year_committers": (raw.get("past_year_committers") or [])[:20],
        "total_committers": raw.get("total_committers"),
    }

raw = get_json(f"{base_url}/issues/raw", token)
if isinstance(raw, dict):
    summary["issues_data"] = {
        "maintainers": raw.get("maintainers"),
        "active_maintainers": raw.get("active_maintainers"),
        "past_year_issue_authors": raw.get("past_year_issue_authors"),
        "past_year_pull_request_authors": raw.get("past_year_pull_request_authors"),
    }
```

Then drop the `Counter` import.

**Minor**

- `except urllib.error.HTTPError as e:` — `e` still unused; drop `as e`.
- With `maintainers` now in neither `triage` nor `security-deep-dive.requires`, nothing enqueues it automatically. That's "manual-only", not "gated to the disclosure path" as the PR description says. Fine for #466's purposes, but please update the description to match.

abhinavgautam01:
thanks for pointing out the API shape difference! 
You're completely right, the ecosyste.ms API already does the aggregation, so the manual counting logic was just dead code falling through to the full payload.

I've pushed a commit that fixes this:
- Replaced the `commits` and `issues` slicing with direct reads of the pre-aggregated fields (`past_year_committers`, `total_committers`, `maintainers`, `active_maintainers`, `past_year_issue_authors` and `past_year_pull_request_authors`).
- Dropped the unnecessary `Counter` import.
- Removed the unused `as e` exception variable.
- Updated the PR description to clarify that the `maintainers` skill is now strictly manual-only rather than gated.

andrew (approved):
All addressed, thanks for sticking with it.

PR #518, opened 2026-06-29
fix: SQL injection via raw {id} path param in GORM inline conditions

connorshea:
This fixes https://github.com/alpha-omega-security/scrutineer/security/advisories/GHSA-2v2f-6322-446q by resolving a SQLi hole in a few places.

I ran another scan over the entire repo to look for any similar issues, and didn't find anything else beyond what was fixed here. But given that a new instance of the problem was introduced since I initially reported it, we should probably add a lint rule or something to protect against reintroducing this issue in the future.

## Summary

Several web handlers route a record ID through GORM's inline `First`/`Delete` condition by passing `r.PathValue("id")` — a **raw string** — straight into the query. With `gorm.io/gorm` v1.31.1, a non-numeric inline-condition string is emitted as a raw SQL expression rather than a parameterized `WHERE id = ?`.

The fix parses the path id with `strconv.Atoi` up front and 404s on a non-numeric value, then queries with the `int` so GORM parameterizes it. This matches the existing `strconv.Atoi` handlers and is behavior-preserving for legitimate traffic (numeric ids unchanged; non-numeric ids already 404'd via SQL error and still do).

## Sinks fixed

`loadByID` (server.go), `packageShow`, `findingShow`, `sbomShow`, `scanReport`, `scanShow`, `findingReport`, `maintainerShow`, and `findingDisclosureHTML` (`finding_disclosure.go`).

Notes vs. the original draft of this fix:
- `findingDisclosureHTML` is a sink that was introduced *after* the fix was first drafted; it's the same vulnerability class, so it's included here.
- `sbomDelete` was already parameterized on `main` by a later transaction-based refactor, so no change was needed there.

## Codebase-wide audit

I swept the whole codebase for this class. Every other SQL sink is safe: numeric (`uint`/`int`) inline conditions are GORM-parameterized, `Where` clauses use `?` placeholders, struct conditions (`Where(db.Repository{...})`) are parameterized, and `Order`/column-name usages resolve through constants or whitelists. No further instances of this class remain.

## Other changes

Extracted `anyPackageHasScope` from `sbomShow` to keep it under the gocognit cap once the id-parse branch is added.

## Testing

- `go build ./internal/web/...` — OK
- `golangci-lint run ./internal/web/...` — 0 issues
- `go test ./internal/web/...` — pass

### DIFF
--- a/internal/web/finding_disclosure.go
+++ b/internal/web/finding_disclosure.go
@@ -5,6 +5,7 @@ import (
 	"html/template"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"scrutineer/internal/db"
@@ -59,7 +60,12 @@ const disclosureHTMLPage = `<!doctype html>
 // out as a plain email body rather than a GitHub advisory.
 func (s *Server) findingDisclosureHTML(w http.ResponseWriter, r *http.Request) {
 	var f db.Finding
-	if err := s.DB.First(&f, r.PathValue("id")).Error; err != nil {
+	id, err := strconv.Atoi(r.PathValue("id"))
+	if err != nil {
+		http.NotFound(w, r)
+		return
+	}
+	if err := s.DB.First(&f, id).Error; err != nil {
 		http.NotFound(w, r)
 		return
 	}
--- a/internal/web/finding_report.go
+++ b/internal/web/finding_report.go
@@ -3,6 +3,7 @@ package web
 import (
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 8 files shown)


andrew (changes requested):
The fix is correct and the audit holds up — I cross-checked every remaining `r.PathValue(...)` in `internal/web/` and nothing else reaches GORM unparameterized.

Two things before this lands:

**Missing regression test.** `TestLoadByID` (server_test.go:93) covers a valid id and a missing numeric id, but nothing asserts that a non-numeric id returns 404 without touching the DB. Given a new instance of this bug was introduced after the original report, a test is the minimum bar until a lint rule exists. One case on `loadByID` with `r.SetPathValue("id", "1; DROP TABLE repositories")` asserting `ok == false` and `w.Code == 404`, plus one on a `Preload`-using handler (e.g. `GET /findings/abc` → 404), would cover both code paths.

**Duplicate comment** in `sboms.go` — the "scope filter only makes sense when…" comment now appears both on `anyPackageHasScope` and inline at the call site. Drop the inline copy.

Thanks for the thorough sweep and writeup.

andrew (approved):
Regression tests cover both the generic `loadByID` path and the inline `Preload`+`First` path; duplicate comment removed. LGTM.

Issue #22, opened 2026-04-20
Feature-gate tabs and jobs for libraries vs applications


andrew:
Scrutineer currently assumes the target is a library published to a package registry with downstream dependents. Applications have neither. Feature-detect from scan results and hide or show tabs, jobs, and workflow branches accordingly.

- [ ] Detect has-packages and has-dependents from scan results
- [ ] Hide the Packages and Dependents tabs when absent
- [ ] Skip the packages and dependents jobs when absent
- [ ] Exposure analysis and VEX only when dependents exist
- [ ] Operator-focused disclosure workflow when no dependents
- [ ] Advisories, maintainers, audit, and threat model apply to both

abhinavgautam01:
#522, #523, #552 and now #553 cover the app/library gates we split out so far:

- empty dependency/dependent UI is hidden when there are no rows
- no-dependent findings use the operator/remediation workflow
- exposure controls and bundle VEX output are gated on recorded dependents
- direct CSAF export is now gated the same way in #553

the only remaining checklist nuance I see is the `packages` job. `dependents` is no longer in the default triage set, while `packages` still runs unconditionally because it is also the discovery step for whether package rows exist.

Do you want a final follow-up to gate `packages` on `brief`'s `has_packages` or is keeping it as the discovery job the intended behavior and #22 can be considered covered once #553 lands?

andrew:
Done across #522, #523, #552, #553 and the triage-skill gating. Tabs hide when empty, exposure/VEX/CSAF gate on dependents, and findings on repos without dependents get the operator-remediation workflow instead of the disclosure one.

PR #522, opened 2026-06-30
Hide empty dependencies tab on repo pages

abhinavgautam01:
## Summary

Part of #22.

This updates the repository detail page to hide the Dependencies tab when a repository has no indexed dependency rows, matching the existing feature-gating behavior for Packages and Dependents.

## Changes

- Hide the Dependencies tab and panel when no dependencies are present.
- Keep the Dependencies tab visible when non-runtime dependencies exist but are hidden by the default runtime-only filter.
- Add repo page tests for both cases.

## Testing

- `git diff --check`
- `go test ./internal/web -run 'TestRepoShow_dependenciesTab|TestRepoShow_depsCap'`
- `go test ./internal/web`
- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go test -race ./...`

### DIFF
--- a/internal/web/repo_show_test.go
+++ b/internal/web/repo_show_test.go
@@ -156,6 +156,24 @@ func TestRepoShow_dependenciesTab_plainManifestWithoutDoneScan(t *testing.T) {
 	}
 }
 
+func TestRepoShow_dependenciesTabHiddenWhenNoDependencies(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+	repo := db.Repository{URL: "https://example.com/r", Name: "r"}
+	s.DB.Create(&repo)
+
+	body := getRepoPage(t, s, repo.ID)
+	for _, hidden := range []string{
+		`id="rt5"`,
+		`id="rp5"`,
+		"No dependencies indexed yet",
+	} {
+		if strings.Contains(body, hidden) {
+			t.Errorf("repo with no dependencies should not render %q", hidden)
+		}
+	}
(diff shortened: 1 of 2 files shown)


andrew (changes requested):
Gate logic is right — `or .Deps .HiddenDeps` shows the tab iff any dependency rows exist, since `hiddenDeps` is counted unconditionally at `internal/web/server.go:1920-1923` regardless of `?deps=all`. Matches the Packages/Dependents pattern. Tests cover both edges.

One thing to clean up:

`internal/web/templates/repo_show.html:395` — the final `{{else}}` ("No dependencies indexed yet. The git-pkgs job populates this.") is now unreachable. The outer gate requires `or .Deps .HiddenDeps`; the inner chain only reaches that `{{else}}` when both are falsy, which the gate already excludes. Drop it so the chain ends at `{{else if .HiddenDeps}}...{{end}}`.

The third string in the new test's `hidden` slice (`"No dependencies indexed yet"`) is then asserting a branch that can't fire — the `rt5`/`rp5` checks already prove the panel is gone, so that one can go too.

abhinavgautam01:
thanks..
i removed the unreachable dependency empty-state branch now that the outer `or .Deps .HiddenDeps` gate prevents it from rendering and dropped the redundant test assertion. The test now just verifies that the tab and panel are absent when there are no dependency rows.

andrew (approved):
Thanks, that's the lot.

PR #524, opened 2026-06-30
feat(worker): run the egress proxy as a sidecar under rootless --hardened

### LINKED ISSUE #498: Run the egress proxy as a sidecar so --hardened works under rootless podman
Under rootless podman, full `--hardened` can't work because the per-scan `--internal` network severs the route to the host, and the egress proxy currently runs as a host process on the other side of that namespace boundary. #492 added `--hardened-rootless-runtime` as the documented compromise (read-only rootfs + no-new-privileges + workspace cap, but cooperative egress: a workload that ignores `HTTPS_PROXY` can still dial out).

The fix is to run the proxy as a container attached to two networks: the per-scan `--internal` network (so the runner can reach it) and the default network (so it can reach `*.anthropic.com` and the host skill API). Then `--hardened` works under rootless with enforced egress.

Most of the scaffolding landed in #492:

- `setupHardenedNetwork` already creates the per-scan network and returns a `hardenedNet` struct with a cleanup func; adding "start proxy container, return its in-network address" slots into the same shape
- the runner's `HTTPS_PROXY` would point at the sidecar's container name on the internal network instead of `host.docker.internal:<port>`
- `verifyHardenedNetwork` already proves the internal network blocks external egress yet reaches the proxy, so it covers the sidecar case unchanged
- the proxy needs its allowlist and bearer token passed in (env or a small config mount); it's a stdlib `http.Server` today so it can be packaged into the runner image or a tiny dedicated image

Open question: whether the sidecar reaches the host skill API via its second network leg (default → `host-gateway`) or whether the skill API also moves behind the sidecar. The former is simpler and matches what `--hardened` already does under docker.

Probably wants to land after the `ContainerRunner` rename so it doesn't churn twice.

dkw-oss:
Until now `--hardened` could not enforce egress under rootless podman. The
per-scan `--internal` network severs the scan from the host, and the in-process
allowlist proxy lives on the host across that boundary, so the scan could never
reach it — hardened scans were refused fail-closed and rootless users got the
non-network hardening only.

This runs the egress proxy as a per-scan sidecar container (`scrutineer proxy`,
from the runner image), dual-homed on the scan's `--internal` network and the
default egress network. The scan points HTTPS_PROXY at the sidecar by name; the
sidecar enforces the same allowlist and forwards out its egress leg, reaching the
host skill API over the host-gateway. Enforced egress is restored under rootless
without the scan ever reaching the host. docker and rootful podman are unchanged
(they keep the in-process host proxy).

Highlights:
- New `scrutineer proxy` subcommand; the runner image now bakes in the static
  scrutineer binary (Dockerfile.runner).
- Fail-closed throughout: the sidecar refuses to serve until it confirms it can
  reach the host skill API and resolve upstream DNS; the per-scan probe refuses
  the scan if it cannot reach the sidecar. An unsupported rootless backend yields
  a clear refusal, never a silent downgrade.
- The per-scan `--internal` network isolates one scan's sidecar from another; the
  proxy token is per-process and is explicitly not the isolation boundary.
- Allowlist denials are captured into the scan record at teardown.
- Startup smoke-tests the runner image for the scrutineer binary and sweeps
  orphan sidecars/networks left by a crashed run.
- Where the rootless backend can't forward host-gateway to the host loopback,
  `--hardened` is refused; `--hardened-rootless-runtime` still applies the
  non-network hardening.

Tests: unit coverage for config parsing, run-args, gating, sweeps, and the
host<->sidecar env contract, plus a build-tagged podman integration test (egress
blocked, scan -> sidecar -> host API, allowlist 403, teardown). See
docs/egress-sidecar.md for operator validation on a real rootless host.

Closes #498


andrew (changes requested):
This is in good shape — the threat-model paragraph, the operator validation doc, and the `TestEgressSidecarEnvContract` round-trip are exactly what I'd want to see for something this load-bearing. Tests, vet, and the full lint set all pass clean.

One blocker:

**Apple + `--hardened` regresses.** `usesEgressSidecar()` (`internal/worker/container.go:150`) and `resolveEgressSidecar()` (`cmd/scrutineer/main.go:757`) both gate on `needsHardenedNetVerify()`, but that returns true for `runtimeApple` as well as rootless podman (`internal/worker/runtime.go:80`). So under `--runtime apple --hardened`:

- `resolveEgressSidecar` calls `VerifyProxyBinary`, which runs `run --rm --pull never`. Apple's CLI doesn't accept `--pull` (per `supportsPullNever()`), so when the image is present locally, startup fails with the misleading "runner image is missing the scrutineer binary" message.
- If the image isn't local yet and `VerifyProxyBinary` no-ops, `setupHardenedNetwork` then hits `usesEgressSidecar() == true` and `startProxySidecar` runs `proxySidecarRunArgs` with hardcoded `--network podman` followed by `network connect`, neither of which Apple's `container` has — every scan fails at sidecar launch.

The doc comments on both functions already say "True only for rootless podman" so the intent is right; the predicate just doesn't match. Neither `TestUsesEgressSidecar` nor `TestResolveEgressSidecar_NoSidecarForNonRootless` has an Apple row, which is how it slipped — adding one fails:

```go
d := ContainerRunner{Hardened: true, Runtime: ContainerRuntime{Bin: runtimeApple}}
if d.usesEgressSidecar() {
    t.Error("apple + --hardened must NOT use the egress sidecar")
}
// --- FAIL
```

Suggest a dedicated `needsEgressSidecar()` (`rt.Bin == "podman" && rt.Rootless`) distinct from `needsHardenedNetVerify()`, used in both places, with `{Bin: runtimeApple}` added to both test tables. The threat-model text ("the in-process host proxy for docker, rootful podman, and Apple's `container`; a per-scan sidecar for rootless podman") is already correct, so no doc change needed.

dkw-oss:
> This is in good shape — the threat-model paragraph, the operator validation doc, and the `TestEgressSidecarEnvContract` round-trip are exactly what I'd want to see for something this load-bearing. Tests, vet, and the full lint set all pass clean.
> 
> One blocker:
> 
> **Apple + `--hardened` regresses.** `usesEgressSidecar()` (`internal/worker/container.go:150`) and `resolveEgressSidecar()` (`cmd/scrutineer/main.go:757`) both gate on `needsHardenedNetVerify()`, but that returns true for `runtimeApple` as well as rootless podman (`internal/worker/runtime.go:80`). So under `--runtime apple --hardened`:
> 
>     * `resolveEgressSidecar` calls `VerifyProxyBinary`, which runs `run --rm --pull never`. Apple's CLI doesn't accept `--pull` (per `supportsPullNever()`), so when the image is present locally, startup fails with the misleading "runner image is missing the scrutineer binary" message.
> 
>     * If the image isn't local yet and `VerifyProxyBinary` no-ops, `setupHardenedNetwork` then hits `usesEgressSidecar() == true` and `startProxySidecar` runs `proxySidecarRunArgs` with hardcoded `--network podman` followed by `network connect`, neither of which Apple's `container` has — every scan fails at sidecar launch.
> 
> 
> The doc comments on both functions already say "True only for rootless podman" so the intent is right; the predicate just doesn't match. Neither `TestUsesEgressSidecar` nor `TestResolveEgressSidecar_NoSidecarForNonRootless` has an Apple row, which is how it slipped — adding one fails:
> 
> ```go
> d := ContainerRunner{Hardened: true, Runtime: ContainerRuntime{Bin: runtimeApple}}
> if d.usesEgressSidecar() {
>     t.Error("apple + --hardened must NOT use the egress sidecar")
> }
> // --- FAIL
> ```
> 
> Suggest a dedicated `needsEgressSidecar()` (`rt.Bin == "podman" && rt.Rootless`) distinct from `needsHardenedNetVerify()`, used in both places, with `{Bin: runtimeApple}` added to both test tables. The threat-model text ("the in-process host proxy for docker, rootful podman, and Apple's `container`; a per-scan sidecar for rootless podman") is already correct, so no doc change needed.

fcd6418

andrew (approved):
Apple split is exactly right now — `needsEgressSidecar()` is rootless-podman-only and `setupHardenedNetwork` keeps the per-scan verify on Apple without ever touching the sidecar path. The extra `TestContainerRuntimeNeedsEgressSidecar` table and the `runtimePodman` constant are both nice. Tests, vet, the extended lint set, and govulncheck all clean locally. Thanks for the quick turnaround.

PR #526, opened 2026-06-30
Populate dependents table from ecosystems prefetch

abhinavgautam01:
## Summary

Part of #411

This is the first additive split from the dependents-skill removal work. It keeps the existing `dependents` skill in place, but makes the ecosyste.ms dependents prefetch also populate the typed `db.Dependent` table.

## Changes

- Add `updateDependentsTable` after the dependents ecosyste.ms cache write.
- Map upstream ecosyste.ms fields into `db.Dependent`:
  - `dependent_repos_count` -> `DependentRepos`
  - `latest_release_number` -> `LatestVersion`
  - `repository_url`, falling back to `repo_metadata.html_url`
- Canonicalize dependent ecosystem values through `db.EcosystemType`.
- Add a focused test using upstream payload field names.

## Testing

- `go test ./internal/worker -run 'Test(UpdateDependentsTable|RefreshEcosystems)'`
- `go test ./internal/worker`
- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `git diff --check`
- `go test -race ./...`

### DIFF
--- a/internal/worker/ecosystems.go
+++ b/internal/worker/ecosystems.go
@@ -255,6 +255,7 @@ func updateDependentsTable(gdb *gorm.DB, repoID uint, payload []byte) error {
 		return fmt.Errorf("decode dependents cache: %w", err)
 	}
 	rows := make([]db.Dependent, 0)
+	seen := make(map[string]bool)
 	for _, entry := range result {
 		for _, raw := range entry.Dependents {
 			var d struct {
@@ -272,6 +273,12 @@ func updateDependentsTable(gdb *gorm.DB, repoID uint, payload []byte) error {
 			}
 			if err := json.Unmarshal(raw, &d); err != nil {
 				return fmt.Errorf("decode dependent: %w", err)
+			}
+			if d.PURL != "" {
+				if seen[d.PURL] {
+					continue
+				}
+				seen[d.PURL] = true
 			}
 			repoURL := d.RepositoryURL
 			if repoURL == "" {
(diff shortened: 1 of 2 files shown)


andrew (changes requested):
This is the right shape and almost there — one gap before it's a faithful replacement for the skill path.

`updateDependentsTable` doesn't de-duplicate across packages. `fetchDependents` returns one `dependentsEntry` per published package, so a repo that publishes two packages with a shared dependent writes two `db.Dependent` rows for the same PURL. The skill handled this (`skills/dependents/SKILL.md:38`: "De-duplicate by `purl` when a dependent appears under more than one of this repo's packages"); now nothing does, and there's no unique index on `(repository_id, purl)` to catch it.

Downstream that means `finding_exposure.go:41-45` (`Order("dependent_repos desc").Limit(exposureTopN)`) burns two top-N slots on the same dependent and enqueues the same exposure scan twice, and the Dependents tab / `/api/repositories/{id}/dependents` / CSAF `product_status` all list it twice.

Fix is small: a `seen[d.PURL]` map in the inner loop (the #516 version had one keyed on `ecosystem:name`; PURL matches what the skill documents). Add a second `dependentsEntry` block to the test that repeats one of the PURLs and assert it still comes out as 2 rows.

Everything else — upstream field names, `repo_metadata.html_url` fallback, `EcosystemType`, the delete-then-`CreateInBatches` transaction — is right.

abhinavgautam01:
thanks @andrew
I added PURL-based de-duplication in `updateDependentsTable` so a dependent shared across multiple published packages only writes one `db.Dependent` row.

I also extended the test with a second `dependentsEntry` that repeats `pkg:gem/rails-x` and kept the assertion at 2 rows, so the exposure top-N/API/UI paths won't see duplicate dependents from shared package dependents.

andrew (approved):
Looks good. The PURL de-duplication gap from the earlier review is fixed, and the new test covers a dependent repeated across multiple package entries.

PR #527, opened 2026-06-30
Add perl runner profile

andrew:
The base runner only carries debian's `perl-base` (no cpan client, no `prove`, thin core library), so CPAN distributions could be read but not exercised: dependencies couldn't be installed and reproducers couldn't `use` anything beyond the project's own `lib/`.

Adds a `perl` profile with full `perl` + `cpanm` + `Module::Build` and a C toolchain for XS, with local::lib pointed at `/work/perl5` so installs land on the writable mount (same pattern as ruby's `GEM_HOME=/work/.gem`). `brief` reports no `package_managers` entry for Perl projects, so detection uses `AnyMarkers` on `Makefile.PL` / `Build.PL` / `cpanfile` / `dist.ini` / `META.{json,yml}`, ordered before `c-cpp` so a dist that also ships a generated `Makefile` routes here.

Also adds `cpan.metacpan.org` and `www.cpan.org` to the egress allowlist; the existing metacpan hosts cover the index/API but cpanm fetches tarballs from these.

Uses Debian's perl (5.40 in trixie) rather than building a pinned version. CPAN dists are broadly tolerant of interpreter version and there's no ruby-build equivalent in wide use, so floating with the runner's Debian release seemed right; easy to revisit if a specific dist needs a pin.

### DIFF
--- a/docker/profiles/perl/Dockerfile
+++ b/docker/profiles/perl/Dockerfile
@@ -1,14 +1,14 @@
 # Perl profile. Extends the scrutineer runner with a full Perl
-# interpreter, cpanm, and a C toolchain so scans of CPAN distributions
-# can install their dependency set, run the test suite, and build
-# reproducers that `use` the project's modules.
+# interpreter, cpanm-bootstrapped cpm, and a C toolchain so scans of
+# CPAN distributions can install dependencies, run the test suite, and
+# build reproducers that `use` the project's modules.
 #
 # The base runner already has perl-base (debian's stripped interpreter
-# that dpkg depends on) but no cpan client, no `prove`, and only a thin
+# that dpkg depends on) but no CPAN client, no `prove`, and only a thin
 # slice of the core library. This installs the full `perl` package from
 # Debian -- the distro version tracks upstream closely enough for
 # auditing and there is no perl-build/plenv equivalent worth pinning, so
-# the version floats with the runner's debian release rather than being
+# the version floats with the runner's Debian release rather than being
 # compiled here.
 #
 # Built on-demand by the worker; the image tag is content-addressed on
@@ -21,12 +21,14 @@ ARG RUNNER_IMAGE=ghcr.io/alpha-omega-security/scrutineer-runner:latest
 FROM ${RUNNER_IMAGE}
 
 USER root
-# `perl` pulls in the full standard library plus prove/perldoc;
-# `cpanminus` is the de-facto installer; `libmodule-build-perl` covers
-# Build.PL dists. The C toolchain plus libperl-dev and the common -dev
-# headers let XS modules compile when cpanm builds them from source --
-# XS is common enough on CPAN that a separate -ext profile (cf. python)
-# isn't worth the split.
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# `perl` pulls in the full standard library plus prove/perldoc.
+# cpanminus is only the bootstrap path for cpm; scans should use cpm
+# itself as the actively-maintained installer. The C toolchain plus a
+# broad set of headers keeps common XS modules on system libs instead of
+# falling back to Alien::* downloads from non-allowlisted hosts.
 RUN apt-get update \
  && apt-get install -y --no-install-recommends \
         perl \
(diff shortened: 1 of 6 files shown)


stigtsp:
Thanks! I'll have a look at the dockerfile and see if there are additional things that are useful.

stigtsp:
Hi, I've steered our agent used for scans to write some suggestions. Can do a PR or commit instead if you'd like?

--- 

The detection ordering, local::lib-on-`/work`, and the
noexec-`/tmp` reasoning all look right, and the "isolate the vulnerable sub
rather than boot a server" reproducer guidance is good. A few suggestions,
roughly in priority order:

1. **Consider `cpm` (App::cpm) as the installer rather than cpanm.** It's the
   actively-maintained modern client, installs in parallel, verifies checksums,
   and resolves/fetches over HTTPS from the metacpan hosts you already allowlist
   — so you could also drop `www.cpan.org` from the egress list. The prep step
   becomes roughly:

   ```bash
   cd src && cpm install -L /work/perl5 --home /work/.perl-cpm --no-test
   ```

   The `--home` matters: cpm's build dir defaults to `$HOME/.perl-cpm` with no
   env override (no `PERL_CPM_HOME`), and with `HOME=/tmp` that lands on the
   noexec tmpfs — `--home /work/.perl-cpm` keeps build scratch on the roomy
   `/work` mount. (Neither client sandboxes the install — both run dependencies'
   build code — so this is about transport/maintenance, not isolation.)

2. **Set a non-interactive build env.** Otherwise a dependency whose
   `Makefile.PL` calls `prompt()` blocks on stdin and hangs the scan in a
   headless container. In the `ENV` block:

   - `PERL_MM_USE_DEFAULT=1` — `prompt()` returns its default instead of waiting
   - `NONINTERACTIVE_TESTING=1` — skips tests that gate on interactivity
   - `AUTOMATED_TESTING=1` — standard "unattended run" signal
   - leave `RELEASE_TESTING` / `AUTHOR_TESTING` **unset** — you don't want
     author-only release tests running during a scan

3. **Broaden the build deps.** `build-essential` + `libperl-dev` are right; add
   `pkg-config` plus a handful of common `-dev` libraries so XS modules build
   against system libs. The reason to be generous: when a system lib is missing,
   modules pull `Alien::*`, which tries to download and build the library from an
   upstream host that isn't allowlisted — so a missing apt package surfaces as a
   confusing network failure instead of a clean fallback. Suggested additions:

   - `pkg-config` — build-time library discovery (most XS + all Alien probes)
   - `libxml2-dev`, `libxslt1-dev` — XML::LibXML / XML::LibXSLT
   - `libsqlite3-dev` — DBD::SQLite (very common in test suites)
   - `default-libmysqlclient-dev` — DBD::mysql / DBD::MariaDB
   - `libpq-dev` — DBD::Pg
   - `libgmp-dev` — Math::BigInt::GMP, several Crypt::* modules
   - `libbz2-dev`, `liblzma-dev`, `libzstd-dev` — Compress::Raw::* family
   - `libffi-dev`, `libpcre2-dev`, `libdb-dev` — FFI::Platypus, regex XS,
     BerkeleyDB

   (`pkg-config` + `libxml2-dev` + `libsqlite3-dev` are the table-stakes
   minimum; the rest is cheap insurance — tune to your image-size budget.)

4. **Since the scanner is an agent reading untrusted repos**, worth stating in
   PROFILE.md that everything in the tree (POD, comments, test fixtures) is data,
   not instructions — prompt-injection from source is a real risk for an agentic
   scanner. And worth confirming the usual resource limits (cpu/mem/disk/timeout)
   apply while installing deps and running the target's tests, since both execute
   untrusted code.

andrew:
@stigtsp this is great, I'll get those add in tomorrow

andrew:
Thanks @stigtsp, addressed this feedback in 806592b.

Changed:
- switched the Perl profile guidance from cpanm to cpm, with cpm bootstrapped into the profile image
- added --home /work/.perl-cpm guidance so cpm build scratch avoids HOME=/tmp
- set non-interactive Perl build env: PERL_MM_USE_DEFAULT=1, NONINTERACTIVE_TESTING=1, AUTOMATED_TESTING=1
- broadened the XS/system-library build dependencies, including pkg-config and common -dev packages
- added profile guidance that source, POD, comments, fixtures, and tests are untrusted data, and that dependency installs/tests execute untrusted code
- added META.yml marker coverage in the profile selection tests

I left the extra CPAN egress hosts out after switching to cpm, since runtime installs should resolve through the MetaCPAN mirror path already on the allowlist.

Verification: go test ./... passes.

PR #536, opened 2026-06-30
Add opencode backend (-backend opencode)

### LINKED ISSUE #211: Support codex and opencode as alternative coding harnesses

andrew:
Stacked on #535 (now merged). Adds `OpencodeHarness` implementing the `Harness` interface, registers it, bundles the `opencode` binary in `Dockerfile.runner` (sha256-pinned musl static for both arches, plus `ripgrep` so opencode's grep/glob tools don't try to download it at runtime), and adds `docs/opencode.md` plus a README section.

The interface held: nothing in `container.go`, `main.go`, or the test mocks needed touching beyond registration. The opencode-specific bits:

- **Provider-agnostic.** `EgressHosts()` returns `models.dev` plus `api.openai.com` and `*.anthropic.com`; `Env()` passes through whichever of `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `OPENCODE_CONFIG_CONTENT` / `OPENCODE_AUTH_CONTENT` the host has, and disables opencode's autoupdate/share/models-fetch. Other providers need an `egress_allow:` entry. `-model-base-url` is accepted and ignored (opencode configures per-provider endpoints via `OPENCODE_CONFIG_CONTENT`).
- **Model IDs** are opencode's `provider/model` form; `DefaultModels()` returns `anthropic/`-prefixed Claude IDs and `normalizeModelID` strips the prefix for pricing lookup.
- **Two-key state dir.** `StateEnv()` returns both `OPENCODE_CONFIG_DIR` and `OPENCODE_DB`.
- **Argv.** `opencode run --format json --auto --model provider/id [--session <id>] <prompt>`; the prompt points at `./.opencode/skill/{name}/SKILL.md` explicitly.

`ParseStream` was verified against a live opencode 1.17.12 run (full triage fan-out on base62/base62.js under `openai/gpt-5.4`): `step_start` → session, `part.type == "tool"` with `part.state.input` → tool events with args, `part.text` → text, `error.data.message` → error, `step_finish` with `part.cost`/`part.tokens` → result with usage and per-step turn count. All ten fan-out skills completed with cost and token counts on each row.

Closes #211. #239 can be closed pointing here.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -32,10 +32,9 @@ type Config struct {
 	// Backend selects the agent CLI the container runner execs:
 	// "claude" (default), "codex", or "opencode". Empty leaves the
 	// built-in default (claude). Non-claude backends require the
-	// containerised runner;
-	// --no-container with a non-claude backend is rejected at startup.
-	// Validated against worker.HarnessByName so the set of accepted
-	// values stays in one place.
+	// containerised runner: --no-container with a non-claude backend
+	// is rejected at startup. Validated against worker.HarnessByName
+	// so the set of accepted values stays in one place.
 	Backend string `yaml:"backend"`
 	// NoContainer disables the containerised runner so claude runs directly on
 	// the host (no isolation). NoDocker is the pre-rename alias, still honoured
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -30,8 +30,9 @@ type Config struct {
 	Skills       []string `yaml:"skills"`
 	SkillsRepo   string   `yaml:"skills_repo"`
 	// Backend selects the agent CLI the container runner execs:
-	// "claude" (default) or "codex". Empty leaves the built-in default
-	// (claude). Non-claude backends require the containerised runner;
+	// "claude" (default), "codex", or "opencode". Empty leaves the
+	// built-in default (claude). Non-claude backends require the
+	// containerised runner;
 	// --no-container with a non-claude backend is rejected at startup.
 	// Validated against worker.HarnessByName so the set of accepted
 	// values stays in one place.
--- a/internal/worker/harness_opencode.go
+++ b/internal/worker/harness_opencode.go
@@ -203,6 +203,17 @@ func (OpencodeHarness) StateEnv(containerPath string) []string {
 	}
 }
 
+func (OpencodeHarness) DefaultModels() []ModelDefault {
+	// opencode is provider-agnostic; it drives whichever provider the
+	// operator's OPENCODE_CONFIG_CONTENT / auth config points at. The
+	// default matches the credentials Env passes through with no extra
+	// config: with ANTHROPIC_API_KEY set, opencode drives Anthropic and
+	// the claude model list applies; an operator pointing it at a
+	// different provider sets models: in config. Reusing the claude
+	// list keeps the pricing tripwire satisfied without a second copy.
+	return ClaudeHarness{}.DefaultModels()
+}
+
 func (OpencodeHarness) AccountErrorText(s string) string {
 	text := strings.TrimSpace(s)
 	if text == "" {
--- a/internal/worker/harness_opencode.go
+++ b/internal/worker/harness_opencode.go
@@ -170,7 +170,10 @@ func (OpencodeHarness) EgressHosts() []string {
 	return []string{"models.dev", "api.openai.com", "*.anthropic.com"}
 }
 
-func (OpencodeHarness) Env(baseURL string) []string {
+// Env returns opencode's container environment. baseURL is accepted for
+// interface symmetry and ignored: opencode has no single base-url
+// override; the operator sets it per-provider via OPENCODE_CONFIG_CONTENT.
+func (OpencodeHarness) Env(_ string) []string {
 	env := []string{
 		"OPENCODE_DISABLE_AUTOUPDATE=1",
 		"OPENCODE_DISABLE_MODELS_FETCH=1",
@@ -186,12 +189,6 @@ func (OpencodeHarness) Env(baseURL string) []string {
 		if os.Getenv(k) != "" {
 			env = append(env, k)
 		}
-	}
-	if baseURL != "" {
-		// opencode has no single base-url override; the operator sets
-		// it per-provider via OPENCODE_CONFIG_CONTENT. baseURL is
-		// accepted for interface symmetry and ignored.
-		_ = baseURL
 	}
 	return env
 }
--- a/internal/worker/harness_opencode.go
+++ b/internal/worker/harness_opencode.go
@@ -216,8 +216,9 @@ func (OpencodeHarness) Env(_ string) []string {
 	// container is the sandbox. OPENCODE_PERMISSION is not set because
 	// opencode JSON-parses it and there is no scalar "allow all" value.
 	env := []string{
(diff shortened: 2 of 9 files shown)


internal/worker/harness_opencode.go
```diff
@@ -0,0 +1,188 @@
...
+			emit(Event{Kind: KindError, Text: "stream read: " + readErr.Error()})
+			return
+		}
+	}
+}
+
+// opencodeLine is the subset of `opencode run --format json` event
+// fields the harness needs. The shape is {type, sessionID, ...} per
+// packages/opencode/src/cli/cmd/run.ts.
+type opencodeLine struct {
```
alexandre-daubois, line 77: The parser reads `text`, `tool`, `input` and `error` as top-level fields, but `opencode run --format json` nests them under `part`. In `run.ts` (v1.17.12) `emit` writes `{type, timestamp, sessionID, ...data}` with `data = {part}` for text/reasoning/tool_use and `{error}` for errors, so text is at `part.text`, the tool name at `part.tool`, the input under `part`, and `error` is an object, not a string. Net effect: every text and tool_use event comes out empty (they match the typed `case`, never the raw-line `default`) and errors fail to unmarshal and fall back to a raw line, so the scan log is blank for the whole opencode backend. We should decode a `part` sub-struct and an object `error`, and guard `part.text != ""` like `CodexHarness`.

internal/worker/harness_opencode_test.go
```diff
@@ -0,0 +1,169 @@
...
+		if want && got == "" {
+			t.Errorf("AccountErrorText(%q) = empty, want non-empty", in)
+		}
+		if !want && got != "" {
+			t.Errorf("AccountErrorText(%q) = %q, want empty", in, got)
+		}
+	}
+}
+
+func TestOpencodeHarness_ParseStream(t *testing.T) {
```
alexandre-daubois, line 128: The fixture here is flat, but opencode nests everything under `part`, so this test passes on a shape opencode never emits and can't catch the empty-event bug above. Maybe use real `part`-nested lines (and an object `error`), and we could assert the text, tool name and error message come through non-empty.

internal/worker/harness_opencode.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	switch {
+	case ev.Type == "step_start" && ev.SessionID != "":
+		emit(Event{Kind: KindSession, SessionID: ev.SessionID})
+	case isOpencodeToolEvent(ev):
+		name := ev.Part.Tool
+		if name == "" {
+			name = ev.Part.Name
+		}
+		emit(Event{Kind: KindTool, Tool: name, Text: summariseInput(name, ev.Part.Input)})
```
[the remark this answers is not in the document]
andrew: Fixed in the base branch (464d046 on #535): `summariseInput` now matches tool names case-insensitively and accepts `path` alongside `file_path`, so opencode (and codex) get the concise summary.

internal/worker/harness_opencode.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+	}
+	if baseURL != "" {
+		// opencode has no single base-url override; the operator sets
+		// it per-provider via OPENCODE_CONFIG_CONTENT. baseURL is
+		// accepted for interface symmetry and ignored.
+		_ = baseURL
+	}
+	return env
+}
```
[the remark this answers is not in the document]
andrew: Fixed in 3ee4a14: parameter renamed to `_` and the reason moved to the doc comment.

internal/config/config.go
```diff
@@ -30,8 +30,9 @@ type Config struct {
 	Skills       []string `yaml:"skills"`
 	SkillsRepo   string   `yaml:"skills_repo"`
 	// Backend selects the agent CLI the container runner execs:
-	// "claude" (default) or "codex". Empty leaves the built-in default
-	// (claude). Non-claude backends require the containerised runner;
+	// "claude" (default), "codex", or "opencode". Empty leaves the
+	// built-in default (claude). Non-claude backends require the
+	// containerised runner;
 	// --no-container with a non-claude backend is rejected at startup.
```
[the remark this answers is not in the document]
andrew: Fixed in 3ee4a14.

internal/worker/harness_opencode.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (OpencodeHarness) DefaultModels() []ModelDefault {
+	// opencode is provider-agnostic; it drives whichever provider the
+	// operator's OPENCODE_CONFIG_CONTENT / auth config points at. The
+	// default matches the credentials Env passes through with no extra
+	// config: with ANTHROPIC_API_KEY set, opencode drives Anthropic and
+	// the claude model list applies; an operator pointing it at a
+	// different provider sets models: in config. Reusing the claude
+	// list keeps the pricing tripwire satisfied without a second copy.
+	return ClaudeHarness{}.DefaultModels()
```
alexandre-daubois: `DefaultModels()` returns bare claude ids like `claude-opus-4-6`, but opencode's `--model` wants `provider/model` form, like the docs, the README and `TestOpencodeHarness_Args` all use. So a default install runs `opencode run --model claude-opus-4-6`, which opencode can't route. You may return `anthropic/`-prefixed ids instead? They won't match the bare-keyed `modelPricing` table though, so pricing needs the prefixed keys too.

internal/worker/harness_opencode.go
andrew: The `part` nesting was added in 422107e; 591d85b then fixed the next layer down (`part.state.input` for tool args and `error.data.message` for provider errors).

internal/worker/harness_opencode_test.go
andrew: Rewritten in 591d85b to the real v1.17.12 shapes (`part.state.input`, object `error` with `data.message`, `step_finish` with cost/tokens) and now asserts the tool arg, error text, and usage come through non-empty.

internal/worker/harness_opencode.go
andrew: Done in 591d85b: `DefaultModels()` now returns `anthropic/`-prefixed IDs. Rather than duplicate pricing rows, `normalizeModelID` strips a leading `provider/` before the lookup, so the tripwire and `costFromUsage` both resolve; `TestOpencodeHarness_DefaultModelsUseProviderPrefix` covers it.

PR #539, opened 2026-07-01
Add Sonnet 5 model option

andrew:
- add Sonnet 5.0 (`claude-sonnet-5`) to the built-in model picker
- document the option in README and sample config examples
- add a regression test that the built-in list accepts the Sonnet 5 model id


connorshea (approved):
LGTM with the caveat that we should test and make sure this actually works if we haven't bumped Claude Code to the latest release. I saw something yesterday about needing a newer version for Sonnet 5 to be available, but maybe that's only for the UI picker.

PR #545, opened 2026-07-01
fix: prefer active scan for repo list status badge

p-linnane:
The Repositories list showed a repo as `queued` when it had a newer queued scan, even though an older scan was still `running`. The repo's own Scans tab correctly showed `running`, so the list and the detail page disagreed.

The repository-level status badge now prefers active work over recency, ranking each repo's scans `running > queued > paused > newest terminal` in a single batched query and driving the badge from that. The "Last scan" column is unchanged and still reflects the newest scan for recency. Ranking is computed from `status` rather than the denormalized `status_priority` column so a stale value cannot skew the badge, and the precedence is documented to stay in lockstep with `db.StatusPriorityFor`.

Covered by a new table-driven test (`TestRepoList_statusScanPrecedence`) exercising each precedence rung plus the terminal-only fallback, and the existing list-page query-count bound still holds (the badge adds one batched query, independent of row count).

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -558,6 +558,7 @@ func paginate(r *http.Request, total int64) Page {
 type repoRow struct {
 	db.Repository
 	LastScan      *db.Scan
+	StatusScan    *db.Scan
 	FindingsTotal int
 	DiskBytes     int64
 	// Branches lists the distinct non-default refs this repo has been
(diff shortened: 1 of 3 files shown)


andrew (approved):
Ran the precedence test locally. The window-function ranking matches StatusPriorityFor's ordering and the timestamp column staying on LastScan is locked by the test.

PR #546, opened 2026-07-02
feat: deeper scanning of ruby native extensions and rails

dkw-oss:
**DRAFT for feedback**

@andrew we'd welcome any input on this, we're playing with it currently

Two Ruby profiles that supersede the base `ruby` runtime, plus selection
wiring in internal/worker/profile.go (+ tests).

- ruby-ext   — Ruby built from source under ASan/UBSan; native gems (C/C++,
  or Rust via rb-sys) are scanned for memory-safety/UB. Auto-selected on
  spec.extensions, ext/**/extconf.rb, or ext/**/Cargo.toml.
- ruby-rails — stock Ruby + Brakeman (Rails SAST). Auto-selected on
  config/application.rb.

Both are strict supersets of `ruby`, so a mis-route costs build time, not
coverage.

ruby-ext sanitizer-recipe fixes:
- Scope the fatal sanitizer options (abort_on_error / halt_on_error /
  RUBY_FREE_AT_EXIT) to scan time, not the instrumented `make` — the
  interpreter build was aborting on benign UB/leaks (exit 2).
- `-fno-sanitize=alignment` quiets the interpreter's benign iseq alignment
  UB; a post-build rbconfig rewrite re-enables the check plus
  -fno-sanitize-recover for extensions, since mkmf ignores env CFLAGS
  (unlike the autoconf/distutils-based php-ext / python-ext profiles).


internal/worker/profile.go
```diff
@@ -85,6 +106,42 @@ var builtinProfiles = []Profile{
...
+		// Markers (OR): the gemspec's spec.extensions is RubyGems' own
+		// definition of "has native code" and is authoritative; the bounded
+		// ext/ walks catch the conventional ext/<name>/extconf.rb (mkmf, also
+		// used by rb-sys Rust) and ext/<name>/Cargo.toml (Cargo-native gems)
+		// at any depth; the root extconf.rb covers the older single-dir style.
+		Name: "ruby-ext",
+		AnyMarkers: []ProfileMarker{
+			{Path: "*.gemspec", Glob: true, Contains: ".extensions"},
+			{Path: "ext/extconf.rb", Walk: true},
+			{Path: "ext/Cargo.toml", Walk: true},
```
andrew: The `ext/Cargo.toml` Walk marker has no `Ecosystem`/`Ecosystems` constraint, and `ruby-ext` is registered ahead of the `rust` profile (line 160-ish) in `builtinProfiles`. I confirmed by calling `matchProfile` directly against a synthetic repo with `package_managers: [{"name": "Cargo"}]` and a nested `ext/vendored-thing/Cargo.toml` — it returns `"ruby-ext"`, not `"rust"`.

Unlike its relationship to the base `ruby` profile, `ruby-ext` is *not* a superset of `rust` (no Miri, only a minimal rustc for rb-sys shims), so a pure-Rust project with any `ext/` directory containing a nested crate silently loses Rust-specific sanitizer/Miri coverage for the whole scan. Might be worth gating this marker on an ecosystem check, or scoping the walk root more narrowly.

internal/worker/profile.go
```diff
@@ -85,6 +106,42 @@ var builtinProfiles = []Profile{
...
+	},
+	{
+		// Before ruby: a Rails app (config/application.rb is the canonical
+		// boot file) also gets Brakeman, the Rails-specific SAST, on top of
+		// the ruby runtime. Like ruby-ext this is a superset of the ruby
+		// profile. Marker-only so it does not collide with ruby's Bundler
+		// ecosystem in the registry-sanity test.
+		Name: "ruby-rails",
+		Markers: []ProfileMarker{
+			{Path: "config/application.rb"},
```
andrew: Same structural gap as the `ext/Cargo.toml` marker: `config/application.rb` has no ecosystem constraint and `ruby-rails` is checked ahead of every other language profile. Lower confidence here since the filename is fairly Rails-specific, but any repo (regardless of detected package manager) that happens to contain that exact path gets routed to `ruby-rails` first.

internal/worker/profile.go
```diff
@@ -85,6 +106,42 @@ var builtinProfiles = []Profile{
...
+		// the scan's profile) re-pick ruby-ext and actually reproduce an
+		// ASan crash, instead of landing on stock ruby and reporting the
+		// finding inconclusive.
+		//
+		// Markers (OR): the gemspec's spec.extensions is RubyGems' own
+		// definition of "has native code" and is authoritative; the bounded
+		// ext/ walks catch the conventional ext/<name>/extconf.rb (mkmf, also
+		// used by rb-sys Rust) and ext/<name>/Cargo.toml (Cargo-native gems)
+		// at any depth; the root extconf.rb covers the older single-dir style.
+		Name: "ruby-ext",
```
andrew, line 144: `ruby-ext` is registered ahead of `ruby-rails`, so a Rails app that also ships a native extension (has both `spec.extensions`/`ext/**/extconf.rb` and `config/application.rb`) matches `ruby-ext` only — first-match-wins means Brakeman never runs. Confirmed against the Dockerfiles: `docker/profiles/ruby-ext/Dockerfile` has no Brakeman install, so the two "superset" profiles aren't supersets of each other, and this ordering silently drops Rails-SAST coverage for repos that qualify for both.

docker/profiles/ruby-ext/PROFILE.md
```diff
@@ -0,0 +1,193 @@
...
+RUBY_FREE_AT_EXIT=1              free the VM at exit so shutdown UAF/leaks are attributable
+ASAN_OPTIONS=
+  detect_leaks=0                 Ruby has known startup allocations; flip to 1 only when
+                                 chasing a focused suspect (with RUBY_FREE_AT_EXIT=1)
+  detect_stack_use_after_return=0  Ruby's conservative GC scans the C stack for VALUEs;
+                                 ASan's fake stack hides them and causes spurious GC frees.
+                                 Keep 0 unless you understand the interaction.
+  abort_on_error=1               a detected bug is a hard, pinpointed stop
+  symbolize=1 / print_summary=1  readable trace + one-line SUMMARY
+UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1   extensions are built -fno-sanitize-recover,
```
andrew: This documents `UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1`, but the Dockerfile's actual scan-time ENV (line 238, from the "absan/ubsan flag passing fixes" commit) sets `halt_on_error=0:abort_on_error=1`. The scanning agent's only reference for reasoning about sanitizer behavior is this doc — right now it describes the pre-fix behavior that commit specifically eliminated.

docker/profiles/ruby-ext/Dockerfile
```diff
@@ -0,0 +1,240 @@
...
+ && /usr/local/ruby-asan/bin/ruby -v
+
+# Give mkmf-built extensions the flags the interpreter build could not carry
+# (see the note above): rewrite the RECORDED `cppflags` in rbconfig so extension
+# compiles get the alignment check back AND -fno-sanitize-recover=undefined (UB
+# aborts = the evidence). This edits only the recorded config — the interpreter
+# binary is already built and stays quiet. It aborts if the entry's shape ever
+# changes, so it fails loudly instead of silently no-op'ing (the trap the env
+# CFLAGS fell into), then re-reads CPPFLAGS to assert the result.
+RUN /usr/local/ruby-asan/bin/ruby -rrbconfig -e 'path=$LOADED_FEATURES.grep(/rbconfig\.rb$/).first; cfg=File.read(path); cfg.sub!(/^(\s*CONFIG\["cppflags"\] = ".*)"\s*$/){ $1.gsub(" -fno-sanitize=alignment","") + " -fno-sanitize-recover=undefined\"" } or abort("ruby-ext: cppflags entry not found in #{path}"); File.write(path, cfg)' \
```
andrew, line 180: This patches the recorded lowercase `CONFIG["cppflags"]` entry via regex, but the very next line verifies success by reading the separate, uppercase `RbConfig::CONFIG["CPPFLAGS"]`. That relies on an unverified assumption that the two are linked by a `$(cppflags)` macro re-expanded at Ruby load time. It held on a locally-testable Ruby build, but I couldn't verify it against the actual Debian/glibc Ruby 3.4.9 rbconfig.rb template this container produces — worth double-checking against a real build of this image, since if the assumption doesn't hold, extensions could silently compile without the sanitizer-recover flags this whole mechanism exists to guarantee.

docker/profiles/ruby-rails/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+ARG RUBY_BUILD_REF=v20260520
+# Brakeman is pure Ruby; pin it like the rest of the image's tooling. Bump as
+# new releases land.
+ARG BRAKEMAN_VERSION=6.2.1
+
+USER root
+# Same toolchain as the ruby profile: ruby-build compiles CRuby from source, so
+# it needs the C toolchain plus the headers Ruby links against. The toolchain
+# also lets Bundler build a Rails app's native gem dependencies in-place.
+RUN apt-get update \
```
andrew: This apt-get + ruby-build-compile block (through line 56) is byte-for-byte identical to `docker/profiles/ruby/Dockerfile`'s equivalent block, and `RUBY_VERSION`/`RUBY_BUILD_REF` are independently hardcoded in both files, kept in sync only by comment ("kept in lockstep... bump both together"). Since ruby-rails is documented as using "the same interpreter built the same way," building FROM the already-built `ruby` image (passed as a build-arg, same pattern as `RUNNER_IMAGE`) would avoid recompiling CRuby from scratch a second time and remove the drift risk.

internal/worker/profile.go
```diff
@@ -85,6 +106,42 @@ var builtinProfiles = []Profile{
...
+		AnyMarkers: []ProfileMarker{
+			{Path: "*.gemspec", Glob: true, Contains: ".extensions"},
+			{Path: "ext/extconf.rb", Walk: true},
```
andrew, line 147: These two Walk markers (`ext/extconf.rb`, `ext/Cargo.toml`) share the same `ext/` root but each triggers its own independent `filepath.WalkDir` via `walkMarkerPresent`. When neither exists (the common non-Ruby-native-extension case, evaluated on every scanned repo since this profile has no ecosystem gate), the subtree gets walked twice from scratch instead of once checking for either basename in a single pass.

Separately: profile detection isn't cached across the scan → revalidate → verify pipeline (confirmed via `container.go`/`revalidate_enqueue.go` — both auto-chained jobs re-run `DetectProfile` from scratch against unchanged repo state), so this doubled walk cost is paid repeatedly per finding, not once per scan.

internal/worker/profile.go
```diff
@@ -230,18 +280,83 @@ func anyMarkersMatch(markers []ProfileMarker, srcDir string) bool {
...
+			return nil // unreadable entry: skip it, don't abort the walk
+		}
+		if d.IsDir() {
+			if strings.Count(filepath.Clean(path), string(os.PathSeparator))-rootDepth > markerWalkMaxDepth {
+				return fs.SkipDir
+			}
+			return nil
+		}
+		files++
+		if files > markerWalkMaxFiles {
```
andrew: `markerWalkMaxFiles` (the 50k-file safety cap) has no test, unlike the sibling depth-bound cap which is explicitly covered by `TestWalkMarkerPresent_bounded`. Worth a test creating >50k entries to confirm the walk actually aborts via `SkipAll` rather than scanning unbounded.

andrew (commented):
Ran a multi-angle review over the diff. Findings below as inline comments, most severe first. Verified the Cargo.toml misroute by running `matchProfile` directly against a synthetic repo.

dkw-oss:
Thanks @andrew
All 8 addressed: 7 fixed, 1 deferred, most-severe first.

1. **`ext/Cargo.toml` → pure-Rust misroute.** Fixed — gated the walk marker on an
   `rb-sys` mention. Chose a content gate over an ecosystem gate (which breaks bare-gem
   detection when brief reports no Bundler) or a narrower root (`ext/` is the Ruby
   convention); real Cargo-native gems also ship extconf.rb/`spec.extensions`, so the
   sibling markers back it up. +tests (Cargo+bare→rust, Cargo+rb-sys→ruby-ext).
   
2. **`config/application.rb` ungated.** Fixed — now requires `Rails::Application`. +neg test.

3. **ruby-ext drops ruby-rails' Brakeman.** Fixed — ruby-ext now installs Brakeman, so
   it's a genuine superset of ruby-rails and the ordering keeps Rails SAST. +pin-parity test.
   
4. **Stale UBSAN_OPTIONS doc.** Fixed — matches the Dockerfile (`halt_on_error=0:abort_on_error=1`).

5. **cppflags verified via the `CPPFLAGS` proxy.** Added a build-time gate that compiles a
   probe extension and greps the real `make V=1` line (your §5 check, automated on every
   build). Confirmed: the ruby-ext image builds clean, so the gate passes — the rewrite
   reaches the mkmf compile line on the real Debian Ruby 3.4.9 you flagged, and any future
   rbconfig change that broke it would fail the build.
   
6. **ruby-rails duplicates ruby's CRuby build.** Fixed as suggested — builds `FROM` the ruby
   profile image (new `BaseProfile` field); dropped the duplicated toolchain + version pins.
   Dockerfile 76→46 lines.
   
7. **Double walk + no detection caching.** Deferred, tracked in https://github.com/alpha-omega-security/scrutineer/issues/548. 
   Traced it: revalidate/verify are separate
   DB-backed jobs that re-run `brief` only because the derived scan carries no profile; the
   fix is to copy the parent's resolved profile onto them via the existing `Scan.Profile`
   column + detection-skip path (also makes verify reproduce on the exact image the scan
   used). Pipeline plumbing + the auto-enqueue paths, so its own PR, not this branch. Double
   walk left as-is (only costs on an `ext/` dir without extconf.rb; `#1` gave the two markers
   different predicates, so merging isn't worth it).
   
9. **`markerWalkMaxFiles` untested.** Fixed — extracted `walkMarkerPresentBounded` + a
   deterministic cap test (no 50k files).

Green locally, and all four profile images (runner → ruby → ruby-rails, and ruby-ext) build clean.

andrew (approved):
All eight addressed; verified each fix locally and checked b7b486e for regressions. The BaseProfile chain and the mkmf build-time probe are both nice.

flavorjones:
@dkw-oss A couple of comments after reading this:

- My local ASan config for Nokogiri also sets `use_sigaltstack=0`. Though it may a bit out of date because I mostly use Valgrind, I think this was necessary to avoid weird stack overflow false positives.
- Some extensions' test suites may blow up under ASan (or Valgrind, or memory debuggers generally) when enumerators are being used -- again to avoid false "stack overflow" errors when jumping around during iteration. In the Nokogiri test suite I [skip enumerator-related tests](https://github.com/sparklemotion/nokogiri/blob/6b0f8ee5a68cb064c9675771ea6e8f5a2823e826/test/xml/test_node_set.rb#L885) when running under ASan.

PR #551, opened 2026-07-02
feat: auto-resume scans after a Claude rate-limit reset

p-linnane:
When claude-code hits an account-level rate or usage limit on a Claude subscription token, scrutineer now pauses the triggering scan and the rest of the queued batch, reads the reset time from the `rate_limit_event` claude-code emits in its own stream output, and re-queues the batch automatically once that reset passes. No configuration is required.

The reset comes from the stream scrutineer already parses, so there is no extra API call, credential, or scope to manage. A batch is bound to the latest exhausted window: if concurrent scans hit different windows (five-hour vs seven-day), the whole batch waits for the furthest reset rather than waking early and re-pausing. Permanent account problems (access disabled or revoked) and API-key accounts report no reset, so they pause for manual resume instead of looping.

The `/usage` page shows the latest per-window status (window, status, and reset) captured from recent scans.

Screenshot 2026-07-02 at 10 08 35

### DIFF
--- a/internal/worker/claude.go
+++ b/internal/worker/claude.go
@@ -153,13 +153,11 @@ func (l LocalClaude) RunSkill(ctx context.Context, sj SkillJob, emit func(Event)
 
 	emit(Event{Kind: KindText, Text: "$ claude -p <skill:" + sj.Name + ">"})
 	accountErrText := ""
-	var lastRateLimit *RateLimitInfo
+	var rateLimitReset *RateLimitInfo
 	wrappedEmit := func(e Event) {
 		accountErrText = preferAccountErrText(accountErrText, claudeAccountErrorText(e.Text))
-		// Keep the latest reported window; account-pause extension reconciles
-		// concurrent scans to the furthest reset.
 		if e.Kind == KindRateLimit && e.RateLimit != nil {
-			lastRateLimit = e.RateLimit
+			rateLimitReset = preferRateLimitReset(rateLimitReset, e.RateLimit)
 		}
 		emit(e)
 	}
@@ -191,7 +189,7 @@ func (l LocalClaude) RunSkill(ctx context.Context, sj SkillJob, emit func(Event)
 			return res, &MaxTurnsReachedError{}
 		}
 		if accountErrText != "" {
-			return res, &ClaudeAccountError{Detail: accountErrText, ResetAt: resumableReset(accountErrText, lastRateLimit)}
+			return res, &ClaudeAccountError{Detail: accountErrText, ResetAt: resumableReset(accountErrText, rateLimitReset)}
 		}
 		return res, fmt.Errorf("claude exited: %w", waitErr)
 	}
--- a/internal/worker/claude.go
+++ b/internal/worker/claude.go
@@ -153,9 +153,13 @@ func (l LocalClaude) RunSkill(ctx context.Context, sj SkillJob, emit func(Event)
 
 	emit(Event{Kind: KindText, Text: "$ claude -p <skill:" + sj.Name + ">"})
 	accountErrText := ""
+	var lastRateLimit *RateLimitInfo
 	wrappedEmit := func(e Event) {
-		if accountErrText == "" {
-			accountErrText = claudeAccountErrorText(e.Text)
+		accountErrText = preferAccountErrText(accountErrText, claudeAccountErrorText(e.Text))
+		// Keep the latest reported window; account-pause extension reconciles
+		// concurrent scans to the furthest reset.
+		if e.Kind == KindRateLimit && e.RateLimit != nil {
+			lastRateLimit = e.RateLimit
 		}
 		emit(e)
 	}
@@ -187,7 +191,7 @@ func (l LocalClaude) RunSkill(ctx context.Context, sj SkillJob, emit func(Event)
 			return res, &MaxTurnsReachedError{}
 		}
 		if accountErrText != "" {
-			return res, &ClaudeAccountError{Detail: accountErrText}
+			return res, &ClaudeAccountError{Detail: accountErrText, ResetAt: resumableReset(accountErrText, lastRateLimit)}
 		}
 		return res, fmt.Errorf("claude exited: %w", waitErr)
 	}
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -503,6 +503,10 @@ func (w *Worker) finalizeScan(ctx context.Context, scan *db.Scan, report string,
 				w.scheduleAccountResumeAt(*effective)
 			}
 		}
+		// resolveAccountReset emitted after the Save above; persist that log tail.
+		if logErr := w.DB.Model(&db.Scan{}).Where("id = ?", scan.ID).Update("log", scan.Log).Error; logErr != nil {
+			w.Log.Warn("account-pause log update failed", "scan", scan.ID, "err", logErr)
+		}
 	}
 	// The clone cache may have grown (clone/fetch/unshallow); refresh the
 	// cached size so the repo list reads it from the row instead of walking
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -492,7 +492,7 @@ func (w *Worker) finalizeScan(ctx context.Context, scan *db.Scan, report string,
 	}
 	w.maybeFireScanFinalized(scan, err)
 	if accErr, isAccountErr := errors.AsType[*ClaudeAccountError](err); isAccountErr && scan.Status == db.ScanPaused {
-		w.pauseQueuedOnAccountError(scan.ID, nil)
+		w.pauseQueuedOnAccountError(scan.ID)
 		if rawReset := w.resolveAccountReset(accErr, emit); rawReset != nil {
 			effective, applyErr := w.applyAccountPauseReset(scan.ID, scan.Error, rawReset)
 			if applyErr != nil {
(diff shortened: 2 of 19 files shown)


internal/worker/worker.go
```diff
@@ -399,8 +491,18 @@ func (w *Worker) finalizeScan(ctx context.Context, scan *db.Scan, report string,
 		return saveErr
 	}
 	w.maybeFireScanFinalized(scan, err)
-	if _, isAccountErr := errors.AsType[*ClaudeAccountError](err); isAccountErr && scan.Status == db.ScanPaused {
-		w.pauseQueuedOnAccountError(scan.ID)
+	if accErr, isAccountErr := errors.AsType[*ClaudeAccountError](err); isAccountErr && scan.Status == db.ScanPaused {
+		w.pauseQueuedOnAccountError(scan.ID, nil)
+		if rawReset := w.resolveAccountReset(accErr, emit); rawReset != nil {
```
andrew, line 496: `resolveAccountReset`'s `emit` calls ("rate limit reset detected; auto-resume after …" / "no rate-limit reset reported…") run *after* `w.DB.Save(scan)` at line 490, and `scanEmitter` only flushes `scan.Log` to the DB when its 2s interval has elapsed — nothing after `finalizeScan` returns re-saves the row (`wrap` just returns the result). So the diagnostic text is appended to the in-memory `scan.Log` but usually not persisted.

`paused_until` and `error` *are* persisted (via `applyAccountPauseReset`'s own `Updates`), so the banner and auto-resume work; only the log-tail explanation is at risk.

Cheapest fix is a single `w.DB.Model(scan).Update("log", scan.Log)` after this block (covers all three emit branches), or move the emit calls into `finishScan` so they land before the Save.

andrew (changes requested):
One finding on log persistence; the auto-resume itself is solid (forward-only extension, far-future clamp, restart pickup all covered by the tests).

internal/worker/worker.go
p-linnane: Addressed in bc9616a.

internal/worker/worker.go
```diff
@@ -517,3 +684,202 @@ func (w *Worker) pauseQueuedOnAccountError(triggerID uint) {
...
+	w.scheduleAccountResumeAt(*scan.PausedUntil)
+}
+
+func (w *Worker) resumeAccountPaused(ctx context.Context) (int, error) {
+	if w.Queue == nil {
+		return 0, errors.New("queue not configured")
+	}
+	var scans []db.Scan
+	if err := w.DB.Select("id", "kind", "finding_id", "error", "paused_until").
+		Where("status = ? AND error LIKE ? AND paused_until IS NOT NULL AND paused_until <= ?",
```
connorshea, line 845: **Auto-resume never fires (and the worker hot-loops) on any behind-UTC host.**

`PausedUntil` is stored in UTC, but `w.now()` here (and in the `maxAcceptable` query at line ~700) is `time.Now()` in the host's **local** zone. glebarez/sqlite stores/compares `time.Time` as RFC3339 text, so `paused_until <= ?` is a lexicographic cross-zone compare — and it's wrong.

Reproduced against the real driver: with a reset stored at `2026-07-01T12:00:00Z` and `now` = 12:01 UTC expressed as `05:01:00-07:00` (Pacific), the query returns **0** due rows (want 1); with `now.UTC()` it returns 1.

Consequence on any behind-UTC deployment: when a reset elapses the timer fires, matches nothing, `scheduleNextAccountResume` re-selects the same past-due row and reschedules with `delay` clamped to 0 → the `AfterFunc` refires immediately in a tight loop and the batch is never resumed. The tests pass only because they inject a UTC `Now`.

Fix: bind `w.now().UTC()` in these `paused_until` comparisons (or store/compare `paused_until` in a consistent zone).

internal/worker/claude.go
```diff
@@ -187,7 +191,7 @@ func (l LocalClaude) RunSkill(ctx context.Context, sj SkillJob, emit func(Event)
 			return res, &MaxTurnsReachedError{}
 		}
 		if accountErrText != "" {
-			return res, &ClaudeAccountError{Detail: accountErrText}
+			return res, &ClaudeAccountError{Detail: accountErrText, ResetAt: resumableReset(accountErrText, lastRateLimit)}
```
connorshea: **The captured reset is the *last* `rate_limit_event` seen, regardless of which window is actually rejected.**

`wrappedEmit` keeps `lastRateLimit` unconditionally, and `resumableReset` calls `rl.ResetTime()` without consulting `Status` (or `overageStatus`) or `Type`. claude-code emits per-window events, including routine `status:"allowed"` heartbeats — the test fixture itself shows `status:"allowed", overageStatus:"rejected"`.

So if an `allowed` five-hour event streams *after* the seven-day window that is the real wall, the batch gets pinned to the wrong reset: it resumes days late, or resumes too early and immediately re-hits the wall and re-pauses (there is no backoff/attempt cap, so this can thrash). Same defect is mirrored in `container.go`.

Fix: select the reset from the *rejected* window (or the furthest reset among rejected windows), not simply the last event.

internal/worker/worker.go
```diff
@@ -473,41 +599,82 @@ func finishErroredScan(scan *db.Scan, report string, err error, emit func(Event)
...
+	detail := err.Error()
+	if len(detail) > maxAutoResumeErrorBytes {
+		detail = detail[:maxAutoResumeErrorBytes] + "..."
+	}
+	return msg + autoResumeFailurePrefix + detail
+}
+
+// pauseQueuedOnAccountError pauses every queued scan behind an account-level
+// Claude failure; queued jobs later drop themselves when wrap sees the DB state.
+func (w *Worker) pauseQueuedOnAccountError(triggerID uint, resetAt *time.Time) {
```
connorshea: **The new `resetAt` parameter is dead in production.**

The only caller (`finalizeScan`) passes `nil`, and no test exercises the non-nil path. `applyAccountPauseReset` then immediately re-writes these same queued rows with the effective reset, so every queued scan is `UPDATE`d twice per account error, and the `accountPauseReason(resetAt)` / `"paused_until": resetAt` branches inside this function are untested code that looks live.

Suggest dropping the `resetAt` parameter (and letting `applyAccountPauseReset` own the reset), or dropping the redundant second write.

connorshea:
Ran a Claude review on it after reviewing the code myself, and posted the 3 findings that seemed worth resolving.

p-linnane:
Thanks @connorshea. All items have been actioned in 915f6ba.

connorshea:
## Review — auto-resume after a Claude rate-limit reset

Nicely self-tested PR. One blocking correctness bug in the reset-selection semantics, plus a few lower-severity items. The concurrency/timer machinery and the `held`→`paused` wording change both hold up under scrutiny.

---

### 🔴 Blocking — `RateLimitInfo.Rejected()` treats a healthy window as rate-limited

`Rejected()` returns true when `OverageStatus == "rejected"` even if `Status == "allowed"`. Combined with `preferRateLimitReset` keeping the **furthest** rejected reset, a scan that only hit the 5-hour wall gets pinned to a ~7-day reset.

I validated this against real claude-code output (not just the fixture in this PR):

- `{status:"allowed", overageStatus:"rejected", isUsingOverage:false}` is the **normal, non-blocking steady state** for any Max/Pro account that hasn't enabled extra-usage billing — the exact JSON appears in [claude-agent-sdk#599](https://github.com/anthropics/claude-agent-sdk-python/issues/599). `overageStatus:"rejected"` just means "pay-as-you-go overage isn't available."
- A genuine denial carries `status:"rejected"` ([claude-code#57096](https://github.com/anthropics/claude-code/issues/57096): `{status:"rejected", overageStatus:"rejected", overageDisabledReason:"out_of_credits"}`). Blocked ⇔ `status=="rejected"`; overage alone never blocks.

**Failure scenario:** a run hits the 5-hour limit (`{status:"rejected", five_hour, resets +15m}`) and also emits the routine `{status:"allowed", overageStatus:"rejected", seven_day, resets +7d}`. Both pass `Rejected()`; +7d is furthest, so `resumableReset` returns +7d and `PausedUntil` becomes +7d. The 8-day cap (`defaultMaxRateLimitAutoResumeAge`) waves it through (7d < 8d). The batch waits ~7 days instead of ~15 minutes — the opposite of the feature's intent. `TestPreferRateLimitReset` currently asserts this (buggy) behavior, which is why it slipped through.

**Caveat on blast radius:** the 7-day harm requires a `seven_day` event in the same run as the 5-hour rejection. I couldn't confirm whether current claude-code emits per-bucket events (five_hour + seven_day) per run or only the single "representative" window ([SDK#50518](https://github.com/anthropics/claude-code/issues/50518) requests per-bucket exposure, hinting single-window today). But this PR's `preferRateLimitReset` "furthest wins" design *assumes* multi-window emission — under its own premise the bug is live — and the misclassification is a latent correctness bug regardless. Fix is small and low-risk, so it should land either way.

**Fix:**

1. Add the discriminator field to `RateLimitInfo` (`internal/worker/stream.go`):
   ```go
   IsUsingOverage bool `json:"isUsingOverage"`
   ```
2. Only count overage when the account is actually drawing on it:
   ```go
   func (r *RateLimitInfo) Rejected() bool {
       if r == nil {
           return false
       }
       if strings.EqualFold(r.Status, "rejected") {
           return true
       }
       // overageStatus:"rejected" is the steady-state normal value for accounts
       // without extra-usage billing enabled; it only blocks when the account
       // is currently drawing on overage.
       return r.IsUsingOverage && strings.EqualFold(r.OverageStatus, "rejected")
   }
   ```
   (`return strings.EqualFold(r.Status, "rejected")` alone is also correct — issue #57096 shows overage-routed denials still surface as `status:"rejected"` — but the `isUsingOverage`-gated form is a safe belt-and-suspenders.)
3. Tests: flip `TestPreferRateLimitReset` so the `{status:"allowed", overageStatus:"rejected", seven_day}` window is **ignored** rather than preferred over a genuinely-rejected `five_hour`; add a regression test that an informational `seven_day +7d` event does not override a real `five_hour +15m` rejection.

Nothing downstream changes — `preferRateLimitReset`/`resumableReset`/`resolveAccountReset` already gate on `Rejected()`, and this also makes the `/usage` panel's `status` column accurate (it currently shows raw `status`, so an overage-only pause would render as "allowed").

<sub>🤖 automated review via `/review`</sub>

p-linnane:
Thanks @connorshea, confirmed and fixed in e3cac67.

`Rejected()` now counts a window as rate-limited only when `status == "rejected"`, or when the account is actually drawing on overage (`isUsingOverage && overageStatus == "rejected"`). The informational `{status:"allowed", isUsingOverage:false, overageStatus:"rejected"}` steady state no longer registers as a rejection, so `preferRateLimitReset` can't pin the batch to a far `seven_day` reset when only the `five_hour` wall was hit.

Verified against the shapes you cited:
- anthropics/claude-agent-sdk-python#599: `{status:"allowed", isUsingOverage:false, overageStatus:"rejected"}` is informational (that issue is the SDK crashing on this non-blocking event).
- anthropics/claude-code#57096: a real denial carries `status:"rejected"` (apiErrorStatus 429), which the status-first check catches regardless of `isUsingOverage`.

Tests: `TestPreferRateLimitReset` now asserts the allowed/overage-unavailable `seven_day` window is ignored in favor of a genuine `five_hour` rejection, plus an active-overage case (`isUsingOverage:true`). `TestResumableReset` and the stream-parse test cover the new field. The /usage `status` column is accurate as a result.

andrew (approved):
Log-persistence fix and Connor's three rounds all check out; pushed one small follow-up (1822174) so an enqueue failure doesn't leave a stale `Auto-resume after <past>` timestamp alongside the failure text.

PR #552, opened 2026-07-02
Gate exposure and VEX exports on dependents

abhinavgautam01:
Part of #22

## Summary

This finishes another application-vs-library feature gate by hiding dependent exposure and VEX bundle output when a repository has no recorded dependents.

## Changes

- Hide the dependent exposure section on finding pages unless dependents exist.
- Keep exposure controls visible for findings whose repository has dependents.
- Stop advertising CSAF VEX in the disclosure bundle tooltip for no-dependent repos.
- Omit `csaf.json` from disclosure bundles when there are no dependents.
- Add tests for both dependent and no-dependent finding/bundle paths.

## Testing

- `git diff --check`
- `go test ./internal/web -run 'TestFindingShow_(hidesExposureForZizmorFindings|hidesExposureWhenNoDependents|rendersExposureWhenDependentsExist)|TestFindingBundle_(containsManifestAndExports|omitsCSAFWhenNoDependents|skipsPatchWhenAbsent)'`
- `go test ./internal/web`
- `go test ./...`

### DIFF
--- a/internal/web/finding_bundle.go
+++ b/internal/web/finding_bundle.go
@@ -114,6 +114,10 @@ func (s *Server) bundleEntries(f *db.Finding, repo *db.Repository) ([]bundleEntr
 		return nil, fmt.Errorf("load finding dependents: %w", err)
 	}
 	deps := loadFindingDependents(s, fdRows)
+	var dependentCount int64
+	if err := s.DB.Model(&db.Dependent{}).Where("repository_id = ?", f.RepositoryID).Count(&dependentCount).Error; err != nil {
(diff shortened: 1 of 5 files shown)


andrew (approved):
Ran the new tests locally. The gate uses the repo-level Dependent count (matching findingShow's hasDependents), not per-finding exposure rows, so it reads as intended. Direct /findings/{id}/csaf.json is not gated, which is consistent with the PR scope (stop advertising + omit from bundle); can follow up if you want that to 404 too.

PR #553, opened 2026-07-03
Gate CSAF export on dependents

abhinavgautam01:
Part of #22, Follow-up according to  https://github.com/alpha-omega-security/scrutineer/pull/552#pullrequestreview-4620253695

## Summary

This applies the same dependent-aware gate to the direct CSAF endpoint that #552 added for the finding page and disclosure bundle. Repositories without recorded dependents no longer expose a VEX-shaped CSAF export.

## Changes

- Return `404` from `/findings/{id}/csaf.json` when the finding repository has no recorded dependents.
- Use the repo-level `db.Dependent` count, matching the finding page and bundle behavior.
- Keep CSAF export behavior unchanged for repositories with dependents.
- Add endpoint coverage for the no-dependent case.
- Update existing CSAF tests to seed dependents where CSAF generation is under test.

## Testing

- `git diff --check`
- `go test ./internal/web -run 'TestFindingCSAF|TestCSAF|TestFindingBundle|TestFindingShow_(hidesExposureWhenNoDependents|rendersExposureWhenDependentsExist)'`
- `go test ./internal/web`
- `go test ./...`

### DIFF
--- a/internal/web/finding_csaf.go
+++ b/internal/web/finding_csaf.go
@@ -91,6 +91,12 @@ func (s *Server) findingCSAF(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "finding is a duplicate; export not available", http.StatusGone)
 		return
 	}
+	var dependentCount int64
+	s.DB.Model(&db.Dependent{}).Where("repository_id = ?", f.RepositoryID).Count(&dependentCount)
+	if dependentCount == 0 {
+		http.Error(w, "CSAF VEX export is unavailable because this repository has no recorded dependents", http.StatusNotFound)
+		return
+	}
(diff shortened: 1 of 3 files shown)


andrew (approved):
Does exactly what the #552 note left open. Gate matches the query shape at `server.go:1577` and `finding_bundle.go:118`; `TestFindingCSAF_noDependentsReturns404` covers it and `seedCSAFFinding` seeds a dependent so existing callers pass through.

`seedCSAFFindingWithoutDependents` is a bit heavy for one caller (`seedCSAFFinding` + a `Delete(&db.Dependent{})` would have done), but not worth a round-trip.

PR #554, opened 2026-07-03
feat: ruby ext durability and allow defining fallback profiles

dkw-oss:
- Cover some more entrypoints in native extensions
- Make profile fallback more deterministic
- Present a warning about image freshness

### DIFF
--- a/docker/profiles/ruby-ext/Dockerfile
+++ b/docker/profiles/ruby-ext/Dockerfile
@@ -105,9 +105,13 @@ ENV DEBIAN_FRONTEND=noninteractive \
 # ffi, gmp, readline, ncurses, gdbm). gdb/strace/valgrind for triage. The apt
 # `ruby` doubles as the BASERUBY the from-source build needs and as the stock
 # interpreter for fast reads / valgrind; ruby-dev lets it build a gem in-place
-# for a valgrind run. wget fetches rustup-init.
+# for a valgrind run. wget fetches rustup-init. ragel/flex regenerate a gem's C
+# scanner/lexer from a committed *.rl / *.l (and bison its *.y parser) when the
+# generated *.c is not checked in — many such gems ship only the generator
+# source and produce the .c at gem-build time, so the cloned repo can't compile
+# without them.
 RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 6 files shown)


andrew (changes requested):
The `FallbackProfile` mechanism and the staleness warning both look right, and the ruby-ext PROFILE.md addition is a good steer.

One thing before merge: the degrade loop in `resolveProfile` has no behavioral test. `TestBuiltinProfiles_registrySanity` guards the registry data (no unknown or self-referencing names) but nothing runs the `for err != nil && p.FallbackProfile != ""` path. `TestResolveProfile_SubPath` in `internal/worker/container_test.go` shows the shape: drive `resolveProfile` against a `ProfilesDir` where the first profile's Dockerfile is missing (so `EnsureImage` returns the read error) and the fallback's is present, then assert the emitted `falling back to` line and that the returned profile name is the fallback.

While you're in the sanity test, a cycle check on `FallbackProfile` (walk the chain and fail if a name repeats) would close the only way this loop can spin, though `BaseProfile` has the same gap so not a blocker.

dkw-oss:
> The `FallbackProfile` mechanism and the staleness warning both look right, and the ruby-ext PROFILE.md addition is a good steer.
> 
> One thing before merge: the degrade loop in `resolveProfile` has no behavioral test. `TestBuiltinProfiles_registrySanity` guards the registry data (no unknown or self-referencing names) but nothing runs the `for err != nil && p.FallbackProfile != ""` path. `TestResolveProfile_SubPath` in `internal/worker/container_test.go` shows the shape: drive `resolveProfile` against a `ProfilesDir` where the first profile's Dockerfile is missing (so `EnsureImage` returns the read error) and the fallback's is present, then assert the emitted `falling back to` line and that the returned profile name is the fallback.
> 
> While you're in the sanity test, a cycle check on `FallbackProfile` (walk the chain and fail if a name repeats) would close the only way this loop can spin, though `BaseProfile` has the same gap so not a blocker.

Cheers, actioned in 6356373

andrew (approved):
Thanks — `TestResolveProfile_DegradesToFallback` covers the degrade loop exactly as discussed, and `assertProfileChainAcyclic` closes the cycle gap on both chains. The runtime `tried` guard is a nice extra.

PR #555, opened 2026-07-03
fix(worker): bind egress sidecar to its internal leg only; skip host proxy under rootless hardened

### LINKED ISSUE #525: egress sidecar: bind to internal-network interface only; skip host proxy under rootless hardened
Follow-up to #524, two residuals noted during review.

The sidecar listens on `:3128` (all interfaces), so once `network connect` attaches the egress leg, other containers on the rootless user's default `podman` bridge can reach the listener. They get 407 without the token so it's a probe surface rather than a bypass, and the threat model is explicit that the token isn't the isolation boundary. Binding only on the internal-network interface would remove the surface entirely. The complication is ordering: `proxySidecarRunArgs` starts the sidecar on `--network podman` and `startProxySidecar` connects the internal network afterwards, so the internal IP isn't known at bind time. Starting on the internal network first and connecting `podman` second would let the sidecar discover its own IP before `ListenAndServe`, at the cost of `WaitHostAPIReachable` needing to wait for the egress leg to attach.

Separately, `setupRunner` still calls `StartEgressProxy` under rootless `--hardened`, so the in-process host proxy starts and listens on a host port even though every scan routes through the sidecar and `d.ProxyURL` is shadowed by `hnet.proxyEndpoint` in `buildRunArgs`. Harmless but unused; could be skipped when `egress.GatewayIP != ""`.

Neither blocks #524.

alexandre-daubois:
Fix #525

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -670,23 +670,32 @@ func setupRunner(f *flags, cfg *config.Config, log *slog.Logger) (worker.SkillRu
 		allow = append(allow, apiHost)
 	}
 	token := worker.NewProxyToken()
-	port, err := worker.StartEgressProxy(&worker.EgressProxy{
-		Allow:    allow,
-		Token:    token,
-		APIPort:  addrPort(f.addr),
-		APIHosts: []string{apiHost},
-		Log:      log,
-	})
-	if err != nil {
-		return nil, "", fmt.Errorf("start egress proxy: %w", err)
-	}
-	// Rootless --hardened runs the egress proxy as a sidecar reusing the host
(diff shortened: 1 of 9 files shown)


andrew (approved):
Addresses both #525 residuals cleanly. The internal-first ordering plus the `first-iface` keyword fails closed at every step (`firstIfaceIPv4` errors rather than falling through to the egress leg, `resolveListen` propagates resolver failure), and the integration probe from `--network podman` to the egress-leg IP expecting `000` is the direct proof. Traced `ProxyURL == ""` under sidecar mode through `buildRunArgs`, `verifyHostProxyReachable`, and the Apple branch — all safe.

PR #556, opened 2026-07-03
feat(skills): add advisory-deep-dive skill auditing past advisories for bypasses and incomplete fixes

alexandre-daubois:
cc @connorshea


connorshea:
Hm, do these end up in the Scanners bucket because we haven't added this skill to the list of skills that output real "Findings"?

alexandre-daubois:
Ah yes indeed, it only showed up in "Scanners". It's fixed by the last commit!

andrew (changes requested):
The skill and its wiring look right — schema matches the finding shape `internal/worker/findings.go` already ingests (including `references`), `scrutineer.requires: [advisories]` lines up with security-deep-dive, and the classification change is threaded through all four sync points with `TestFindingClassification_advisoryDeepDiveIsCurated` driving both the GORM subquery and raw-SQL paths.

One gap: `TestDeepDiveSkillNameSafeForSplicing` at `internal/web/server_test.go:1415` still iterates only `deepDiveSkillName` and `vulnScanSkillName`. `advisoryDeepDiveSkillName` is now spliced into `findingsBucketSkillSQL` the same way, so it should join that loop — the tripwire exists to cover every spliced constant.

Two cosmetic while-you're-there bits: the comment at `finding_classification_test.go:57` says "(two bound skill names)" but `scannerScanFilter` uses spliced literals, not binds (and it's three now) — drop the parenthetical; and the reflow at `server.go:1151` left `An // import is a` orphaned mid-sentence.

alexandre-daubois:
Here you go

andrew (approved):

Issue #126, opened 2026-04-30
Audit and fix N+1 queries for large datasets


andrew:
With a few thousand repos and findings the UI gets noticeably slow. Likely culprits are N+1 queries on list pages (repos, findings, scans) where associations are loaded per-row instead of preloaded.

Work needed:

- Profile query counts on list endpoints with a populated database
- Add `Preload` / `Joins` where GORM is issuing per-row lookups
- Consider pagination or count caching for badge/aggregate columns
- Add a benchmark test so regressions are caught early

andrew:
Related:
- #134 — `/scans` CASE sort fixed in #138
- #135 — `/repositories/{id}` `SELECT * FROM packages` fixed in #138
- #133 — per-request query count/timing in HTTP logs (draft)

andrew:
Reopening: #358 added the regression coverage and fixed the /scans badge counts, but two per-row costs remain outside what the query-count tests can see — `worker.RepoDiskUsage` filesystem stats per repo row on the repo list, and the correlated findings-sort subquery. Neither is a DB N+1, so closing criteria here should name them explicitly.

PR #569, opened 2026-07-05
Add expected findings benchmark support

### LINKED ISSUE #568: Expected findings for benchmarking scan skills
We'd like to point the model-backed scan skills (`vuln-scan`, `security-deep-dive`) at a repository where we already know what the vulnerabilities are and see whether the skill finds them.

Proposal is an `expected_findings` table, a match against the existing `findings` table, and a rollup view. Same idea as the `threat_model.json` override: operator-supplied known input the skill is checked against.

Schema:

```
expected_findings
  id
  repository_id   FK repositories
  file            path relative to repo root, no line number
  cwe             CWE-NNN
  cve             optional, for reference
  note            optional, free text
```

A scan finding matches an expected row when the path part of its `location` equals `file` and its `cwe` is equal. Line numbers are ignored because they drift between versions and the CWE disambiguates when one file has more than one sink.

On the repo findings page, when the repo has expected rows, badge each expected row as matched or missed and each scan finding as expected or unexpected. Comparison is per scan, so re-running the skill produces a fresh set of matches against the same expected rows and the scans list is the history.

Rollup: a `/benchmark` page listing repos that have expected rows, with per-repo and total recall (matched / expected) and precision (matched / scan findings at Medium and above) for the most recent completed scan. Columns and filters for skill, model, and harness version so a change in score can be attributed. The scans table already records model; if it doesn't already capture harness version and skill `scrutineer.version`, add those columns as part of this. F1 if wanted, though recall is the number people will actually watch.

API: `POST /repositories/{id}/expected` to add a row, `GET` to list, `DELETE /repositories/{id}/expected/{id}` to remove. Bulk add on the repo create form would be handy but isn't required to start.

We'd populate it with real projects at known-vulnerable versions and have a starting list with documented sink locations to contribute once the table exists.

abhinavgautam01:
Fixes #568

## Summary

Adds expected findings support for benchmarking model-backed scan skills against known vulnerable repositories.

This introduces repository-scoped expected findings, matches them against scan findings by file path + CWE while ignoring line numbers and adds a benchmark rollup page for recall/precision tracking.

## Changes

- Add `expected_findings` storage via `db.ExpectedFinding`
- Add read-only API support to list expected findings
- Add operator UI routes to add and delete expected findings
- Add a Benchmark tab on repository pages
- Badge scan findings as expected or unexpected when benchmark rows exist
- Add `/benchmark` rollup with recall, precision, F1 and filters for skill/model/harness
- Snapshot `skill_schema_version` on scans for benchmark attribution
- Update OpenAPI docs
- Add focused tests for read-only API access, UI writes, matching, repo UI badges, benchmark rollup and scan version snapshots

## API scope

The scan-token API exposes expected findings as read-only. Creating and deleting expected findings is intentionally limited to the operator UI routes so running skills cannot mutate benchmark ground truth and inflate their own recall or precision.

## Tests

- `go test ./internal/web -run 'TestExpected|TestBenchmark|TestBuildExpected|TestRepoShowBenchmark|TestRepoShow'`
- `go test ./...`
- `go vet ./...`
- `git diff --check`

Main fixes are:

- Change “Add API endpoints to list, add and delete expected findings” to “read-only API support”.
- Add the API scope section explaining why writes are UI/operator-only.
- Update tests to match what you actually ran.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -106,8 +106,9 @@ type Repository struct {
 	CreatedAt time.Time
 	UpdatedAt time.Time
 
-	Scans       []Scan       `gorm:"constraint:OnDelete:CASCADE"`
-	Maintainers []Maintainer `gorm:"many2many:repository_maintainers"`
+	Scans            []Scan            `gorm:"constraint:OnDelete:CASCADE"`
+	ExpectedFindings []ExpectedFinding `gorm:"constraint:OnDelete:CASCADE"`
+	Maintainers      []Maintainer      `gorm:"many2many:repository_maintainers"`
 }
 
 // IsLocal reports whether this Repository points at a directory on disk
@@ -160,9 +161,10 @@ type Scan struct {
 	// the skill is deleted. APIToken is a random bearer generated per-run
 	// so skills can call back into scrutineer's HTTP API from inside the
 	// workspace; it is cleared when the scan reaches a terminal state.
-	SkillID      *uint `gorm:"index"`
-	SkillVersion int
-	SkillName    string `gorm:"index"`
+	SkillID            *uint `gorm:"index"`
+	SkillVersion       int
+	SkillSchemaVersion int
+	SkillName          string `gorm:"index"`
 	// FindingID is set when a scan is finding-scoped (verify, patch,
 	// disclose). Skills read it from context.json to know which finding
 	// they are acting on.
@@ -441,6 +443,24 @@ type Dependency struct {
 	ManifestPath          string
 	ManifestKind          string
 	CreatedAt             time.Time
+}
+
+// ExpectedFinding is an operator-supplied benchmark target for a repository.
+// It records a known vulnerable file/CWE pair that model-backed scans should
+// rediscover. Matching intentionally ignores line numbers because they drift
+// between versions.
+type ExpectedFinding struct {
+	ID           uint `gorm:"primarykey"`
+	RepositoryID uint `gorm:"index;not null;uniqueIndex:idx_expected_repo_file_cwe,priority:1"`
+	Repository   Repository
+
+	File string `gorm:"not null;uniqueIndex:idx_expected_repo_file_cwe,priority:2"`
+	CWE  string `gorm:"not null;uniqueIndex:idx_expected_repo_file_cwe,priority:3"`
+	CVE  string
+	Note string `gorm:"type:text"`
+
+	CreatedAt time.Time
+	UpdatedAt time.Time
 }
 
 // FindingResolution says how a finding got resolved. Set by the analyst
@@ -1083,7 +1103,7 @@ func Open(dsn string) (*gorm.DB, error) {
 		&Repository{}, &Scan{},
 		&Finding{}, &FindingLabel{}, &FindingNote{},
 		&FindingCommunication{}, &FindingReference{}, &FindingHistory{}, &FindingReview{},
-		&Dependency{}, &Package{}, &Dependent{}, &FindingDependent{}, &Advisory{},
+		&Dependency{}, &ExpectedFinding{}, &Package{}, &Dependent{}, &FindingDependent{}, &Advisory{},
 		&Maintainer{}, &Skill{}, &Subproject{},
 		&SBOMUpload{}, &SBOMPackage{}, &CNA{}, &Setting{},
 	); err != nil {
--- a/internal/web/api.go
+++ b/internal/web/api.go
@@ -94,6 +94,9 @@ func (s *Server) apiHandler() http.Handler {
 	mux.HandleFunc("GET /repositories/{id}/advisories", s.apiListAdvisories)
 	mux.HandleFunc("GET /repositories/{id}/dependents", s.apiListDependents)
 	mux.HandleFunc("GET /repositories/{id}/ecosystems/{source}/raw", s.apiGetEcosystemsRaw)
+	mux.HandleFunc("GET /repositories/{id}/expected", s.apiListExpectedFindings)
+	mux.HandleFunc("POST /repositories/{id}/expected", s.apiAddExpectedFinding)
+	mux.HandleFunc("DELETE /repositories/{id}/expected/{expected_id}", s.apiDeleteExpectedFinding)
 	mux.HandleFunc("GET /repositories/{id}/dependencies", s.apiListDependencies)
 	mux.HandleFunc("GET /repositories/{id}/findings", s.apiListFindings)
 	mux.HandleFunc("POST /repositories/{id}/findings", s.apiStreamFinding)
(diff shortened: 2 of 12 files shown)


internal/web/api.go
```diff
@@ -94,6 +94,9 @@ func (s *Server) apiHandler() http.Handler {
 	mux.HandleFunc("GET /repositories/{id}/advisories", s.apiListAdvisories)
 	mux.HandleFunc("GET /repositories/{id}/dependents", s.apiListDependents)
 	mux.HandleFunc("GET /repositories/{id}/ecosystems/{source}/raw", s.apiGetEcosystemsRaw)
+	mux.HandleFunc("GET /repositories/{id}/expected", s.apiListExpectedFindings)
+	mux.HandleFunc("POST /repositories/{id}/expected", s.apiAddExpectedFinding)
+	mux.HandleFunc("DELETE /repositories/{id}/expected/{expected_id}", s.apiDeleteExpectedFinding)
```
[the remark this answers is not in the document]
abhinavgautam01: The scan-token API now exposes expected findings as read-only: only `GET /repositories/{id}/expected` remains registered and documented. The write handlers were removed from the bearer-token API surface, while the existing host/operator form routes still handle create/delete from the UI.

internal/web/expected_findings.go
```diff
@@ -0,0 +1,327 @@
...
+	gdb.Where("repository_id = ?", repoID).Order("file, cwe").Find(&expected)
+	latestScanID := uint(0)
+	if latest != nil {
+		latestScanID = latest.ID
+	}
+	visibleFindings := make([]db.Finding, 0, len(rf.DeepDive)+len(rf.Scanners))
+	visibleFindings = append(visibleFindings, rf.DeepDive...)
+	visibleFindings = append(visibleFindings, rf.Scanners...)
+	return repoExpectedView{
+		Matches:       expectedMatchesForRows(gdb, repoID, latestScanID, expected),
```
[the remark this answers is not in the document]
abhinavgautam01: Repo expected-finding badge state now includes `latest.Findings` in addition to the open tab slices, so closed findings shown in the Latest scan table are classified correctly instead of defaulting to unexpected.

internal/web/expected_findings.go
```diff
@@ -0,0 +1,327 @@
...
+			}
+		}
+	}
+	return out
+}
+
+func findingMatchesExpected(f db.Finding, expected db.ExpectedFinding) bool {
+	if strings.TrimSpace(f.CWE) != strings.TrimSpace(expected.CWE) {
+		return false
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: CWE comparison now normalizes both sides with uppercase + trim and the create path still stores normalized CWE values. Added coverage for a lowercase scan finding matching an uppercase expected row.

internal/web/expected_findings.go
```diff
@@ -0,0 +1,327 @@
...
+	if scanID == 0 || len(expected) == 0 {
+		return out
+	}
+	var findings []db.Finding
+	gdb.Where("repository_id = ? AND scan_id = ?", repoID, scanID).Find(&findings)
+	for _, f := range findings {
+		if db.SeverityAtLeast(f.Severity, "Medium") {
+			out.FindingTotal++
+		}
+		for i := range out.Expected {
```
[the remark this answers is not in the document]
abhinavgautam01: Benchmark matching now skips findings below Medium before both `FindingTotal` and match accounting, so `MatchedTotal` and `FindingTotal` use the same severity set.

internal/web/expected_findings.go
```diff
@@ -0,0 +1,327 @@
...
+		CVE:          strings.TrimSpace(cve),
+		Note:         strings.TrimSpace(note),
+	}
+	if row.File == "" || row.File == "." {
+		return row, fmt.Errorf("file is required")
+	}
+	if row.CWE == "" {
+		return row, fmt.Errorf("cwe is required")
+	}
+	return row, nil
```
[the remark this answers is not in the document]
abhinavgautam01: Expected finding paths are now validated on create: absolute paths and any `..` path segment are rejected, while valid repo-relative paths are still normalized.

internal/web/expected_findings.go
```diff
@@ -0,0 +1,317 @@
...
+	MatchedTotal  int
+	FindingTotal  int
+}
+
+type repoExpectedView struct {
+	Matches       expectedMatchSet
+	FindingStatus map[uint]bool
+}
+
+func loadRepoExpectedView(gdb *gorm.DB, repoID uint, latest *db.Scan, rf repoFindings) repoExpectedView {
```
alexandre-daubois, line 153: `loadRepoExpectedView` uses the deep-dive-only `latest`, so a repo scanned only by `vuln-scan` gets `latestScanID=0` and its Benchmark tab shows every expected row as "missed" (0/N), while the per-finding badges on the same page mark the matching finding "expected" and `/benchmark` counts it as recall 1/1. You should resolve the repo-page comparison against the same latest benchmark-eligible scan `latestBenchmarkScan` selects (deep-dive or vuln-scan), not the deep-dive `latest`.

internal/web/benchmark.go
```diff
@@ -0,0 +1,121 @@
...
+			matches := expectedMatchesForRows(gdb, repo.ID, scan.ID, expected)
+			row.Scan = scan
+			row.Matched = matches.MatchedTotal
+			row.Findings = matches.FindingTotal
+			row.Skill = scan.SkillName
+			row.Model = scan.Model
+			row.HarnessSHA = scan.SkillsRepoSHA
+			row.SkillSchemaVersion = scan.SkillSchemaVersion
+			row.Recall = ratio(row.Matched, row.Expected)
+			row.Precision = ratio(row.Matched, row.Findings)
```
alexandre-daubois: This mixes units: `Matched` is matched *expected rows* but `Findings` is Medium+ *findings*. A single finding matching two same-CWE expected files makes precision 2.0 (renders "200%"), and two findings for one expected row understate it. Use the count of true-positive findings (the Medium+ entries in `FindingStatus`) as the precision numerator, or clamp and document the metric.

internal/db/db.go
```diff
@@ -106,8 +106,9 @@ type Repository struct {
 	CreatedAt time.Time
 	UpdatedAt time.Time
 
-	Scans       []Scan       `gorm:"constraint:OnDelete:CASCADE"`
-	Maintainers []Maintainer `gorm:"many2many:repository_maintainers"`
+	Scans            []Scan            `gorm:"constraint:OnDelete:CASCADE"`
+	ExpectedFindings []ExpectedFinding `gorm:"constraint:OnDelete:CASCADE"`
```
alexandre-daubois, line 110: The new `expected_findings` table and the `scans.skill_schema_version` column need a `docs/database.md` entry in this PR: right now database.md documents `skill_version` but neither of the new additions.

internal/web/expected_findings.go
abhinavgautam01: The repo-page expected-finding comparison now selects the same latest benchmark-eligible scan as `/benchmark` (`security-deep-dive` or `vuln-scan`) instead of only using the deep-dive latest scan. Added a vuln-scan-only repo-page regression test.

internal/web/benchmark.go
abhinavgautam01: Precision now uses true-positive findings as the numerator, while recall still uses matched expected rows. Added coverage for one finding matching two expected rows to ensure precision does not exceed 100%.

internal/db/db.go
abhinavgautam01: added `docs/database.md` documentation for `scans.skill_schema_version` and the new `expected_findings` table, including the unique key and matching semantics.

andrew (changes requested):
Thanks for this, the matching logic and tests look solid. A few things to sort before merge:

**`repo_show.html` has hard tabs mixed into a space-indented file.** Every line you touched (the rt12/rt13 buttons, the badge blocks, the whole rp13 panel, and the `{{if .Advisories}}` line) now starts with a tab followed by the original spaces. Looks like an editor applied Go indentation to the HTML. Please revert those to spaces so the diff only shows actual changes and blame stays clean on the lines that didn't functionally move (the Workbench button block, for instance, is byte-identical apart from the leading tab).

**The `repoShow` refactor dropped three explanatory comments:** the `activeScans drives both the delete-confirm warning...` block, the `Cached on the row, refreshed by the worker after each scan (#126)` note on `DiskBytes`, and the `Count failed scans in the latest-per-skill set...` comment. Please carry those over to the new helpers; they explain why, not what.

**API write path:** #568 asked for `POST`/`DELETE` on `/api/repositories/{id}/expected`, but only `GET` is wired into `apiHandler()`. The test asserting scan tokens can't write is exactly right, but there's no operator write path via the API either — writes are UI-form only. If that's deliberate, fine, just note it in the PR body; otherwise the bulk-populate case in the issue would want it.

Minor: the `case int:` branch in `skillSchemaVersion` is dead since `json.Unmarshal` into `map[string]any` always yields `float64` for numbers.

You'll also want to rebase on main — #571 just merged and both touch `internal/web/server.go`, though in different functions so it should be clean.

abhinavgautam01:
Thanks, fixed

- Rebased this branch on latest `main`.
- Cleaned up the accidental leading tabs in `repo_show.html`; those lines are back to space indentation.
- Restored the dropped explanatory comments in the new repo-show helper flow.
- Removed the dead `case int:` branch from `skillSchemaVersion`.
- Updated the PR body to clarify the expected-findings API scope: scan-token API access is read-only, while add/delete remains operator UI-only so skills cannot mutate benchmark ground truth.

andrew (approved):
Thanks for the quick turnaround, all four points addressed.

PR #570, opened 2026-07-06
feat(findings): add Repo ID column to the findings table

dkw-oss:
feat(findings): add Repo ID column to the findings table

Surface the repository's primary key next to the Repository name on the
/findings page. It's the same value the JSON API emits as repository_id,
so having it visible in the list makes finding-level exports easier to
cross-reference without opening each repo.

Pure template change: the repo record (and its ID) was already in scope
in findings.html via the Repos map, so no handler, model, or serializer
change is needed. The cell reuses the existing "#<id>" muted style used
for the adjacent Scan column and on the repository page.

### DIFF
--- a/internal/web/templates/findings.html
+++ b/internal/web/templates/findings.html
@@ -121,6 +121,7 @@
     <thead><tr>
       <th class="w-24">Severity</th><th>Title</th><th class="w-24">Status</th>
       <th class="w-32">Repository</th>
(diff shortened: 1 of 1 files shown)


andrew (approved):

dkw-oss:
@andrew called it `Repo ID` instead of `Repository ID` as it overflows the table margins otherwise, requiring more UI changes, are you ok with this?

PR #571, opened 2026-07-07
feat: sortable table columns

dkw-oss:
Click a column header on any index table (Repositories, Findings, Packages,
SBOMs, Advisories, Maintainers, Scans, Organizations) to sort; click again to
flip ascending/descending. Reuses the existing server-side `?sort` +
pagination, so sorting survives filtering and paging.

- Direction is folded into the `sort` token — no new query params. The active
  column shows a direction arrow + `aria-sort`; other sortable columns show a
  faded chevron affordance.
- ORDER BY direction is carried as a boolean (`orderByExpr`), so no request
  text ever reaches a query — includes an injection regression test.


dkw-oss:
@andrew all copilot feedback actioned

andrew (approved):

PR #576, opened 2026-07-07
Add bounded retries for external API fetches

### LINKED ISSUE #544: Add bounded retry/backoff for ecosyste.ms and GitHub API fetches
### Description

Scrutineer performs several background or ingestion-time HTTP requests against third-party services, including ecosyste.ms and GitHub. These paths generally call `http.DefaultClient.Do(req)` directly and fail immediately on transient upstream problems.

Examples include:

- `ecosystemsGetWithLink` in `internal/worker/ecosystems.go`
- package lookup helpers such as `FetchPackagesByPURL`
- GitHub organization import requests in `internal/web/org_import.go`
- other idempotent metadata/API prefetches

These requests can fail due to temporary rate limiting or gateway errors such as `429`, `502`, `503` or `504`. A short, bounded retry policy would make repository ingestion and cache hydration more resilient.

### Impact

Temporary upstream failures can leave repository metadata partially hydrated, even when a retry a few seconds later would succeed. This makes scans and UI data less complete than necessary.

### Suggested Fix

Introduce a small shared HTTP helper for idempotent `GET` requests that:

- Retries only safe transient failures such as network errors, `429`, `502`, `503`, and `504`.
- Respects `Retry-After` when present.
- Uses bounded exponential backoff with jitter.
- Honors the caller's context deadline.
- Keeps max attempts low so background jobs do not hang.

Use it for ecosyste.ms prefetches and GitHub API fetches where retrying is safe.

Add tests covering:

- Success after a transient failure.
- No retry on non-retryable `4xx`.
- `Retry-After` handling.
- Context cancellation.

abhinavgautam01:
closes #544
## Summary

Adds a shared retry helper for idempotent external `GET` requests and applies it to metadata/API fetch paths used during repository ingestion and cache hydration.

## Changes

- Add `internal/httpx.DoRetry` with bounded retry/backoff behavior.
- Retry transient network errors and `429`, `502`, `503`, `504`.
- Respect `Retry-After` when upstream sends it.
- Honor caller context cancellation/deadlines.
- Use the helper for:
  - ecosyste.ms prefetches
  - packages.ecosyste.ms PURL lookup
  - CNA sync
  - GitHub organization import

## Testing

- `go test ./internal/httpx ./internal/worker -run 'TestDoRetry|TestRefreshEcosystems|TestFetchDependentsPaginationStopsAtCaps|TestAppendEcosystemsPagesStopsWhenInitialBodyHitsCap|TestSyncCNAs'`
- `go test ./internal/web -run 'TestOrgImport'`
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/internal/httpx/retry.go
+++ b/internal/httpx/retry.go
@@ -1,0 +1,150 @@
+package httpx
+
+import (
+	"context"
+	"fmt"
+	"math/rand/v2"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+const (
+	defaultAttempts  = 3
+	defaultBaseDelay = 200 * time.Millisecond
+	defaultMaxDelay  = 2 * time.Second
+	backoffFactor    = 2
+	jitterDivisor    = 4
+)
+
+// RetryOptions configures DoRetry. Zero values use small bounded defaults.
+type RetryOptions struct {
+	Attempts  int
+	BaseDelay time.Duration
+	MaxDelay  time.Duration
+	Sleep     func(context.Context, time.Duration) error
+}
+
+// DoRetry performs an idempotent HTTP request with a small retry budget for
+// transient upstream failures. It retries network errors and 429/502/503/504
+// responses, respects Retry-After when present, and always honors req.Context().
+func DoRetry(req *http.Request, opts RetryOptions) (*http.Response, error) {
+	if req.Method != http.MethodGet {
+		return nil, fmt.Errorf("retry helper only supports GET, got %s", req.Method)
+	}
+	attempts := defaultedAttempts(opts.Attempts)
+	baseDelay := defaultedDuration(opts.BaseDelay, defaultBaseDelay)
+	maxDelay := defaultedDuration(opts.MaxDelay, defaultMaxDelay)
+	sleep := opts.Sleep
+	if sleep == nil {
+		sleep = sleepContext
+	}
+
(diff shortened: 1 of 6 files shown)


andrew (changes requested):
Looks good overall, one blocking issue:

`retryAfterDelay` doesn't cap the parsed value. GitHub sends `Retry-After` values in the hundreds or thousands of seconds on secondary rate limits, and ecosyste.ms could do similar. With the callers here running under long-lived job contexts, that means a background worker can sit in `sleepContext` for an hour on a single 429. The issue asked for bounded backoff specifically so jobs don't hang.

Please clamp the Retry-After delay to `maxDelay` (or a separate ceiling if you'd rather keep them independent), and add a test asserting a large `Retry-After` gets capped.

Minor, non-blocking: the retryable response body is closed but not drained, so the connection won't go back to the pool. `_, _ = io.Copy(io.Discard, resp.Body)` before the close would fix that.

abhinavgautam01:
Thanks, fixed

- Clamped parsed `Retry-After` delays to the configured `MaxDelay`, so large secondary-rate-limit values cannot park a worker for minutes or hours.
- Added a regression test covering a large `Retry-After` value being capped.
- Also drained retryable response bodies before closing them so the underlying connection can be reused.

andrew (approved):
Thanks, both points addressed.

PR #584, opened 2026-07-08
Harden Codex harness event parsing

### LINKED ISSUE #583: Codex harness follow-ups from #535 review
Three minor items left over from the #535 review, none blocking:

**`parseCodexLine` session arm is field-driven** (`internal/worker/harness_codex.go` around `case ev.SessionID != "" || ev.ThreadID != "":`). It fires before every other arm, so if a future codex release stamps `thread_id`/`session_id` on `turn.completed` or item events (common in JSONL protocols), usage and error payloads collapse to a bare `KindSession` and every scan reports \$0. At the pinned 0.142.5 only `thread.started` carries `thread_id`, so this is defensive. Gate on `ev.Type` (`thread.started`, `session.created`, `init`) or move the arm below the `turn.completed`/item/error arms.

**`"billing details"` in `CodexHarness.AccountErrorText`** is a loose substring. `AccountErrorText` runs against every emitted line, so a codex agent that mentions "billing details" mid-run and then fails for an unrelated reason (timeout, network) pauses the queue as an account error instead of a plain failure. `insufficient_quota` already covers the structured error; consider dropping this phrase or tightening it to the exact fragment OpenAI returns.

**Stale comment at `internal/web/models.go:161`** says "the built-in Anthropic list (which carries no tier tags)"; `ClaudeHarness.DefaultModels()` now tags mid/high/max. Comment-only.

abhinavgautam01:
Fixes #583

## Summary

Applies the remaining Codex harness follow-ups from the #535 review.

This makes Codex JSONL parsing more defensive, reduces false-positive account error detection and updates a stale model-tier comment.

## Changes

- Gate Codex session events on explicit event types: `thread.started`, `session.created` and `init`
- Preserve usage, text and error payloads if future Codex events include `thread_id` or `session_id`
- Remove the loose `"billing details"` account-error substring
- Update the stale model-tier fallback comment in `internal/web/models.go`
- Add regression tests for session parsing and account-error false positives

## Tests

- `go test ./internal/worker -run 'TestCodexHarness'`
- `git diff --check`
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/internal/web/models.go
+++ b/internal/web/models.go
@@ -159,8 +159,8 @@ func builtinModelForTier(tier, fallback string) string {
 			return m.ID
 		}
 	}
-	// Otherwise the substring heuristic covers the built-in Anthropic list
-	// (which carries no tier tags): Mid → first sonnet, Max → latest opus.
+	// Otherwise the substring heuristic keeps older or minimally tagged
+	// model lists usable: Mid → first sonnet, Max → latest opus.
 	// A custom list that neither tags tiers nor matches these needles falls
 	// through to DefaultModel; set the tiers in /settings for that case.
 	switch tier {
--- a/internal/worker/harness_codex.go
+++ b/internal/worker/harness_codex.go
@@ -149,7 +149,7 @@ func parseCodexLine(raw []byte, emit func(Event)) {
 		return
 	}
 	switch {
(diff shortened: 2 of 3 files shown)


andrew (approved):
All three follow-ups from #583 addressed, thanks.

PR #595, opened 2026-07-09
Make finding API patches atomic

### LINKED ISSUE #563: Make multi-field finding patches atomic
### Description

`internal/web/api_finding_writes.go` lets API clients update multiple finding fields in one `PATCH /api/findings/{id}` request. The handler loops over `body.Fields` and calls `db.WriteFindingField` once per field:

```go
for field, value := range body.Fields {
    if err := db.WriteFindingField(s.DB, uint(id), field, value, source, body.By); err != nil {
        writeAPIError(w, http.StatusUnprocessableEntity, err.Error())
        return
    }
}
```

Each `WriteFindingField` call loads the finding and runs its own transaction. If one field succeeds and a later field fails validation, the earlier update remains committed.

### Impact

Multi-field API patches are not atomic. A malformed later field can leave the finding partially updated, which is surprising for clients and can produce partial history rows for a request that ultimately returned an error.

### Suggested Fix

Wrap all field updates from one API patch request in a single database transaction.

A focused first step can reuse `db.WriteFindingField` with the transaction handle so all writes commit or roll back together. A larger follow-up could reduce duplicate SELECTs and batch history inserts, but that should preserve:

- field validation,
- no-op behavior when the value is unchanged,
- `FindingHistory` rows,
- CVSS score synchronization for CVSS vector fields.

### Suggested Tests

- A patch with two valid fields commits both.
- A patch with one valid field and one invalid field commits neither.
- History rows are rolled back when the patch fails.
- CVSS score synchronization still works inside the transaction.

abhinavgautam01:
Fixes #563
## Summary

Wraps multi-field `PATCH /api/findings/{id}` updates in a single database transaction so partial updates cannot be committed when a later field fails validation.

## Changes

- Run all finding field writes from one API PATCH request inside one transaction
- Keep existing `db.WriteFindingField` validation, history logging, and CVSS score synchronization behavior
- Apply fields in deterministic sorted order
- Add regression coverage for rollback behavior
- Add coverage that CVSS score synchronization still works inside the transaction

## Tests

- `go test ./internal/web -run 'TestAPIPatchFinding|TestAPIListFindingHistory|TestSourceFromRequest'`
- `go test ./...`
- `go vet ./...`
- `git diff --check`
- `golangci-lint run`

### DIFF
--- a/internal/web/api_finding_writes.go
+++ b/internal/web/api_finding_writes.go
@@ -3,10 +3,13 @@ package web
 import (
 	"encoding/json"
 	"net/http"
+	"sort"
 	"strconv"
 	"time"
 
 	"scrutineer/internal/db"
+
+	"gorm.io/gorm"
 )
 
 // The handlers below let skills (and the browser UI) mutate a finding:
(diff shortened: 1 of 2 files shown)


andrew (approved):
Reviewed in a separate worktree. No blocking findings; targeted tests, full test suite, go vet, and git diff --check passed.

PR #596, opened 2026-07-09
Use bulk updates for scan status actions

### LINKED ISSUE #575: Use bulk updates for queued scan pause/cancel operations where safe
### Summary

`internal/web/scans.go` currently performs per-row updates for bulk scan status changes, especially in `scansPauseQueued`.

### Problem

`scansPauseQueued` loads every queued scan into memory and updates each row individually. For large queues, this becomes one `SELECT` plus N `UPDATE` statements.

`scansCancelAll` is more nuanced: running scans still need to call `Worker.Cancel(scan.ID)` so the worker can unwind correctly. However, queued scans are not actively running and can be updated in bulk.

### Suggested Fix

Replace `scansPauseQueued` with a single bulk `UPDATE` for all queued scans.

For `scansCancelAll`, keep the worker cancellation path for running scans, but consider handling queued scans with one bulk update scoped to the repository. Running scans should continue to go through `Worker.Cancel` or equivalent worker-aware logic.

### Impact

Reduces database roundtrips from O(N) to O(1) for queued-only bulk status changes, while preserving required worker side effects for running scans.

abhinavgautam01:
Fixes #575
## Summary

Replaces safe per-row scan status updates with bulk updates while preserving worker-aware behavior for running scans.

## Changes

- Bulk-pause queued scans in one database update
- Use `RowsAffected` for the pause flash count
- Bulk-resume paused scan rows before enqueueing the returned scans
- Restore a scan back to paused if resume enqueue fails
- Bulk-cancel queued scans in `scansCancelAll`
- Keep running scan cancellation on the existing worker-aware per-row path
- Add/expand tests for pause, resume and cancel behavior

## Tests

- `go test ./internal/web -run 'TestScans(PauseQueued|ResumePaused|CancelAll)|TestScanCancel'`
- `go test ./...`
- `go vet ./...`
- `git diff --check`
- `golangci-lint run`

### DIFF
--- a/internal/web/scans.go
+++ b/internal/web/scans.go
@@ -2,6 +2,7 @@ package web
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -11,6 +12,9 @@ import (
 
 	"scrutineer/internal/db"
 	"scrutineer/internal/worker"
+
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 func (s *Server) jobs(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


andrew (approved):
Reviewed. No blocking findings. Targeted scan status tests, related worker account-resume tests, full test suite, go vet, and git diff --check passed.

Issue #115, opened 2026-04-30
Per-repo scan config stored on the repository row (extends #66)


andrew:
#66 covers path scoping. Generalise it to a per-repo config blob the analyst can edit in the UI and the skills read from `context.json`:

```yaml
focus_areas:
  - name: parser
    paths: [src/parse/**]
    surface: "accepts arbitrary bytes from network"
known_bugs:
  - GHSA-xxxx-yyyy   # don't re-report these
  - "null deref in foo() with empty input, wontfix per #432"
attack_surface: >
  Single-tenant CLI tool. Config file is operator-trusted.
  Only stdin and the -f argument are attacker-controlled.
skip:
  - vendor/**
  - tests/**
```

`focus_areas` seeds the recon skill (which can add more). `known_bugs` feeds the dedup judge and the find prompt so agents don't re-report wontfixes. `attack_surface` is a human-written sentence the threat-model skill takes as ground truth instead of inferring. `skip` is #66's path filter.

For repos where nobody's written one, `recon` and `threat-model` generate it; for repos an analyst cares about, they edit it once and every subsequent scan is sharper.

andrew:
Landed in #620.

PR #620, opened 2026-07-10
Add repository scan configuration

abhinavgautam01:
Part of #115

## Summary

Adds analyst-managed repository scan configuration that carries persistent security-review guidance into skill workspaces and applies repository-specific path exclusions before scans run.

## Changes

- Add `Repository.ScanConfig` for validated YAML configuration
- Add a repository-page Scan config tab with save and clear actions
- Support:
  - `focus_areas`
  - `known_bugs`
  - `attack_surface`
  - `skip`
- Stage parsed config in every workspace as `scrutineer.scan_config`
- Layer repository `skip` patterns onto existing per-skill path filters
- Update `vuln-scan`, `threat-model`, `security-deep-dive`, and `finding-dedup` to use the new guidance
- Let completed root `threat-model` scans seed a valid config only when no analyst config exists
- Preserve analyst configuration with a conditional database update
- Document the repository column and `context.json` contract

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `go test ./internal/worker -run 'TestBundledSchemas_(compileAndAcceptSamples|rejectBadShapes)' -count=1`
- `git diff --check`

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -46,6 +46,7 @@ The central entity. One row per git URL.
 | clone_error | text | Last clone/fetch failure message; non-empty means the repo is currently unreachable. Cleared on next successful clone. |
 | disk_bytes | integer | Cached on-disk size of the persistent clone cache, so the repo list renders the disk badge from a column instead of walking each repo's cache per row. Refreshed by the worker after each scan and backfilled once at startup; 0 for local repos and remote repos not scanned since the column was added. |
 | threat_model | text | Operator's working-copy threat-model JSON. When set, the worker writes it to `./threat_model.json` in every skill workspace and `security-deep-dive` loads it instead of fetching the latest `threat-model` scan. Edited via the threat-model workbench tab. Empty = no override. |
+| scan_config | text | Analyst-authored YAML with `focus_areas`, `known_bugs`, `attack_surface`, and `skip` patterns. Validated by the repository editor, staged as `scrutineer.scan_config` in every skill workspace, and its `skip` patterns layer onto each skill's path-filter deny list. Empty = no repository-specific guidance. |
 | created_at | datetime | |
 | updated_at | datetime | |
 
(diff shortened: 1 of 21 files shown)


andrew (changes requested):
One behaviour gap and one nit.

`internal/web/repo_scan_config_seed.go:34` — `WHERE id = ? AND scan_config = ''` won't match repositories that existed before this column landed: AutoMigrate leaves the new column NULL on existing rows, and `NULL = ''` is false in SQLite. So pre-existing repos never get auto-seeded from a threat-model run. The pattern elsewhere is `(col = '' OR col IS NULL)` — see `internal/db/db.go:1315`, `:1339`, `internal/web/server.go:1183`. Should be:

```go
Where("id = ? AND (scan_config = '' OR scan_config IS NULL)", scan.RepositoryID)
```

The seed test starts from a fresh `db.Repository{}` create (which writes `''`), so it doesn't exercise the NULL case.

`internal/repoconfig/config.go:48` — `err != io.EOF` should be `!errors.Is(err, io.EOF)`. Works today because yaml.v3 returns the sentinel directly, but worth matching the errorlint convention while you're in there.

Rest looks good — the path validation, skip layering, and JSON-through-YAML seed path all check out.

abhinavgautam01:
Thanks, fixed

- The conditional seed update now matches both empty-string and legacy `NULL` `scan_config` values, while still preserving any analyst-authored config.
- Added a regression test that explicitly sets `scan_config` to `NULL` before finalizing a threat-model proposal.
- Updated the YAML decoder EOF check to use `errors.Is(err, io.EOF)`.

andrew (approved):
Both changes applied — NULL-safe seed guard and `errors.Is(err, io.EOF)` — plus a test for the legacy-NULL path. Thanks.

PR #621, opened 2026-07-10
Add parsing sinks to deep dive

### LINKED ISSUE #425: security-deep-dive: add parsing/format-reader sink class
The Phase 1 sink-class list in `skills/security-deep-dive/SKILL.md` names deserialisation, validation, round-trip integrity, memory safety, agentic, etc. A hand-rolled parser (config format, wire protocol, file header, regex-based extractor) that isn't half of a round-trip pair and isn't in an unsafe language never enters the inventory, so Phase 2 never judges it.

Add a "Parsing / format readers" class to the Phase 1 list: every place the code turns bytes or text from outside into structured data. The sink is the parse function; the danger is malformed input producing wrong structure, unbounded allocation, or a downstream value the rest of the code trusts.

This came from upstream feedback on a report where the project's parsers weren't flagged at all. It changes what the audit catalogues, so it won't retroactively fix existing reports; a re-scan picks parsers up.

abhinavgautam01:
Fixes #425

## Summary

Expands the `security-deep-dive` Phase 1 inventory so hand-rolled parsers and format readers are examined even when they are neither unsafe deserializers nor part of a parse/serialize round-trip.

## Changes

- Add a `Parsing / format readers` sink class to the deep-dive prompt
- Cover configuration and manifest formats, wire protocols, file headers, ad-hoc text formats, and regex-based extractors
- Call out malformed-input structure errors, ambiguous or partial interpretation, and input-controlled allocation/work as parser-specific risks
- Include parser entry points in the exhaustive grep guidance
- Clarify that semgrep rules do not replace parser/format-reader inventory work

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`


andrew (approved):
Matches #425 — parsing/format-reader class slotted after deserialisation, threaded into the semgrep-miss list and grep guidance. Merges clean with #620's edits to the same file.

PR #623, opened 2026-07-10
Add auxiliary structured model calls

### LINKED ISSUE #63: Lightweight auxiliary model helper for structured one-shot calls
Today the only way to call a model is the full `worker/claude.go` path: spin up a workspace, run claude-code as an agent, stream events. That is too heavy for small structured questions like "are these two findings the same root cause" or "summarise these 40 semgrep hits into themes".

Add `internal/llm/haiku.go` (or similar) that does a direct single-turn HTTP call to the Messages API:

- Takes a prompt, a JSON schema, and a max_tokens budget
- Prefills the assistant turn with `{` or `[` based on the schema root to force JSON
- Tolerant extraction that strips code fences and prose around the JSON before unmarshalling
- Validates the result against the schema, returns typed data or an error
- Records token usage and cost on the calling `Scan` row

This unblocks finding dedup, the eval judge, and ad-hoc enrichment without paying for a full agent loop each time.

- [ ] `internal/llm` package with `Call(ctx, prompt, schema, opts) (json.RawMessage, Usage, error)`
- [ ] JSON extraction helper with tests for fenced / prose-wrapped / bare output
- [ ] Usage struct with input/output/cache-read/cache-write token counts
- [ ] Wire cost into `Scan.CostUSD` when called from a worker context

abhinavgautam01:
Fixes #63

## Summary

Adds a lightweight direct Anthropic Messages API helper for structured one-shot model calls that do not need a full agent workspace or tool loop.

## Changes

- Add `internal/llm` with `Call(ctx, prompt, schema, opts)`.
- Request schema-constrained JSON with Anthropic structured outputs using `output_config.format.type: json_schema`.
- Send a single user prompt without unsupported assistant-turn prefilling.
- Return the structured JSON response after confirming it is valid JSON.
- Preserve usage when the API response decodes successfully but its structured text is invalid, so billable requests remain accounted for.
- Add `Worker.CallAuxiliary` to persist auxiliary-call usage and cost on the active `Scan` row.
- Keep Anthropic fresh input, cache-read input, and cache-write input separate in scan usage accounting while normalizing a copy for pricing.
- Default an auxiliary call to the scan model and reject explicit model mismatches.
- Add Anthropic cache-creation pricing, including Haiku, while preserving zero cache-write cost for OpenAI models.
- Reject accounting updates when the target scan no longer exists.

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`

### DIFF
--- a/internal/llm/messages.go
+++ b/internal/llm/messages.go
@@ -10,8 +10,6 @@ import (
 	"io"
 	"net/http"
 	"strings"
-
-	"github.com/santhosh-tekuri/jsonschema/v6"
 )
 
 const (
@@ -44,9 +42,19 @@ type Usage struct {
 }
 
 type messageRequest struct {
-	Model     string    `json:"model"`
-	MaxTokens int       `json:"max_tokens"`
-	Messages  []message `json:"messages"`
+	Model        string       `json:"model"`
+	MaxTokens    int          `json:"max_tokens"`
+	Messages     []message    `json:"messages"`
+	OutputConfig outputConfig `json:"output_config"`
+}
+
+type outputConfig struct {
+	Format structuredOutputFormat `json:"format"`
+}
+
+type structuredOutputFormat struct {
+	Type   string          `json:"type"`
+	Schema json.RawMessage `json:"schema"`
 }
 
 type message struct {
@@ -71,17 +79,16 @@ type apiError struct {
 	} `json:"error"`
 }
 
-// Call sends prompt as one user turn and asks the API to continue an assistant
-// JSON value. The response is extracted tolerantly, then validated against
-// schema before being returned. Usage is returned even when extraction or
-// schema validation fails, so callers can still record billable requests.
+// Call sends prompt as one user turn and requests a response constrained to
+// schema through Anthropic structured outputs. Usage is returned even when a
+// provider response is malformed, so callers can still record billable
+// requests.
 //
 // POST calls are intentionally not retried: the provider may have accepted a
 // timed-out request, and retrying would risk paying for duplicate inference.
 func Call(ctx context.Context, prompt string, schema json.RawMessage, opts Options) (json.RawMessage, Usage, error) {
-	prefix, err := schemaPrefix(schema)
-	if err != nil {
-		return nil, Usage{}, err
+	if !json.Valid(schema) {
+		return nil, Usage{}, fmt.Errorf("llm: schema is not valid JSON")
 	}
 	if strings.TrimSpace(opts.APIKey) == "" {
 		return nil, Usage{}, fmt.Errorf("llm: API key is required")
@@ -102,8 +109,10 @@ func Call(ctx context.Context, prompt string, schema json.RawMessage, opts Optio
 		MaxTokens: opts.MaxTokens,
 		Messages: []message{
 			{Role: "user", Content: prompt},
-			{Role: "assistant", Content: prefix},
 		},
+		OutputConfig: outputConfig{Format: structuredOutputFormat{
+			Type: "json_schema", Schema: schema,
+		}},
 	})
 	if err != nil {
 		return nil, Usage{}, fmt.Errorf("llm: marshal request: %w", err)
@@ -137,35 +146,11 @@ func Call(ctx context.Context, prompt string, schema json.RawMessage, opts Optio
 	if err := json.Unmarshal(data, &decoded); err != nil {
 		return nil, Usage{}, fmt.Errorf("llm: decode response: %w", err)
 	}
-	text := responseText(decoded.Content)
-	result, err := ExtractJSON(text)
-	if err != nil {
-		result, err = ExtractJSON(prefix + text)
-	}
-	if err != nil {
-		return nil, decoded.Usage, err
-	}
-	if err := validateSchema(schema, result); err != nil {
-		return nil, decoded.Usage, err
+	result := json.RawMessage(strings.TrimSpace(responseText(decoded.Content)))
+	if !json.Valid(result) {
+		return nil, decoded.Usage, fmt.Errorf("llm: structured response is not valid JSON")
 	}
 	return result, decoded.Usage, nil
-}
-
-func schemaPrefix(schema json.RawMessage) (string, error) {
-	var root struct {
-		Type string `json:"type"`
-	}
-	if err := json.Unmarshal(schema, &root); err != nil {
-		return "", fmt.Errorf("llm: schema is not valid JSON: %w", err)
-	}
-	switch root.Type {
-	case "object":
-		return "{", nil
-	case "array":
-		return "[", nil
-	default:
-		return "", fmt.Errorf("llm: schema root type must be object or array, got %q", root.Type)
-	}
 }
 
 func responseText(blocks []contentBlock) string {
@@ -176,45 +161,6 @@ func responseText(blocks []contentBlock) string {
 		}
 	}
 	return text.String()
-}
-
-// ExtractJSON finds the first complete object or array embedded in text. It
-// accepts bare JSON, fenced JSON, and model prose surrounding the value.
-func ExtractJSON(text string) (json.RawMessage, error) {
-	for i := range text {
-		if text[i] != '{' && text[i] != '[' {
-			continue
-		}
-		dec := json.NewDecoder(strings.NewReader(text[i:]))
-		var value json.RawMessage
-		if err := dec.Decode(&value); err == nil {
-			return value, nil
-		}
-	}
-	return nil, fmt.Errorf("llm: response does not contain a JSON object or array")
-}
-
-func validateSchema(schema, value json.RawMessage) error {
-	compiler := jsonschema.NewCompiler()
-	doc, err := jsonschema.UnmarshalJSON(bytes.NewReader(schema))
-	if err != nil {
-		return fmt.Errorf("llm: schema is not valid JSON: %w", err)
-	}
-	if err := compiler.AddResource("schema.json", doc); err != nil {
-		return fmt.Errorf("llm: load schema: %w", err)
-	}
-	compiled, err := compiler.Compile("schema.json")
-	if err != nil {
-		return fmt.Errorf("llm: compile schema: %w", err)
-	}
-	instance, err := jsonschema.UnmarshalJSON(bytes.NewReader(value))
-	if err != nil {
-		return fmt.Errorf("llm: extracted response is not valid JSON: %w", err)
-	}
-	if err := compiled.Validate(instance); err != nil {
-		return fmt.Errorf("llm: response does not match schema: %w", err)
-	}
-	return nil
 }
 
 func responseError(resp *http.Response) error {
--- a/internal/llm/messages.go
+++ b/internal/llm/messages.go
@@ -1,0 +1,244 @@
+// Package llm provides small, structured one-shot model calls for work that
+// does not need a full agent workspace or tool loop.
+package llm
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/santhosh-tekuri/jsonschema/v6"
+)
+
+const (
+	DefaultEndpoint = "https://api.anthropic.com/v1/messages"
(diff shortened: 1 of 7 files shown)


internal/llm/messages.go
```diff
@@ -0,0 +1,244 @@
...
+
+	endpoint := opts.Endpoint
+	if endpoint == "" {
+		endpoint = DefaultEndpoint
+	}
+	body, err := json.Marshal(messageRequest{
+		Model:     opts.Model,
+		MaxTokens: opts.MaxTokens,
+		Messages: []message{
+			{Role: "user", Content: prompt},
```
andrew, line 111: Last-assistant-turn prefill returns `400 invalid_request_error` on Sonnet 4.6, Sonnet 5, Opus 4.6/4.7/4.8, and Fable 5, which is every Anthropic model in `modelPricing`. The tests hit an `httptest.Server` mock so they don't catch it, but a real call with `claude-sonnet-4-6` fails.

The API-native way to get schema-shaped JSON is structured outputs: add `"output_config": {"format": {"type": "json_schema", "schema": <schema>}}` to the request body and drop the assistant message. The response text block is then guaranteed to match the schema, so `schemaPrefix`, `ExtractJSON`, and `validateSchema` all go away and `Call` shrinks to marshal → POST → read `content[0].text`.

(Prefill does still work on Haiku 4.5, which is what #63 originally named, but see the pricing comment.)

internal/worker/pricing.go
```diff
@@ -55,7 +55,9 @@ const perMillion = 1e6
 // total prompt token count and CacheReadTokens is the cached subset of
 // it, so uncached = InputTokens - CacheReadTokens. That is what codex's
 // turn.completed usage reports. CacheWriteTokens is not billed
-// separately by OpenAI and codex does not report it, so it is ignored.
+// separately by OpenAI and codex does not report it, so those pricing rows
+// leave CacheWrite at zero. Anthropic's Messages API does bill cache creation,
+// and auxiliary calls report it in CacheWriteTokens.
```
andrew: This function computes `uncached = InputTokens - CacheReadTokens`, which is correct for codex/OpenAI where `input_tokens` is the total prompt count. Anthropic's Messages API reports `input_tokens` as the uncached remainder only (total = `input_tokens + cache_read + cache_creation`), so feeding Anthropic usage in here double-discounts cache reads.

It's latent on the auxiliary path today because `llm.Call` sends no `cache_control` and cache tokens are always 0, but this updated docstring plus the new exported `CostFromUsage` invite callers to pass Anthropic usage directly. Either normalise to codex-shape usage in `recordAuxiliaryUsage` (add cache_read + cache_write into InputTokens before calling), or keep this function unexported and codex-only.

internal/worker/pricing.go
```diff
@@ -24,15 +24,15 @@ type modelPrice struct {
...
 var modelPricing = map[string]modelPrice{
 	// Anthropic
-	"claude-opus-4-6":   {In: 5.00, Out: 25.00, CachedIn: 0.50},
-	"claude-opus-4-7":   {In: 5.00, Out: 25.00, CachedIn: 0.50},
-	"claude-opus-4-8":   {In: 5.00, Out: 25.00, CachedIn: 0.50},
-	"claude-sonnet-4-6": {In: 3.00, Out: 15.00, CachedIn: 0.30},
+	"claude-opus-4-6":   {In: 5.00, Out: 25.00, CachedIn: 0.50, CacheWrite: 6.25},
+	"claude-opus-4-7":   {In: 5.00, Out: 25.00, CachedIn: 0.50, CacheWrite: 6.25},
+	"claude-opus-4-8":   {In: 5.00, Out: 25.00, CachedIn: 0.50, CacheWrite: 6.25},
+	"claude-sonnet-4-6": {In: 3.00, Out: 15.00, CachedIn: 0.30, CacheWrite: 3.75},
```
andrew, line 30: `claude-haiku-4-5` is missing from this table. #63 named the helper `haiku.go` because cheap auxiliary calls are the point; with Haiku absent, `CallAuxiliary` on Haiku records tokens but `costFromUsage` returns $0. Rates are $1.00 / $5.00 / $0.10 / $1.25 per MTok (in / out / cache-read / cache-write).

internal/worker/pricing.go
```diff
@@ -67,9 +69,16 @@ func costFromUsage(model string, u Usage) float64 {
...
 	return (float64(uncached)*p.In +
 		float64(u.CacheReadTokens)*p.CachedIn +
+		float64(u.CacheWriteTokens)*p.CacheWrite +
 		float64(u.OutputTokens)*p.Out) / perMillion
 }
 
+// CostFromUsage exposes the scan pricing calculation to worker helpers that
+// account for direct model calls outside the harness event stream.
+func CostFromUsage(model string, u Usage) float64 {
+	return costFromUsage(model, u)
```
andrew: Rather than an exported wrapper around an identical unexported function, just rename `costFromUsage` → `CostFromUsage` and update the two internal callers.

andrew (changes requested):
Thanks for picking this up. The package shape and the worker/DB accounting side are fine, but the prefill approach doesn't work against the real API on the models this would target, and the cost function has a semantic mismatch when fed Anthropic usage.

internal/llm/messages.go
abhinavgautam01: `Call` now sends a single user message with Anthropic structured outputs:

`output_config.format = {"type":"json_schema","schema": ...}`

The assistant prefill path is removed, along with `schemaPrefix`, tolerant extraction and local schema validation. The helper now returns the structured text block directly, with a defensive JSON-validity check for malformed proxy/provider responses. Tests assert the request has one user message and the expected `json_schema` output configuration.

internal/worker/pricing.go
abhinavgautam01: `recordAuxiliaryUsage` now normalizes Anthropic usage before pricing: worker `InputTokens` is the total prompt count (`input_tokens + cache_read_input_tokens + cache_creation_input_tokens`), while the cache fields remain subsets.

`CostFromUsage` now subtracts cache-write tokens from normal input only for models with a dedicated cache-write price. OpenAI rows keep cache-write tokens as ordinary input because their `CacheWrite` rate is zero. Added coverage for both Anthropic cache creation and the OpenAI behavior.

internal/worker/pricing.go
abhinavgautam01: Added `claude-haiku-4-5` pricing at `$1.00 / $5.00 / $0.10 / $1.25` per MTok for input, output, cache read and cache write respectively, plus a focused pricing test.

internal/worker/pricing.go
abhinavgautam01: Renamed `costFromUsage` to `CostFromUsage` and updated the event-loop caller, removing the redundant wrapper.

internal/llm/messages.go
```diff
@@ -0,0 +1,190 @@
...
+		},
```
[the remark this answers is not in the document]
abhinavgautam01: This was superseded by the earlier maintainer review: assistant-turn prefill fails on the supported Sonnet/Opus/Fable models. The helper now sends one user message with Anthropic `output_config.format.type: json_schema`, which is the supported API-native replacement. The request test covers that shape.

internal/llm/messages.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	var decoded messageResponse
+	if err := json.Unmarshal(data, &decoded); err != nil {
+		return nil, Usage{}, fmt.Errorf("llm: decode response: %w", err)
+	}
+	result := json.RawMessage(strings.TrimSpace(responseText(decoded.Content)))
+	if !json.Valid(result) {
+		return nil, decoded.Usage, fmt.Errorf("llm: structured response is not valid JSON")
+	}
+	return result, decoded.Usage, nil
```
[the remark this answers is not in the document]
abhinavgautam01: This was intentionally replaced by Anthropic structured outputs following the maintainer review. The provider constrains the response with the supplied JSON Schema, so the previous prefill/extraction/manual-validation path was removed rather than retained alongside it.

internal/llm/messages.go
```diff
@@ -0,0 +1,190 @@
...
+		Type    string `json:"type"`
+		Message string `json:"message"`
+	} `json:"error"`
+}
+
+// Call sends prompt as one user turn and requests a response constrained to
+// schema through Anthropic structured outputs. Usage is returned even when a
+// provider response is malformed, so callers can still record billable
+// requests.
+//
```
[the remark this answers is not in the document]
abhinavgautam01: Fixed the comment to match the actual guarantee. Usage is retained when the API response decodes successfully but its structured text is invalid; a truncated or invalid API envelope cannot reliably provide usage and returns zero usage.

internal/worker/auxiliary.go
```diff
@@ -0,0 +1,69 @@
...
+
+func recordAuxiliaryUsage(gdb *gorm.DB, scan *db.Scan, model string, usage llm.Usage) error {
+	if usage == (llm.Usage{}) {
+		return nil
+	}
+	workerUsage := Usage{
+		// Anthropic reports uncached input separately from cache-read and
+		// cache-write input. Worker Usage records total prompt input, with
+		// the cache values as subsets, matching harness usage semantics.
+		InputTokens:      usage.InputTokens + usage.CacheReadTokens + usage.CacheWriteTokens,
```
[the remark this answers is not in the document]
abhinavgautam01: Scan `InputTokens` now stores Anthropic fresh input only, while cache-read and cache-write counters remain separate. This keeps `TotalInputTokens` and cache-hit reporting accurate.

internal/worker/auxiliary.go
```diff
@@ -0,0 +1,69 @@
...
+		OutputTokens:     usage.OutputTokens,
+		CacheReadTokens:  usage.CacheReadTokens,
+		CacheWriteTokens: usage.CacheWriteTokens,
+	}
+	cost := CostFromUsage(model, workerUsage)
```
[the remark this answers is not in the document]
abhinavgautam01: Cost calculation now uses a derived normalized usage value whose input count includes cache categories, while the original Anthropic counters are persisted unchanged on the scan.

internal/worker/auxiliary.go
```diff
@@ -0,0 +1,69 @@
...
+// because the request was still billable.
+func (w *Worker) CallAuxiliary(ctx context.Context, scan *db.Scan, prompt string, schema json.RawMessage, opts llm.Options) (json.RawMessage, error) {
+	if w == nil || w.DB == nil {
+		return nil, fmt.Errorf("auxiliary model call requires a worker database")
+	}
+	if scan == nil || scan.ID == 0 {
+		return nil, fmt.Errorf("auxiliary model call requires a persisted scan")
+	}
+	result, usage, callErr := llm.Call(ctx, prompt, schema, opts)
+	if err := recordAuxiliaryUsage(w.DB, scan, opts.Model, usage); err != nil {
```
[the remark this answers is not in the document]
abhinavgautam01: `CallAuxiliary` defaults an empty option model to `scan.Model` and rejects an explicit model that differs from the scan model, preventing ambiguous scan-level accounting.

internal/worker/auxiliary_test.go
```diff
@@ -0,0 +1,68 @@
...
+		t.Fatal("CallAuxiliary succeeded, want malformed-response error")
+	}
+
+	var got db.Scan
+	if err := gdb.First(&got, scan.ID).Error; err != nil {
+		t.Fatal(err)
+	}
+	if got.InputTokens != 150 || got.OutputTokens != 10 || got.CacheReadTokens != 20 || got.CacheWriteTokens != 30 {
+		t.Errorf("usage = in:%d out:%d read:%d write:%d", got.InputTokens, got.OutputTokens, got.CacheReadTokens, got.CacheWriteTokens)
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: The test now expects `InputTokens` to remain 100, with cache-read and cache-write tokens recorded separately. It also verifies that an omitted auxiliary model defaults to the scan model.

alexandre-daubois (approved):

andrew (approved):

Issue #108, opened 2026-04-30
Recon pass that emits focus areas, not just sub-paths


andrew:
`subprojects` partitions a repo by directory (monorepo packages, workspaces). That's the right unit for build isolation but the wrong unit for audit coverage. The libexpat run in #90 is the example: `lib/` and `xmlwf/` are one package but two completely different attack surfaces, and within `lib/` the interesting partition is by parser subsystem (entity handling, DTD, namespace, encoding conversion) rather than by file.

Add a `recon` skill that runs before `security-deep-dive` and writes a list of focus areas to the repo record. A focus area is a named input-processing subsystem with a one-line attack-surface note and a set of path globs, e.g.

```json
{"name": "DTD parsing", "surface": "external entity declarations from untrusted XML", "paths": ["lib/xmlparse.c:4000-6000", "lib/xmlrole.c"]}
```

Then `security-deep-dive` (or the narrow audit skills proposed in #106) takes a `focus_area` parameter and runs once per area instead of once per repo. Parallel deep-dives over disjoint focus areas instead of one agent trying to inventory everything. This is the actual fix for #20 as well as #90.

The recon agent should be cheap: short turn budget, read-only, no execution. Its only job is to read entry points and emit the partition.

PR #625, opened 2026-07-10
Add focus-area recon skill for deep-dive audits

abhinavgautam01:
Part of #108

## Summary

Adds a lightweight, read-only `recon` skill that maps a repository’s externally reachable input-processing subsystems before a deep-dive audit runs.

`recon` feeds focused, normalized security-surface guidance into `threat-model`; `threat-model` remains the sole author of the durable repository `scan_config`.

This makes audit scope follow security surfaces such as parsers, request handlers, deserializers, archive extraction and plugin loading rather than arbitrary directory partitions.

Focus-area paths intentionally use repository-relative globs rather than line ranges. This matches the durable `repoconfig.FocusArea` contract; line-level fan-out into separate deep-dive scans remains follow-up work under #108.

## Changes

- Add bundled `recon` skill with:
  - Mid-tier model preference
  - 12-turn cap
  - Read-only workspace constraints
  - Strict JSON schema for focus-area output
- Run `recon` during code-repository triage for code repositories.
- Make `recon` a `threat-model` prerequisite.
- Stage a valid recon report for `threat-model`, preferring a same-group sibling and otherwise falling back to the latest completed report with the same ref and subpath.
- Run `recon` in diff-rescan groups so threat-model receives a fresh focus map for the new commit.
- Validate recon reports against the producing skill’s schema and normalize focus-area path globs before staging.
- Keep `threat-model` as the sole proposer of repository `scan_config`, preserving its `known_bugs`, `attack_surface` and `skip` guidance.
- Preserve analyst-authored scan configuration as authoritative.
- Document the recon-to-threat-model handoff and bundled skill role.

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`


internal/web/repo_scan_config_seed.go
```diff
@@ -36,3 +36,7 @@ func (s *Server) autoSeedRepoScanConfig(scan *db.Scan) {
 		s.Log.Warn("save scan config proposal", "scan", scan.ID, "err", err)
 	}
 }
+
+func scanConfigProposer(skillName string) bool {
```
andrew: This makes both `recon` and `threat-model` scan_config proposers, but `autoSeedRepoScanConfig` is first-writer-wins (`WHERE scan_config = '' OR scan_config IS NULL`). Triage enqueues both under `has_code` and they run concurrently; recon is `mid`/12-turn and will finish first almost every time, seeding `focus_areas` only. Threat-model's proposal, which per its SKILL.md also carries `known_bugs`, `attack_surface`, and `skip`, then finds `scan_config` non-empty and is silently dropped. That's a regression from today where threat-model's full proposal lands.

Pick one owner: either drop `threatModelSkillName` here and strip the "propose a scan_config" paragraph from `skills/threat-model/SKILL.md`, or make recon a `requires` prerequisite of threat-model (instead of deep-dive) so threat-model reads recon's areas from `context.json` and remains the sole proposer. If both should keep proposing, the seed needs to merge non-overlapping keys rather than first-wins.

internal/web/repo_scan_config_seed_test.go
```diff
@@ -26,6 +26,25 @@ func TestAutoSeedRepoScanConfig(t *testing.T) {
 	}
 }
 
+func TestAutoSeedRepoScanConfigFromRecon(t *testing.T) {
```
andrew: This exercises the same code path `TestAutoSeedRepoScanConfig` already covers with a different skill name. The new behaviour is two proposers completing on the same repo; once the ownership question above is settled, this test should cover whichever interaction survives (e.g. recon seeds, then threat-model completes and its proposal is discarded/merged as intended).

skills/security-deep-dive/SKILL.md
```diff
@@ -9,6 +9,7 @@ metadata:
   scrutineer.max_turns: 120
   scrutineer.model: max
   scrutineer.requires:
+    - recon
```
andrew: Not blocking: this makes deep-dive wait for recon, but deep-dive still runs once and reads focus areas via `context.json`. #108 asks for one deep-dive per focus area; if that's a follow-up PR that's fine, just flagging that `requires` alone doesn't change how the areas are consumed.

andrew (changes requested):
Thanks for this, the recon skill itself and the schema/validation wiring look right. One blocking issue in how it interacts with threat-model's existing scan_config proposal.

internal/web/repo_scan_config_seed.go
abhinavgautam01: `threat-model` is now the sole `scan_config` proposer.

`recon` emits only its focus-area report and no longer writes `Repository.ScanConfig`. I moved the prerequisite from `security-deep-dive` to `threat-model`; the worker loads and validates the latest completed recon report, then stages it as `scrutineer.recon` in the threat-model workspace. Threat-model carries those areas into its complete proposal alongside `known_bugs`, `attack_surface` and `skip`.

internal/web/repo_scan_config_seed_test.go
abhinavgautam01: Replaced the redundant same-path test with an interaction test covering the intended ownership boundary: recon completion cannot seed repository config and the subsequent threat-model proposal is the only one accepted.

skills/security-deep-dive/SKILL.md
abhinavgautam01: Agreed. This PR establishes the recon -> threat-model -> scan-config handoff only; `security-deep-dive` still consumes the resulting map in one run. Fan-out to one deep-dive scan per focus area needs separate scan/UI orchestration, so I will keep that as the follow-up portion of #108.

internal/worker/skill.go
```diff
@@ -1219,6 +1245,44 @@ func StageWorkspace(workRoot, skillDir, apiBase, forkOrg, metadataDir string, sc
...
 
+func (w *Worker) reconContext(scan *db.Scan, skill *db.Skill) (*skillContextRecon, error) {
+	if skill.Name != "threat-model" || scan.SubPath != "" || scan.Ref != "" {
+		return nil, nil
+	}
+	var reconScan db.Scan
+	err := w.DB.Select("report").
+		Where("repository_id = ? AND skill_name = ? AND status = ? AND sub_path = '' AND ref = '' AND report <> ''",
+			scan.RepositoryID, "recon", db.ScanDone).
+		Order("id DESC").First(&reconScan).Error
```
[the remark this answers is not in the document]
abhinavgautam01: The recon lookup now selects `id`, `report` and `skill_id`, so warning logs retain the actual recon scan ID. When the threat-model scan belongs to a scan group, the lookup is restricted to the same group, preventing it from staging a recon result from another batch. Added coverage with a newer recon scan in a different group to prove it is ignored.

internal/worker/skill.go
```diff
@@ -1219,6 +1245,44 @@ func StageWorkspace(workRoot, skillDir, apiBase, forkOrg, metadataDir string, sc
...
+func validateReconFocusAreas(areas []repoconfig.FocusArea) error {
+	raw, err := json.Marshal(struct {
+		FocusAreas []repoconfig.FocusArea `json:"focus_areas"`
+	}{FocusAreas: areas})
+	if err != nil {
+		return err
+	}
+	_, _, err = repoconfig.Normalise(string(raw))
+	return err
+}
```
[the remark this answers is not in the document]
abhinavgautam01: Recon reports are now validated against the `SchemaJSON` from the skill row that produced the selected scan, which enforces required top-level keys such as `focus_areas` and `notes`. After schema validation, focus areas pass through `repoconfig.Normalise` and the normalized paths are staged. Tests cover backslash/whitespace normalization, traversal rejection and rejection of a report missing `notes`.

internal/worker/skill.go
```diff
@@ -1219,6 +1245,64 @@ func StageWorkspace(workRoot, skillDir, apiBase, forkOrg, metadataDir string, sc
...
+	if skill.Name != "threat-model" || scan.SubPath != "" || scan.Ref != "" {
+		return nil, nil
+	}
+	var reconScan db.Scan
+	query := w.DB.Select("id, report, skill_id").
+		Where("repository_id = ? AND skill_name = ? AND status = ? AND sub_path = '' AND ref = '' AND report <> ''",
+			scan.RepositoryID, "recon", db.ScanDone)
+	if scan.ScanGroup != "" {
+		query = query.Where("scan_group = ?", scan.ScanGroup)
+	}
```
alexandre-daubois: `reconContext`'s test only exercises the named-scan-group branch, but in production recon and threat-model are both enqueued via apiRunSkill with no scan_group, so they always hit the `scan.ScanGroup == ""` branch that no test covers. And because the wrong-group recon scan is created before the in-group one, `ORDER BY id DESC` picks the right scan even without the group filter, so the test would still pass if the `Where("scan_group = ?")` were removed. You should add a case where recon and threat-model both have empty groups, and make the wrong-group recon newer than the in-group one so it actually proves group isolation.

abhinavgautam01: `reconContext` now always filters by `scan_group`, including the empty-group production path used by API-triggered recon and threat-model scans.

The test now creates matching `batch-a` and ungrouped recon reports, followed by a deliberately newer `other`-group report. It verifies that both grouped and ungrouped threat-model scans select only their matching recon report, so removing the scan-group filter makes the test fail.

internal/worker/skill.go
```diff
@@ -1219,6 +1245,62 @@ func StageWorkspace(workRoot, skillDir, apiBase, forkOrg, metadataDir string, sc
...
+	if skill.Name != "threat-model" || scan.SubPath != "" || scan.Ref != "" {
+		return nil, nil
+	}
+	var reconScan db.Scan
+	query := w.DB.Select("id, report, skill_id").
+		Where("repository_id = ? AND skill_name = ? AND status = ? AND sub_path = '' AND ref = '' AND report <> ''",
+			scan.RepositoryID, "recon", db.ScanDone).
+		Where("scan_group = ?", scan.ScanGroup)
+	err := query.Order("id DESC").First(&reconScan).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
```
andrew: This is stricter than the prereq gate. `unsatisfiedPrereqs` checks the same `scan_group` first then falls back repo-wide, so a threat-model rerun proceeds when *any* recon has completed on the repo. But this only matches the exact group with no fallback:

- Diff rescan (`repoDiffScan` enqueues threat-model/semgrep/deep-dive in a fresh group with no recon): prereq gate falls back repo-wide → satisfied → runs; this finds nothing in the group → stages nothing.
- Manual solo threat-model rerun (empty `ScanGroup`): prereq gate goes repo-wide → satisfied; this filters on `scan_group = ''`, which won't match the triage batch's recon.

So recon output only reaches threat-model on the initial triage run and every later rerun loses it. I'd mirror the prereq fallback: try same-group first, else latest `ScanDone` recon on the repo with the same `sub_path`/`ref`.

internal/worker/skill.go
```diff
@@ -1219,6 +1245,62 @@ func StageWorkspace(workRoot, skillDir, apiBase, forkOrg, metadataDir string, sc
 	return nil
 }
 
+func (w *Worker) reconContext(scan *db.Scan, skill *db.Skill) (*skillContextRecon, error) {
+	if skill.Name != "threat-model" || scan.SubPath != "" || scan.Ref != "" {
+		return nil, nil
+	}
```
andrew, line 1257: nit: if there's a constant for `"threat-model"` in this package (there's `threatModelSkillName` in web), use it here and for `"recon"` below.

skills/triage/SKILL.md
```diff
@@ -69,6 +69,7 @@ Only when `has_packages`:
 Only when `has_code`:
 
 - `subprojects`
+- `recon`
```
andrew, line 72: The diff-rescan set at `internal/web/server.go:2304` (`{threat-model, semgrep, deep-dive}`) should probably grow `recon` too now that threat-model consumes it. It's 12 turns on the mid tier — cheap enough to keep the focus map fresh at the new commit, and it fixes the same-group staging gap for that path.

andrew (changes requested):
The skill itself is well-written: tight schema, clear negative instructions, and the "focus areas are attack surfaces not directory partitions" framing is exactly right. Threading it through `threat-model` rather than writing `scan_config` directly keeps the analyst override authoritative. Go plumbing is clean and the tests cover schema-validation, path normalisation, and cross-group isolation. One design gap inline; smaller notes below.

**Scope vs #108** — the issue's example had `"paths": ["lib/xmlparse.c:4000-6000"]`; this PR forbids line numbers and uses path globs only. That's the right call given `repoconfig.FocusArea` is glob-shaped, but worth a line in the PR body since it's "part of" not "closes".

**Minor**
- `skills/recon/SKILL.md` tells recon to preserve analyst `scan_config.focus_areas` when present, but in that case threat-model already ignores recon and won't propose a config, so the instruction is dead. Harmless, just slightly confused about the pipeline.
- Schema requires `notes` but allows `[]`. Assuming that's deliberate to make the model consider it.

internal/worker/skill.go
abhinavgautam01: `reconContext` now mirrors prerequisite behavior: it first uses a completed recon scan from the same scan group, then falls back to the latest completed recon scan with the same repository, `sub_path` and `ref`.

Tests cover same-group preference, an ungrouped manual rerun, a grouped rerun with no sibling recon and ref/subpath-scoped fallback.

internal/worker/skill.go
abhinavgautam01: Added worker-local `threatModelSkillName` and `reconSkillName` constants and replaced the inline names in the recon handoff.

skills/triage/SKILL.md
abhinavgautam01: Diff rescans now enqueue `recon` in the same group before threat-model, semgrep, and security-deep-dive. The diff-rescan server test now verifies all four grouped scans.

abhinavgautam01:
Thanks, addressed the follow-ups.

The PR description now clarifies that recon focus areas intentionally use repository-relative path globs rather than line ranges, matching the durable `repoconfig.FocusArea` contract. Line-level fan-out into separate deep-dive runs remains follow-up work under #108.

I also removed the dead recon instruction about preserving analyst `scan_config.focus_areas`: threat-model already treats analyst configuration as authoritative and does not consume recon output in that case.

`notes` intentionally remains required while allowing `[]`, so recon must consider exclusions or caveats but can explicitly report that none apply.

andrew (approved):
All second-round feedback addressed: reconContext now mirrors the prereq-gate fallback (same-group then repo-wide scoped by ref/subpath), diff rescans enqueue recon in-group, and the tests prove group isolation with a newer wrong-group scan. Skill and schema look good.

PR #631, opened 2026-07-13
feat: release standalone binaries with embedded skills and profiles

p-linnane:
Running scrutineer today requires a Go toolchain and a source checkout because the built-in skills and per-ecosystem runner profiles load from the working tree. This adds a release path producing precompiled Linux and macOS binaries for amd64 and arm64 that an operator can download from GitHub Releases, verify against SHA256SUMS and the build-provenance attestation, and run from any directory. The runtime requirement is unchanged: docker, rootless podman, or Apple's `container` CLI stays external, and nothing is bundled into the executable.

The binary embeds `skills/` and `docker/profiles/` and materialises them at startup into content-addressed directories under the data dir, so the disk-backed loader and auxiliary-file staging keep working unchanged. Configured `-skills` directories and a skills repo override same-named bundled copies without version churn on restart, and local/remote precedence is preserved. `profiles-dir` keeps its precedence: an explicit flag or config value wins (`ProfilesDir` became `*string` so `profiles_dir: ""` genuinely disables profiles), a source checkout keeps using `docker/profiles`, and only a checkout-less binary falls back to the embedded copies. `go run ./cmd/scrutineer -skills ./skills` remains the development workflow. A new `scrutineer version` subcommand reports the version, source commit, build date and paired runner-image digest, and the settings page shows the release version in place of the build commit.

Releases use CalVer `vYYYY.MM.DD.N` tags dispatched manually from main. Preflight validates `VERSION`, refuses a tag pointing at another commit, runs the test suite, and resolves the runner image's commit-matched multi-platform manifest to an immutable digest injected as the released binary's default `--runner-image` (development builds keep `:latest`). Build jobs produce static CGO-free binaries for the four targets with provenance attestations. The release job deletes any incomplete same-version draft, uploads the archives and `SHA256SUMS` to a fresh draft, and publishes only after every upload succeeds, so failed runs are safely rerunnable and re-dispatching a published version is a no-op. GitHub Releases serve as the changelog. macOS artifacts are intentionally unsigned and unnotarized for now.

As discussed with @andrew: GitHub Releases act as the changelog (generated notes, so older releases can be backported later) and versioning is CalVer `YYYY.MM.DD.N` to avoid implied semver expectations. The flow supports a bi-weekly cadence whenever we want to start one.


alexandre-daubois:
`db.go` now documents and writes `source = "bundled"` (with a populated source_path), but `docs/database.md:117-118` still list only local, remote, ui and describe source_path as local/remote-only.

p-linnane:
Good catch. Thanks @alexandre-daubois. Resolved in the latest push.

andrew (approved):
Careful work: SHA-pinned actions, `persist-credentials: false`, per-job least-privilege, draft-then-publish, tag-target validation, length-prefixed content hashing with a framing test, atomic extraction with the concurrent-writer race handled, and old bundle dirs retained so existing skill rows keep resolving auxiliary files. The `LoadDirectoryExcept` override pattern with the version-churn test is exactly right.

Chased three things that all checked out: `actions/attest` with only `subject-path` defaults to SLSA build provenance, `ubuntu-slim` is a real GitHub-hosted runner and ships `gh`, and `new("")` is Go 1.26's `new(expr)`.

Non-blocking notes for a follow-up if you feel like it:

- `ubuntu-slim` has a platform-enforced 15-minute job cap; `timeout-minutes: 20` on the release job is dead. Should still finish well under 15.
- Settings page now shows `dev` instead of the commit SHA on non-release builds. `scrutineer version` covers it on the CLI, but the About panel used to be handy for bug reports. Could show both.
- Worth a line in `docs/skills.md`: you can no longer suppress a built-in skill in a checkout by deleting `skills/<name>/` since the embedded copy fills the gap; deactivate via the UI instead.

PR #632, opened 2026-07-13
feat(disclose): suggest file-level recipients from CODEOWNERS

### LINKED ISSUE #117: Route disclosure to file-level owners, not just repo maintainers
`maintainers` identifies who runs the repo. For large projects the person who owns `crypto/` is not the person who owns `cli/`, and a disclosure landing on the wrong desk sits for weeks.

In `disclose`, before drafting, look up the finding's file in `CODEOWNERS` (and `.github/CODEOWNERS`, `docs/CODEOWNERS`). If there's no CODEOWNERS, fall back to `git log -- {file} | head` for the most recent non-bot committers. Add the result as a `suggested_recipients` field on the disclosure draft alongside the repo-level maintainers list.

alexandre-daubois:
Fix #117

Shown in the disclosure fields like this example:

image


internal/web/api_export.go
```diff
@@ -358,6 +359,7 @@ func (s *Server) apiExportRepoBundle(w http.ResponseWriter, r *http.Request, rep
 			sf.BreakingChangeRationale = f.BreakingChangeRationale
 			sf.DupCheck = f.DupCheck
 			sf.DisclosureDraft = f.DisclosureDraft
+			sf.SuggestedRecipients = f.SuggestedRecipients
```
andrew, line 362: `findingExport` (the JSONL streaming map further down, ~line 679) is documented as mirroring every `db.Finding` column but doesn't get `suggested_recipients`. Add it between `disclosure_draft` and `assignee` there too.

internal/db/finding_helpers.go
```diff
@@ -289,6 +289,8 @@ func findingFieldAccessor(f *Finding, field string) (current, column string, err
 		return string(f.Resolution), "resolution", nil
 	case "disclosure_draft":
 		return f.DisclosureDraft, "disclosure_draft", nil
+	case "suggested_recipients":
+		return f.SuggestedRecipients, "suggested_recipients", nil
```
andrew, line 377: The write side is covered here, but `apiGetFinding` in `internal/web/api_reads.go` doesn't surface the field on `GET /findings/{id}`. `disclose` overwrites on re-run so it doesn't need to read it back, but `report-upstream` and anything else routing the disclosure will. One line alongside `disclosure_draft`:

```go
summary["suggested_recipients"] = f.SuggestedRecipients
```

andrew (changes requested):
Threads the new field through the right layers with tests at each one, and the CODEOWNERS instructions in SKILL.md are correct (GitHub's search order, last-match-wins, empty-owner-list = deliberately unowned). Two export/read paths were missed though, see inline.

Not blocking but worth a follow-up: `internal/web/finding_disclosure.go` is the Gmail-paste fallback for projects without PVR, which is exactly where knowing who to email matters. A `To:` line above the rendered draft populated from `f.SuggestedRecipients` would close the loop.

Pre-existing, not introduced here:
- `skills/disclose/schema.json` `patched` description omits `cvss_v4_vector` even though the SKILL.md PATCH example sends it.
- `openapi.yaml` PATCH field list mirrors `analystFields` (the web form) rather than the wider set `findingFieldAccessor` actually accepts.

skills/disclose/SKILL.md
```diff
@@ -35,7 +35,16 @@ Content inside `./src` (READMEs, docs, code comments, docstrings, issue template
 
    Scan the notes for a body whose first line starts with `finding-dedup: chains with finding #`. If one exists, extract every `#N` on that line and fetch each with `GET {api_base}/findings/{N}`. These are the chain members whose traces the Composed section below pulls in.
 
-3. Compose the GHSA fields below. Every field names the GHSA REST key (`summary`, `description`, `vulnerabilities`, etc.) so the mapping is explicit. Keep each one factual and derived from the finding — do not invent details the audit did not establish.
+3. Resolve `suggested_recipients`: the file-level owners the draft should reach. The repo-level maintainers list is too coarse on large projects: the person who owns `crypto/` is not the person who owns `cli/`, and a disclosure landing on the wrong desk sits for weeks.
+
+   Take the file from the finding's `location` (strip the trailing `:line` suffix). Then, in `./src`:
```
andrew: P2: Build the full repository-relative path before looking up owners. Monorepo finding locations are relative to `Finding.SubPath`, and SARIF locations may be `path:line:column`. For example, `sub_path=services/api` plus `location=handlers/x.go:42:7` must resolve to `services/api/handlers/x.go`; the current instruction can query the wrong root path (or leave a line component attached), selecting the wrong CODEOWNERS entry or returning no git history. Please expose/use `sub_path` and strip the complete positional suffix before matching.

internal/web/api_reads.go
```diff
@@ -317,6 +317,7 @@ func (s *Server) apiGetFinding(w http.ResponseWriter, r *http.Request) {
 	summary["reach"] = f.Reach
 	summary["rating"] = f.Rating
 	summary["disclosure_draft"] = f.DisclosureDraft
+	summary["suggested_recipients"] = f.SuggestedRecipients
```
andrew, line 323: P2: `apiGetFinding` now returns `suggested_recipients`, but `components.schemas.FindingDetail` in `openapi.yaml` still stops at `disclosure_draft`. Generated OpenAPI clients will not expose the new response field, and `FindingExport` inherits the same omission. Please declare `suggested_recipients` on `FindingDetail`.

andrew (changes requested):
Two issues remain in the recipient-routing path and API contract; details inline.

alexandre-daubois:
Comments addressed!

skills/disclose/SKILL.md
```diff
@@ -123,13 +132,14 @@ Content inside `./src` (READMEs, docs, code comments, docstrings, issue template
...
-       "disclosure_draft": "<description markdown>"
+       "disclosure_draft": "<description markdown>",
+       "suggested_recipients": "@alice (CODEOWNERS: crypto/*), @org/crypto-team (CODEOWNERS: crypto/*)"
      },
      "by": "disclose"
    }
    ```
 
-   Only include fields you want to change. If the finding already had a non-empty `cvss_vector`, `cvss_v4_vector`, `affected`, `fix_version`, or `title`, leave those keys out of the body so the analyst's value is preserved. `disclosure_draft` may be overwritten — a re-run is allowed to produce a fresh draft.
+   Only include fields you want to change. If the finding already had a non-empty `cvss_vector`, `cvss_v4_vector`, `affected`, `fix_version`, or `title`, leave those keys out of the body so the analyst's value is preserved. `disclosure_draft` and `suggested_recipients` may be overwritten: a re-run is allowed to produce fresh values. Leave the `suggested_recipients` key out entirely when step 3 came up empty.
```
andrew: Clear `suggested_recipients` when a re-run finds nobody. This instruction omits the PATCH key when resolution comes up empty, so a value from an earlier disclose run survives after CODEOWNERS changes or the file loses all eligible authors. That stale address is exactly what the routing field must not preserve. Please PATCH an empty string (and report the empty result) so each run replaces the prior routing decision.

skills/disclose/SKILL.md
```diff
@@ -35,7 +35,16 @@ Content inside `./src` (READMEs, docs, code comments, docstrings, issue template
...
 
-3. Compose the GHSA fields below. Every field names the GHSA REST key (`summary`, `description`, `vulnerabilities`, etc.) so the mapping is explicit. Keep each one factual and derived from the finding — do not invent details the audit did not establish.
+3. Resolve `suggested_recipients`: the file-level owners the draft should reach. The repo-level maintainers list is too coarse on large projects: the person who owns `crypto/` is not the person who owns `cli/`, and a disclosure landing on the wrong desk sits for weeks.
+
+   Take the file from the finding's `location` and strip the whole positional suffix, which may be `:line` or `:line:column` (`handlers/x.go:42:7` → `handlers/x.go`). When the finding's `sub_path` is non-empty, the location is relative to that sub-folder: prepend it to get the repository-relative path (`sub_path=services/api` → `services/api/handlers/x.go`). Use that repository-relative path for both routes below. Then, in `./src`:
```
andrew: Strip the range form of a finding location too. Scrutineer documents and emits `file:start-end` locations (`lib/x.rb:10-20`), but this instruction only covers `:line` and `:line:column`. A range suffix can therefore remain attached when matching CODEOWNERS and when passed to `git log`, producing no or incorrect recipients. Please normalize `:start-end` alongside the other positional suffixes.

andrew (changes requested):
Two issues remain in the recipient-routing path; details inline.

andrew (changes requested):
Both points from the last round are addressed; the `:start-end` suffix and the always-PATCH-empty-string change look right, and the `FindingDetail` openapi addition is a good catch.

`internal/db/finding_helpers_test.go` now conflicts with main: #669 merged today and added `TestWriteFindingField_concurrentUpdatesKeepHistoryChain` at the same insertion point as your `TestWriteFindingField_acceptsSuggestedRecipients`. Both are additive so it's a trivial keep-both rebase. Everything else auto-merges.

alexandre-daubois:
Rebase is done 🙂

andrew (approved):
Rebase over #669 is clean, both tests kept, builds and tests pass on current main.

PR #633, opened 2026-07-13
feat(scans): scheduled recurring scans with new-commit skip

### LINKED ISSUE #145: Scheduled and repeating scans for repositories
Add the ability to schedule recurring scans for a repository. The scan should check whether the repo has new commits since the last scan and skip if nothing changed.

Rough scope:

- Per-repo schedule (e.g. daily, weekly, or cron expression) configurable from the repo settings page
- Before enqueuing, compare remote HEAD against the `commit` column on the most recent completed scan -- skip if unchanged
- Log the skip so it's visible in the scans list (something like a `skipped` status with a reason)
- Optionally a global default schedule that applies to all repos unless overridden

alexandre-daubois:
Fix #145

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -108,6 +108,19 @@ type Repository struct {
 	// scrutineer.scan_config in every skill workspace.
 	ScanConfig string `gorm:"type:text"`
 
+	// ScanSchedule drives recurring scans: "daily", "weekly", or a
+	// cron expression. Empty inherits the global scan_schedule setting;
+	// "off" disables scheduling even when a global default is set.
+	ScanSchedule string
+	// UpstreamURL, when set, names the upstream this repository is a
+	// pushed staging copy of (no forge fork relationship). The scheduler
+	// fast-forwards the repository from it before the new-commit check.
+	UpstreamURL string
+	// NextScheduledScanAt is scheduler bookkeeping: when the next
+	// scheduled run is due. Null means "recompute on the next tick";
+	// schedule edits reset it rather than computing inline.
+	NextScheduledScanAt *time.Time `gorm:"index"`
+
 	CreatedAt time.Time
 	UpdatedAt time.Time
 
@@ -134,11 +147,16 @@ func (r Repository) LocalPath() string {
 type ScanStatus string
 
 const (
-	ScanQueued    ScanStatus = "queued"
-	ScanRunning   ScanStatus = "running"
-	ScanPaused    ScanStatus = "paused"
-	ScanDone      ScanStatus = "done"
-	ScanFailed    ScanStatus = "failed"
+	ScanQueued  ScanStatus = "queued"
+	ScanRunning ScanStatus = "running"
+	ScanPaused  ScanStatus = "paused"
+	ScanDone    ScanStatus = "done"
+	ScanFailed  ScanStatus = "failed"
+	// ScanSkipped records a scheduled scan the scheduler decided not to run
+	// (remote HEAD unchanged, a scan already in flight, ...). The row never
+	// enters the queue; Error carries the reason so the scans list shows
+	// why nothing ran.
+	ScanSkipped   ScanStatus = "skipped"
 	ScanCancelled ScanStatus = "cancelled"
 )
 
@@ -1092,7 +1110,7 @@ func isEmptyJSONValue(v any) bool {
 }
 
 func (s ScanStatus) Terminal() bool {
-	return s == ScanDone || s == ScanFailed || s == ScanCancelled
+	return s == ScanDone || s == ScanFailed || s == ScanCancelled || s == ScanSkipped
 }
 
 const (
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -114,7 +114,8 @@ type Repository struct {
 	ScanSchedule string
 	// UpstreamURL, when set, names the upstream this repository is a
 	// pushed staging copy of (no forge fork relationship). The scheduler
-	// fast-forwards the repository from it before the new-commit check.
+	// force-syncs the repository from it (a mirror push that overwrites
+	// local-only commits) before the new-commit check.
 	UpstreamURL string
 	// NextScheduledScanAt is scheduler bookkeeping: when the next
 	// scheduled run is due. Null means "recompute on the next tick";
--- a/internal/web/schedule.go
+++ b/internal/web/schedule.go
@@ -1,0 +1,165 @@
+package web
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/robfig/cron/v3"
+
+	"scrutineer/internal/db"
+)
+
+// ScheduleOff disables scheduled scans for a repository even when a global
+// default schedule is configured. An empty repo schedule inherits the global
+// setting instead.
+const ScheduleOff = "off"
+
+// scheduleKind marks the scans-list rows the scheduler writes when it decides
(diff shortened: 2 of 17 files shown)


internal/web/schedule.go
```diff
@@ -0,0 +1,165 @@
...
+
+// runScheduledScan performs one scheduled firing for repo: sync from the
+// upstream when one is configured, then compare the remote HEAD against the
+// most recent completed scan's commit and either record a skip or enqueue a
+// diff-rescan group (which falls back to full coverage when no baseline
+// exists, e.g. on a never-scanned repository).
+func (s *Server) runScheduledScan(ctx context.Context, repo db.Repository) {
+	var inflight int64
+	s.DB.Model(&db.Scan{}).
+		Where("repository_id = ? AND status IN ?", repo.ID, []db.ScanStatus{db.ScanQueued, db.ScanRunning}).
```
andrew: `ScanPaused` isn't in the list. With the auto-pause-on-rate-limit work (#551, #512) a paused scan is very much still in flight, and the scheduler will enqueue a fresh diff-rescan trio on top of it. Either add `db.ScanPaused` here or flip to a `NOT terminal` check.

internal/worker/upstream.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	branch, err := gitWithEnv(ctx, dst, env, "symbolic-ref", "--short", "HEAD")
+	if err != nil {
+		return fmt.Errorf("resolve default branch: %s: %w", strings.TrimSpace(branch), err)
+	}
+	branch = strings.TrimSpace(branch)
+	if out, err := gitWithEnv(ctx, dst, env, "fetch", "--quiet", "--", upstreamURL, "HEAD"); err != nil {
+		return fmt.Errorf("fetch upstream: %s: %w", strings.TrimSpace(out), err)
+	}
+	if out, err := gitWithEnv(ctx, dst, env, "push", "--quiet", "--force", "origin", "FETCH_HEAD:refs/heads/"+branch); err != nil {
```
andrew: This is `push --force`, and `TestSyncUpstream_overwritesRewrittenHistory` proves that's deliberate. But `docs/database.md`, the `Repository.UpstreamURL` struct comment, and the `repo_show.html` help text all say "fast-forwards". Anyone reading those and setting an upstream on a repo with local commits (the `fork` skill's staging repos carry a metadata directory on top of upstream, for example) will lose them silently.

Either reword to "mirrors" / "force-syncs" in all four places, or drop `--force` and let a non-ff surface as a skip reason. I'd lean toward the wording change since the mirror semantics seem intentional.

andrew (changes requested):
Solid implementation with good test coverage. The null-`NextScheduledScanAt`-means-recompute pattern is clean and the `repoDiffScan` -> `enqueueDiffRescanGroup` extraction reads well. Two things to fix before merge (inline), plus follow-ups below.

**Follow-up, not blocking**

- `SyncUpstream` does a full `--bare` clone into a tempdir each time upstream moves. The worker already keeps a persistent clone cache (`disk_bytes` etc.); reusing it would avoid re-cloning large repos daily.
- `scheduleTick` runs each due repo's `ls-remote` (and possibly clone+fetch+push) serially inside the tick. A global `daily` default fires every repo at `@daily` = midnight local, so the first post-midnight tick walks them all one by one. Fine at current scale; will bite later.
- Skip rows accumulate without bound. An idle repo on `@hourly` writes 24 scan rows a day forever. Might want a retention sweep, or collapse consecutive identical-reason skips.
- `robfig/cron/v3` v3.0.1 is from 2019 and effectively unmaintained. It works and half the Go ecosystem uses it, just flagging.
- The `gorm:"index"` on `NextScheduledScanAt` is never queried; `scheduleTick` loads all repos unconditionally.

**Nits**

- `recordScheduledSkip` sets `FinishedAt` but not `StartedAt`; some views compute duration from the pair.
- `TestScheduleTick_notDueYet` uses `time.Time.Equal` on a value round-tripped through SQLite; if that ever gets timezone/precision-normalised the test will flake. `WithinDuration` would be safer.

alexandre-daubois:
Handled the three follow-ups: clone-cache reuse for upstream sync, collapsed consecutive skip rows, and an index-filtered scheduler tick. Also rebased the PR

internal/web/server.go
```diff
@@ -2300,33 +2346,48 @@ func (s *Server) repoScan(w http.ResponseWriter, r *http.Request) {
...
+// skills are tolerated; a missing deep-dive is errDeepDiveMissing. Shared by
+// the "Diff rescan" button and the scheduler.
+func (s *Server) enqueueDiffRescanGroup(ctx context.Context, repoID uint, model, subPath string) (int, error) {
 	group := uuid.NewString()
-	model := r.FormValue("model")
-	subPath := strings.TrimSpace(r.FormValue("sub_path"))
-	names := []string{reconSkillName, threatModelSkillName, "semgrep", deepDiveSkillName}
 	queued := 0
-	for _, name := range names {
+	for _, name := range []string{reconSkillName, threatModelSkillName, "semgrep", deepDiveSkillName} {
```
andrew, line 2377: Check the required deep-dive skill before enqueueing any auxiliaries. With the current order, if recon/threat-model/semgrep are active but security-deep-dive is missing or disabled, those scans are already queued when this returns `errDeepDiveMissing`. The scheduler then also writes a skipped row, so the run is simultaneously reported as skipped and partially running. Please preflight the required skill (or make group creation atomic) before the first enqueue.

internal/web/schedule.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	head, err := s.resolveRemoteHead(ctx, repo)
+	if err != nil {
+		s.recordScheduledSkip(repo, "remote HEAD lookup failed: "+err.Error())
+		return
+	}
+	var last db.Scan
+	found := s.DB.Select("id, `commit`").
+		Where("repository_id = ? AND status = ? AND `commit` <> ''", repo.ID, db.ScanDone).
+		Order("id desc").First(&last).Error == nil
```
andrew: Handle database errors separately from “no completed scan.” This boolean treats every `First` error as no baseline, so a lock or connection failure falls through to `enqueueDiffRescanGroup` and may trigger a full rescan. Please allow `gorm.ErrRecordNotFound` as the no-history case, but log and return on other errors.

internal/db/db.go
```diff
@@ -134,11 +148,16 @@ func (r Repository) LocalPath() string {
...
+	ScanQueued  ScanStatus = "queued"
+	ScanRunning ScanStatus = "running"
+	ScanPaused  ScanStatus = "paused"
+	ScanDone    ScanStatus = "done"
+	ScanFailed  ScanStatus = "failed"
+	// ScanSkipped records a scheduled scan the scheduler decided not to run
+	// (remote HEAD unchanged, a scan already in flight, ...). The row never
+	// enters the queue; Error carries the reason so the scans list shows
+	// why nothing ran.
+	ScanSkipped   ScanStatus = "skipped"
```
andrew, line 160: Thread the new status through the OpenAPI contract too. `components.schemas.Scan` and `ScanExport` still enumerate only queued/running/paused/done/failed/cancelled, and the scan status filters omit `skipped` as well. Scheduler rows are returned by those endpoints, so generated clients cannot represent or filter them until the enums include `skipped`.

andrew (changes requested):
Three issues remain in the scheduler enqueue/error paths and API contract; details inline.

andrew (approved):

PR #653, opened 2026-07-16
Fan out security deep dives by focus area

### LINKED ISSUE #108: Recon pass that emits focus areas, not just sub-paths

abhinavgautam01:
Fixes #108

## Summary

Completes the focus-area execution path by turning a completed root `threat-model` scan into one scoped `security-deep-dive` scan per configured focus area.

Each deep-dive receives a stable snapshot of its assigned area and only sees that area’s selected source paths.

## Changes

- Add `Scan.FocusArea` to persist a normalized focus-area snapshot per scan.
- Fan out completed root `threat-model` scans into parallel `security-deep-dive` scans.
- Preserve a single unscoped deep dive when no focus areas are configured.
- Filter the staged source tree to each child scan’s configured path globs.
- Stage `scrutineer.focus_area` in `context.json` for scoped deep-dive runs.
- Keep focus area scope across scan retries.
- Require diff rescans to use a baseline with the same focus-area snapshot.
- Update triage and diff-rescan orchestration to wait for threat-model fan-out instead of enqueueing an early unscoped deep dive.
- Document the scoped audit behavior.

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`


internal/web/focus_area_deep_dive.go
```diff
@@ -0,0 +1,91 @@
...
+
+// autoEnqueueFocusAreaDeepDives turns a completed root threat-model into one
+// security-deep-dive per configured input-processing subsystem. The complete,
+// normalized focus area is snapshotted on every child scan so later edits to
+// scan_config cannot change already queued work.
+func (s *Server) autoEnqueueFocusAreaDeepDives(scan *db.Scan) {
+	if scan == nil || scan.Status != db.ScanDone || scan.SkillName != threatModelSkillName ||
+		scan.SubPath != "" || scan.Ref != "" {
+		return
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: Removed the root-only guard so completed threat-model scans now fan out for subpath- and ref-scoped runs as well.

internal/web/focus_area_deep_dive.go
```diff
@@ -0,0 +1,91 @@
...
+	if _, err := s.enqueueSkillWith(context.Background(), parent.RepositoryID, skillID, ScanOpts{
+		Model:      parent.Model,
+		Effort:     parent.Effort,
+		Profile:    parent.Profile,
+		RescanMode: parent.RescanMode,
+		ScanGroup:  group,
+		FocusArea:  focusArea,
+	}); err != nil {
```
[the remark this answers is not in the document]
abhinavgautam01: Focus-area deep-dive children now inherit `SubPath` and `Ref` from the completed threat-model scan, so they run against the same code scope.

internal/web/focus_area_deep_dive.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func (s *Server) enqueueFocusAreaDeepDive(parent *db.Scan, skillID uint, group, focusArea string) {
+	var existing int64
+	if err := s.DB.Model(&db.Scan{}).
+		Where("repository_id = ? AND skill_id = ? AND scan_group = ? AND focus_area = ? AND status IN ?",
+			parent.RepositoryID, skillID, group, focusArea,
+			[]db.ScanStatus{db.ScanQueued, db.ScanRunning, db.ScanDone}).
+		Count(&existing).Error; err != nil {
```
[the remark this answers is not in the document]
abhinavgautam01: The idempotency lookup now includes the parent scan’s `sub_path` and `ref`, in addition to repository, skill, group and focus-area snapshot.

internal/web/focus_area_deep_dive_test.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	s.DB.Create(&parent)
+
+	s.onScanFinalized(&parent)
+	var deepDives []db.Scan
+	s.DB.Where("repository_id = ? AND skill_name = ?", repo.ID, deepDiveSkillName).Find(&deepDives)
+	if len(deepDives) != 1 || !strings.Contains(deepDives[0].FocusArea, `"name":"parser"`) {
+		t.Fatalf("deep dives = %+v", deepDives)
+	}
+}
```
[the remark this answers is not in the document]
abhinavgautam01: Added a scoped fan-out regression test. It verifies a completed subpath/ref threat-model queues a child with the same scope, and that an existing same-group deep dive for another subpath/ref does not suppress it.

internal/db/db.go
```diff
@@ -225,6 +225,12 @@ type Scan struct {
 	// scans that were not launched as part of such a batch.
 	ScanGroup string `gorm:"index"`
 
+	// FocusArea is the normalized JSON focus-area payload for an audit split
+	// out of a repository scan configuration. Empty means the scan covers its
+	// normal repository or subproject scope. It is stored on the scan so queued
+	// work remains reproducible if the repository configuration changes.
+	FocusArea string `gorm:"type:text"`
```
alexandre-daubois, line 251: `docs/database.md` isn't updated for the new `scans.focus_area` column. Every new `Scan` column must be added to the `## scans` table in the same PR. Please add a `focus_area` row describing the normalized JSON snapshot and that empty means unscoped.

internal/web/focus_area_deep_dive.go
```diff
@@ -0,0 +1,92 @@
...
+	if group == "" {
+		// A manually launched threat-model has no group. A stable derived
+		// group makes the completion hook idempotent if it is called again.
+		group = fmt.Sprintf("focus-%d", scan.ID)
+	}
+	if len(config.FocusAreas) == 0 {
+		// Preserve normal coverage until a repository has a useful partition.
+		s.enqueueFocusAreaDeepDive(scan, skill.ID, group, "")
+		return
+	}
```
alexandre-daubois, line 57: The zero-focus-areas fallback in `autoEnqueueFocusAreaDeepDives` (the single unscoped deep-dive) has no test, yet it's now the only path that enqueues a deep-dive for repos without configured focus areas since `security-deep-dive` was dropped from the other enqueue paths. Please add a test asserting exactly one deep-dive with empty `FocusArea` is queued, and that a second call stays idempotent.

alexandre-daubois (changes requested):

internal/db/db.go
abhinavgautam01: Added the scans.focus_area entry to docs/database.md, documenting that it stores the normalized JSON focus-area snapshot used by split deep-dives and that an empty value represents normal unscoped repository or subproject coverage.

internal/web/focus_area_deep_dive.go
abhinavgautam01: Added TestAutoEnqueueFocusAreaDeepDivesFallsBackToUnscoped, which uses a repository with no configured focus areas, invokes the completion hook twice and asserts that exactly one deep-dive is queued with an empty FocusArea. It also checks that the parent scan group is preserved.

internal/web/focus_area_deep_dive.go
```diff
@@ -0,0 +1,92 @@
...
+	"scrutineer/internal/db"
+	"scrutineer/internal/repoconfig"
+)
+
+// autoEnqueueFocusAreaDeepDives turns a completed threat-model into one
+// security-deep-dive per configured input-processing subsystem. The complete,
+// normalized focus area is snapshotted on every child scan so later edits to
+// scan_config cannot change already queued work.
+func (s *Server) autoEnqueueFocusAreaDeepDives(scan *db.Scan) {
+	if scan == nil || scan.Status != db.ScanDone || scan.SkillName != threatModelSkillName {
```
andrew: This gates on `Status == ScanDone`, and `maybeFireScanFinalized` only invokes the hook on `ScanDone` or a fail-on-threshold (which is `ScanFailed`, so it early-returns here anyway). Combined with the removal of `security-deep-dive` from triage and `repoDiffScan`, a threat-model runner error or timeout now means no deep-dive runs at all for that batch. Previously deep-dive was enqueued in parallel and ran regardless.

Either enqueue an unscoped deep-dive when threat-model finalizes without `ScanDone` (so coverage falls back to today's behaviour), or document threat-model as a hard prerequisite and confirm that retrying it re-triggers fan-out.

internal/web/focus_area_deep_dive.go
```diff
@@ -0,0 +1,92 @@
...
+	if err := s.DB.Model(&db.Scan{}).
+		Where("repository_id = ? AND skill_id = ? AND scan_group = ? AND sub_path = ? AND ref = ? AND focus_area = ? AND status IN ?",
+			parent.RepositoryID, skillID, group, parent.SubPath, parent.Ref, focusArea,
+			[]db.ScanStatus{db.ScanQueued, db.ScanRunning, db.ScanDone}).
+		Count(&existing).Error; err != nil {
+		s.Log.Warn("focus-area deep dives: check existing", "scan", parent.ID, "err", err)
+		return
+	}
+	if existing > 0 {
+		return
```
andrew, line 82: `Count(...)` then `enqueueSkillWith(...)` isn't serialized, so two concurrent deliveries of the same threat-model completion could both pass the count and double-enqueue. #667 (merged) added `s.agentEnqueueMu` on `Server` for the same check-then-insert pattern in the scan-token API; taking it here would close the window.

andrew (changes requested):
The snapshot-on-scan approach and the seed-then-fan-out ordering are right, and the retry/diff-baseline propagation is thorough. Two things to address before this lands.

internal/web/focus_area_deep_dive.go
abhinavgautam01: Added a post-persistence `OnScanFailed` worker hook and wired it to focus-area fan-out. A failed or timed-out threat-model now queues one unscoped deep-dive fallback, preserving the parent scan group, subpath and ref. Successful threat-model scans still seed `scan_config` before normal focus-area fan-out. Added an idempotence test for the failed scoped path.

internal/web/focus_area_deep_dive.go
abhinavgautam01: Wrapped the existing-scan check and enqueue operation in `agentEnqueueMu`, reusing the server-process mutex introduced by #667. Concurrent completion deliveries can no longer both pass the count check and enqueue duplicate focus-area deep-dives.

andrew (approved):

PR #654, opened 2026-07-16
fix: handle repository lookup errors in OSV exports

### LINKED ISSUE #597: Handle Repository Lookup Errors in OSV Finding Export
### Summary

`internal/web/finding_osv.go` ignores the database error when loading the repository for a finding before building the OSV JSON export.

### Details

In `findingOSV`, the finding itself is loaded with error handling, but the repository lookup is not:

```go
var repo db.Repository
s.DB.First(&repo, f.RepositoryID)
```

If that query fails because the repository row is missing or the database returns an operational error, `repo` remains a zero-value `db.Repository`. The handler then continues building and validating the OSV document and can return `200 OK` with incomplete or misleading repository context.

### Impact

Clients downloading the OSV export may receive a syntactically valid JSON document that silently omits or corrupts repository-specific data. That is worse than a clear failure because downstream tooling may ingest and trust the bad export.

### Suggested Fix

Check the repository query error explicitly:

- Return `404 Not Found` if the repository row is missing.
- Return `500 Internal Server Error` for other database errors.
- Add a regression test proving that a missing repository does not return a successful OSV document.

ychampion:
## Summary

- stop OSV exports when their repository lookup fails
- return 404 for a missing repository and 500 for operational database errors
- cover both failure paths with handler regressions

## Why

The OSV handler ignored repository lookup errors and continued with a zero-value repository, allowing incomplete security data to be returned as a successful export.

Closes #597

## Validation

- `go test -race ./...` — passed
- `go vet ./...` — passed
- `go vet -tags podman ./...` — passed
- `golangci-lint run ./...` — passed
- `govulncheck ./...` — passed
- `go mod tidy -diff` — passed

### DIFF
--- a/internal/web/finding_osv.go
+++ b/internal/web/finding_osv.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"embed"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"regexp"
@@ -15,6 +16,7 @@ import (
 	"github.com/git-pkgs/purl"
 	"github.com/git-pkgs/vulns"
 	"github.com/santhosh-tekuri/jsonschema/v6"
+	"gorm.io/gorm"
 
 	"scrutineer/internal/db"
 )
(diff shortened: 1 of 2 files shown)


andrew (approved):
Thanks!

PR #658, opened 2026-07-16
Postgres support

### LINKED ISSUE #636: Postgres support: remaining blockers
Postgres support is on the near-term roadmap. #635 removed the hardcoded backtick quoting reported in #629; these are the remaining blockers found in a sweep of `internal/`.

- [ ] Rename reserved-word columns so quoting isn't needed at all: `Scan.Commit` → `commit_sha`, `FindingNote.By` → `actor` (or similar). Schema migration.
- [ ] `db.Open`: DSN-based driver selection (`gorm.io/driver/postgres`), skip the `_pragma=foreign_keys(1)&_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)` string and the `PRAGMA busy_timeout` in the backup path when not on SQLite.
- [ ] Backup: `VACUUM INTO` is SQLite-only; needs a `pg_dump` path.
- [ ] Queue: `maragu.dev/goqite` is SQLite-backed; check its Postgres story or swap.
- [ ] `LIKE` in search UIs (18 sites across `internal/web` and `internal/worker`): case-insensitive on SQLite, case-sensitive on Postgres, so search would silently narrow. `LOWER(col) LIKE LOWER(?)` is portable; `ILIKE` is Postgres-only.
- [ ] CI job against real Postgres so regressions surface; the DryRun guard in #635 only proves the dialector is doing the quoting, not that the SQL runs.

cc @walterhpearce who I believe has been looking at this.

walterhpearce:
Resolves #636

This is a first pass at postgres support. I'm submitted a PR for first pass review and feedback, it is not fully complete but is functional. A migration path for importing an existing sqlite -> postgres exists, and we are running it on our production instance.

I've copied the issues initial checklist here:

Completed:
- [x] `db.Open`: DSN-based driver selection (`gorm.io/driver/postgres`), skip the `_pragma=foreign_keys(1)&_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)` string and the `PRAGMA busy_timeout` in the backup path when not on SQLite.
- [x] Queue: `maragu.dev/goqite` is SQLite-backed; check its Postgres story or swap.
- [x] Migration: A migration path is generated in scripts and documented.
- [x] Documentation: Updated across the project for postgres vs. sqlite 

Incomplete:
- [ ] Rename reserved-word columns: This was not done in this branch yet - do we still want it done here? It seems a orthogonal issue and requires migration routes for all existing implementations, while this allows just a drop in postgres exchange with no migrations for the sqlite version.
- [ ] Backup: I have not addressed this at all yet.
- [ ] Search case sensitivity: not addressed yet.
- [ ] CI job against real Postgres so regressions surface; the DryRun guard in #635 only proves the dialector is doing the quoting, not that the SQL runs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 /data/
 /data-*/
 /scrutineer
+# Compiled helper binaries (build from source under scripts/ instead).
+/migrate-sqlite-to-postgres
 *.db
 *.db-shm
 *.db-wal
--- a/cmd/scrutineer/backup.go
+++ b/cmd/scrutineer/backup.go
@@ -65,6 +65,9 @@ func runBackup(args []string, out io.Writer) error {
 	if err != nil {
 		return err
 	}
+	if err := errPostgresManaged(cfg); err != nil {
+		return err
+	}
 	dbPath := filepath.Join(resolveDataDir(cfg, dataDir), dbFileName)
 	if _, err := os.Stat(dbPath); err != nil {
 		return fmt.Errorf("no database at %s: %w", dbPath, err)
@@ -111,6 +114,9 @@ func runRestore(args []string, out io.Writer) error {
 
 	cfg, err := config.Load(configPath)
 	if err != nil {
+		return err
+	}
+	if err := errPostgresManaged(cfg); err != nil {
 		return err
 	}
 	if ok, err := isSQLiteFile(from); err != nil {
@@ -212,6 +218,18 @@ func serverRunning(addr string) bool {
 	}
 	_ = conn.Close()
 	return true
+}
+
+// errPostgresManaged returns a non-nil error when the config selects the
+// postgres backend. The backup/restore subcommands are SQLite-only (they use
+// VACUUM INTO and file swaps); PostgreSQL deployments back up through the
+// operator's own tooling (pg_dump/pg_restore, managed snapshots), so scrutineer
+// declines rather than pretending to.
+func errPostgresManaged(cfg *config.Config) error {
+	if cfg != nil && cfg.Database.Driver == "postgres" {
+		return errors.New("backup/restore is SQLite-only; PostgreSQL backups are operator-managed (use pg_dump/pg_restore or your provider's snapshots)")
+	}
+	return nil
 }
 
 // resolveDataDir mirrors the server's precedence: an explicit flag wins over
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -118,9 +118,16 @@ type flags struct {
 	recipientsFile        string
 	identityFile          string
 	autoRejectMissedCount int
-	federationSalt        string
-	federationContact     string
-	skillLocal            skillDirs
+	// dbDriver selects the database backend ("" or "sqlite" for the embedded
+	// file, "postgres" for an external server). dbDSN is the connection
+	// string for postgres; sqlite ignores it and uses dataDir/scrutineer.db.
+	// These are config-only (no CLI flag) — a DSN on the command line would
+	// land in shell history and process listings.
+	dbDriver          string
+	dbDSN             string
+	federationSalt    string
+	federationContact string
+	skillLocal        skillDirs
 
 	// set records which flags were passed on the command line so merge
 	// knows not to let the config file override them.
@@ -272,6 +279,10 @@ func (f *flags) merge(cfg *config.Config) {
 	if cfg.AutoRejectMissedCount > 0 && !f.set["auto-reject-missed-count"] {
 		f.autoRejectMissedCount = cfg.AutoRejectMissedCount
 	}
+	// Database backend is config-only, so no f.set guard.
+	f.dbDriver = cfg.Database.Driver
+	f.dbDSN = cfg.Database.DSN
+
 	if cfg.FederationSalt != "" {
 		f.federationSalt = cfg.FederationSalt
 	}
@@ -305,6 +316,26 @@ func (f *flags) merge(cfg *config.Config) {
 	if cfg.Theme != "" {
 		web.SetTheme(cfg.Theme)
 	}
+}
+
+// databaseOptions maps the merged config onto db.OpenBackend's Options. For
+// postgres the DSN comes from config verbatim; for sqlite (the default) the
+// DSN is the database file inside the data directory, matching the historical
+// db.Open(dataDir/scrutineer.db) call.
+func (f *flags) databaseOptions() db.Options {
+	if f.dbDriver == "postgres" {
+		return db.Options{Dialect: db.DialectPostgres, DSN: f.dbDSN}
+	}
+	return db.Options{Dialect: db.DialectSQLite, DSN: filepath.Join(f.dataDir, dbFileName)}
+}
+
+// queueDialect selects the goqite backend to match databaseOptions so the
+// queue table lives in the same database as everything else.
+func (f *flags) queueDialect() queue.Dialect {
+	if f.dbDriver == "postgres" {
+		return queue.Postgres
+	}
+	return queue.SQLite
 }
 
 func (f *flags) fullClone() bool { return f.cloneMode == "full" }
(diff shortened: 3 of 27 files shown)


andrew (changes requested):
I found the following issues that were not covered by the existing Copilot review:

1. **[P1] The SQLite-to-PostgreSQL migration strands queued scans.** The migrator copies `scans` rows, including rows whose status is `queued`, but explicitly skips the `goqite` table as transient ([main.go:279-284](https://github.com/alpha-omega-security/scrutineer/blob/d1fec6ee13a6ce9714a81410c47eef1003650915/scripts/migrate-sqlite-to-postgres/main.go#L279-L284)). Nothing reconstructs queue messages after migration. In a live migration test, the destination contained one queued scan and zero queue messages, so that scan could never run. The migrator should either reject a source with pending work or rebuild/copy the queue transactionally.

2. **[P1] The default Compose configuration exposes PostgreSQL with known credentials.** [docker-compose.yml:13-20](https://github.com/alpha-omega-security/scrutineer/blob/d1fec6ee13a6ce9714a81410c47eef1003650915/docker-compose.yml#L13-L20) publishes `5432:5432` on every host interface while setting the password to `scrutineer`. Anyone who can reach the host port can authenticate to a database containing sensitive scan data. Please bind the port to loopback, omit the host port, or require externally supplied credentials. This is distinct from Copilot's comment about the inaccurate Compose header.

3. **[P2] Repository sorting generates invalid PostgreSQL SQL.** [internal/web/scans.go:40](https://github.com/alpha-omega-security/scrutineer/blob/d1fec6ee13a6ce9714a81410c47eef1003650915/internal/web/scans.go#L40) orders by `` `Repository`.name ``, which PostgreSQL rejects with `SQLSTATE 42601`. This should use a dialect-aware GORM column/expression rather than a raw backtick-quoted identifier.

4. **[P2] Concurrent first startup is not safe.** The queue runs extension and trigger DDL from [schema_postgres.sql](https://github.com/alpha-omega-security/scrutineer/blob/d1fec6ee13a6ce9714a81410c47eef1003650915/internal/queue/schema_postgres.sql#L5) during every `queue.New`. Running 16 initializers concurrently against a fresh PostgreSQL database failed on the `pg_extension_name_index` unique constraint. Schema installation needs to be serialized, for example with a migration/advisory lock, before the documented multi-instance configuration is safe. This is separate from Copilot's `EXECUTE PROCEDURE` syntax comment.

5. **[P2] The PR branch retains PostgreSQL failures that have already been fixed on `main`.** Live testing the PR head found invalid backtick SQL in [audit.go:170](https://github.com/alpha-omega-security/scrutineer/blob/d1fec6ee13a6ce9714a81410c47eef1003650915/internal/db/audit.go#L170) and [diff_rescan.go:154](https://github.com/alpha-omega-security/scrutineer/blob/d1fec6ee13a6ce9714a81410c47eef1003650915/internal/worker/diff_rescan.go#L154). Those paths fail on PostgreSQL. PR #635 fixed them on `main`, so the currently conflicting branch needs to be rebased while preserving those fixes.

6. **[P2] A generated, platform-specific binary is committed at the repository root.** [`migrate-sqlite-to-postgres`](https://github.com/alpha-omega-security/scrutineer/blob/d1fec6ee13a6ce9714a81410c47eef1003650915/migrate-sqlite-to-postgres) is a 24.6 MB unstripped Linux x86-64 executable whose Go build metadata records a dirty source tree. It cannot run on other supported host platforms and permanently bloats repository history. Please remove it and build the migration command from source when needed.

The migration, repository-sort, and concurrent-startup results above were reproduced against PostgreSQL 16.

andrew (changes requested):
Thanks for getting this running in production; the core db/queue/config approach looks like the right direction. Before I can do a proper code review of those pieces, a few structural things need sorting:

1. `migrate-sqlite-to-postgres` at the repo root is a 24.6 MB compiled binary checked into git. It needs removing from history on this branch and `.gitignore`ing; the source is already at `scripts/migrate-sqlite-to-postgres/main.go`.
2. `docker/postgresdb/initdb/10-scrutineer-schema.sql` (1952 lines, 24 tables) is already stale: #665 added `audit_events` today and it's not in there. A hand-maintained dump will drift on every model change. Either drop it and let `AutoMigrate` build the schema on first connect, or generate it from the models at build time.
3. `go.sum` is incomplete (3 new lines for 6 new modules) and the branch is behind main on `x/crypto`/`x/term`. Needs a rebase onto main + `go mod tidy`.
4. Unrelated changes are bundled in and should be their own PRs so postgres can be reviewed on its own:
   - `scansCancelQueued` + the "Cancel queued" button in `templates/jobs.html`
   - the Dockerfile relocation (`Dockerfile` → `docker/cmd/`, `Dockerfile.runner` → `docker/runner/`, every `docker/profiles/*` path update, `runner-image.yml`, README)
5. CI hasn't run on this branch (`gh pr checks` reports nothing), likely fallout from the workflow path changes. Needs to be green before review.
6. On rebase: #669 (merged today) added `isSQLiteBusy` in `internal/db/finding_helpers.go` for CAS retry; that'll want a postgres SQLSTATE `40001` branch.

The pgx dependency stack is fine (pure Go, no cgo). The `theme.go` dialector branch is legitimate. Once the above is scoped down I'll go through the db/queue/config/migration code properly.

andrew (changes requested):
Thanks — the initdb dump is gone, go.mod/go.sum is tidy after the merge, CI is green, and `isPostgresSerializationFailure` (40001/40P01, structural `SQLState()` match) wired into the CAS retry is exactly right.

Two of the earlier points are still open and both block the db/queue/config review:

**Binary still in branch history.** 2ea9b73 removes it at tip and `.gitignore`s it, but the 23.5 MB blob is still reachable at 855c2ac and d1fec6e. This repo allows merge-commits and rebase-merges, and either would land the blob in main permanently. It needs dropping via an interactive rebase (drop the file from those two commits, force-push the branch).

**Unrelated changes still bundled.** `scansCancelQueued` + `templates/jobs.html` + its test are still here, and the whole Dockerfile relocation is still here (`Dockerfile`→`docker/cmd/`, `Dockerfile.runner`→`docker/runner/`, every `docker/profiles/*`, `runner-image.yml`, `docs/egress-sidecar.md`, `container.go`, `worker_test.go`, `preflight_test.go`, `podman_integration_test.go`, `check-claude-version-pins.sh`, README). Each of those is a reasonable change on its own; bundled in they make the postgres diff about twice the size it needs to be. Please split them into their own PRs — the interactive rebase you need for the binary is the natural place to peel them off.

One new thing: there are now two near-identical compose files, root `docker-compose.yml` and `docker/postgresdb/docker-compose.yaml`, differing only by a `name:` field and the header invocation. A root `docker-compose.yml` reads as "this runs scrutineer", which it doesn't — keep the one under `docker/postgresdb/` and drop the root copy.

Once the diff is postgres-only I'll go through `internal/db`, `internal/queue`, the config wiring, and the migrator properly.

andrew (changes requested):
The PostgreSQL path still has several correctness issues:

- The SQLite migrator rejects every current database because its copied model list omits `audit_events`, `package_alternatives`, `advisory_audits`, `conversations`, and `chat_messages`. Add copiers for every AutoMigrate model and a test that keeps the lists synchronized.
- `copyRaw` cannot migrate populated join tables. `FindInBatches` with `[]map[string]any` has no primary key and fails with `model value required`. Use typed join rows with deterministic batching.
- Concurrent first startup races in PostgreSQL `AutoMigrate`, producing duplicate catalog keys and `relation already exists`. Serialize the complete model migration with a PostgreSQL advisory lock and test concurrent startup against a fresh database.
- `-force` bypasses the queued/running scan guard even though queue data is not migrated. Keep that guard unconditional or require a separate explicit data-loss option.
- User-facing searches use `LIKE`, which is case-sensitive on PostgreSQL. Add dialect-aware case-insensitive matching and cover values such as `Homebrew`.
- The root and nested PostgreSQL compose files still duplicate one another. Keep the nested database example and remove the root copy.

The PostgreSQL integration tests currently skip unless `SCRUTINEER_TEST_PG_DSN` is set. Add a PostgreSQL CI service so these failures are exercised.

abhinavgautam01:
minor typo in the PR description: _Resolves #6368_ → Resolves #636

andrew:
@walterhpearce checking in — the last review round (2026-07-29) has six open items and the branch is now well behind main. Are you still working on this, or should we pick it up from where it is?

PR #664, opened 2026-07-18
Audit fail-open security input omissions

### LINKED ISSUE #660: Audit missing and empty security inputs for fail-open validation bypasses
## Summary

Teach `security-deep-dive` to test the absence, null value, or empty value of inputs that gate a security decision.

The current deep-dive is intentionally sink-first and traces values that reach dangerous operations. That can miss a different shape of vulnerability: the unsafe behavior happens because a value is *not present*, so validation is skipped while the protected operation continues.

Example:

```go
if sessionCookie != "" {
    if err := validateSession(sessionCookie); err != nil {
        return unauthorized()
    }
}
return serveAccountData()
```

An attacker does not need to taint `sessionCookie`; omitting it bypasses the check.

This is a concrete, small follow-up to the `audit-authz` direction discussed in #106. It should improve the existing skill rather than introduce a new skill or orchestration stage.

## Scope

Update `skills/security-deep-dive/SKILL.md` with a narrowly scoped omission check:

- When reviewing authentication, authorization, session validation, CSRF, rate-limit/lockout, or credential-validation code, identify inputs whose presence controls whether a security check runs.
- For each such input, inspect the missing, null, and empty-value branches.
- Report a candidate when the check is skipped but the security-sensitive operation remains reachable.
- When two inputs participate in the same security decision, consider whether omitting one causes the other to be accepted without its normal validation.
- Treat explicit deny/error branches as fail-closed and rule them out.
- Preserve the existing threat-model rules: dev-only paths, trusted-only entry points, and preconditions that subsume the impact are not findings.

Keep this targeted to code already identified as security-decision code. Do not add a second whole-repository inventory pass.

Use the existing findings output shape. No schema, database, worker, or UI change should be required.

## Suggested wording location

The check fits in Phase 2, after the trust-boundary/control review and before reproduction. It should be phrased as a mandatory question for security-decision inputs:

> If this input is absent, null, or empty, does the control fail closed, or is the protected operation still reachable?

## Evals

Add a small fixture and scenario under `evals/` covering both sides:

1. **Should find:** an endpoint where session/token validation is conditional on an optional cookie/header, but sensitive data is returned when the value is omitted.
2. **Should not find:** a similar-looking endpoint whose missing-value branch immediately rejects the request.

The positive assertion should require evidence naming both the conditional check and the protected operation. The negative case prevents the prompt from flagging every optional security-related input.

## Acceptance criteria

- `security-deep-dive` explicitly audits absent/null/empty behavior for inputs gating security controls.
- The audit is scoped to authn/authz/session/CSRF/rate-limit/credential decisions rather than all inputs.
- A model-backed eval detects the fail-open fixture.
- The paired fail-closed fixture produces no matching finding.
- Existing schema and finding ingestion remain unchanged.

## Non-goals

- Building a general input inventory.
- Adding a new `audit-authz` skill.
- Changing finding lifecycle or automatic severity.
- Treating optional non-security inputs as vulnerabilities.

abhinavgautam01:
Closes #660

## Summary

Extends `security-deep-dive` to examine missing, null, and empty inputs that determine whether a security control runs.

This catches fail-open paths where an optional session, credential, CSRF, authorization, or rate-limit input skips validation while the protected operation remains reachable.

## Changes

- Add a targeted Phase 2 omission check for existing security-decision code.
- Cover authentication, authorization, session validation, CSRF, rate-limit/lockout and credential-validation controls.
- Require candidate findings to identify both:
  - the conditional validation check that is skipped
  - the protected operation still reachable after omission
- Treat explicit reject, error, and redirect-to-login branches as fail-closed.
- Preserve existing threat-model constraints for trusted-only, dev-only and impact-subsuming paths.
- Add an `auth-omission` eval fixture containing:
  - a fail-open optional-session `/account` endpoint
  - a paired fail-closed `/account-safe` endpoint
- Add positive and negative deep-dive eval assertions with evidence requirements.

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`

### DIFF
--- a/evals/security-deep-dive-auth-omission.yaml
+++ b/evals/security-deep-dive-auth-omission.yaml
@@ -15,4 +15,3 @@ should_not_find:
     path: app.py
     evidence_contains:
       - safe_account
-      - abort(401)
--- a/evals/security-deep-dive-auth-omission.yaml
+++ b/evals/security-deep-dive-auth-omission.yaml
@@ -1,0 +1,18 @@
+given: an optional session cookie skips validation while an account endpoint still returns protected data
+fixture: fixtures/auth-omission
+skill: security-deep-dive
+should_find:
+  - finding: session
+    cwe: CWE-306
+    path: app.py
+    evidence_contains:
+      - session_cookie
+      - serve_account_data
+    required: true
+should_not_find:
+  - finding: session
+    cwe: CWE-306
+    path: app.py
+    evidence_contains:
+      - safe_account
+      - abort(401)
--- a/internal/evals/evals_test.go
+++ b/internal/evals/evals_test.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"os"
 	"path/filepath"
-	"slices"
 	"strings"
 	"testing"
 
@@ -36,32 +35,14 @@ func TestAuthOmissionScenario(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if sc.Skill != "security-deep-dive" || sc.Fixture != "fixtures/auth-omission" {
-		t.Fatalf("scenario = %+v", sc)
+	report := `{"findings":[{"title":"Session omission bypass","severity":"High","cwe":"CWE-306","location":"app.py:18","trace":"session_cookie skips validation before serve_account_data."}]}`
+	results, err := (HeuristicJudge{}).Judge(sc, report)
+	if err != nil {
+		t.Fatal(err)
 	}
-	if len(sc.ShouldFind) != 1 || len(sc.ShouldNotFind) != 1 {
-		t.Fatalf("assertions = %+v, want one positive and one negative", sc)
+	if len(results) != 2 || !results[0].Matched || !results[1].Matched {
+		t.Fatalf("scenario results = %+v, want passing positive and negative assertions", results)
 	}
-	positive := sc.ShouldFind[0]
-	if !positive.Required || positive.CWE != "CWE-306" || positive.Path != "app.py" || !hasEvidenceTerms(positive.Evidence, "session_cookie", "serve_account_data") {
-		t.Errorf("positive assertion = %+v", positive)
-	}
-	negative := sc.ShouldNotFind[0]
-	if negative.CWE != "CWE-306" || negative.Path != "app.py" || !hasEvidenceTerms(negative.Evidence, "safe_account", "abort(401)") {
-		t.Errorf("negative assertion = %+v", negative)
-	}
-}
-
-func hasEvidenceTerms(evidence []string, terms ...string) bool {
-	if len(evidence) != len(terms) {
-		return false
-	}
-	for _, term := range terms {
-		if !slices.Contains(evidence, term) {
-			return false
-		}
-	}
-	return true
 }
 
 func TestLoadScenarioDefaultsRequiredButAllowsOptional(t *testing.T) {
--- a/internal/evals/evals_test.go
+++ b/internal/evals/evals_test.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


skills/security-deep-dive/SKILL.md
```diff
@@ -130,6 +130,29 @@ Even where the input is attacker-controlled, check whether the project has alrea
 
 Even where the input is attacker-controlled, check the precondition does not subsume the conclusion. If reaching the sink requires the attacker to already hold a capability equal to or stronger than what the sink grants — write access to a directory documented as holding executable hooks, MITM position on a connection the finding claims to let them influence — the finding is circular. The attack path's first step already arrives at its last. Write "precondition subsumes conclusion" and move to the next sink.
 
+### Security-decision omission checks
```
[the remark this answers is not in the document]
abhinavgautam01: Updated the omission guidance to Step 2a: Security-decision omission checks, making its placement between Steps 2 and 3 explicit.

internal/evals/evals_test.go
```diff
@@ -30,6 +31,27 @@ func TestLoadScenarios(t *testing.T) {
...
+	if sc.Skill != "security-deep-dive" || sc.Fixture != "fixtures/auth-omission" {
+		t.Fatalf("scenario = %+v", sc)
+	}
+	if len(sc.ShouldFind) != 1 || len(sc.ShouldNotFind) != 1 {
+		t.Fatalf("assertions = %+v, want one positive and one negative", sc)
+	}
+	positive := sc.ShouldFind[0]
+	if !positive.Required || positive.CWE != "CWE-306" || positive.Path != "app.py" || !slices.Equal(positive.Evidence, []string{"session_cookie", "serve_account_data"}) {
+		t.Errorf("positive assertion = %+v", positive)
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: Updated the positive assertion test to verify the required evidence terms by membership rather than slice order, matching the evaluator’s order-independent behavior.

internal/evals/evals_test.go
```diff
@@ -30,6 +31,27 @@ func TestLoadScenarios(t *testing.T) {
...
+	negative := sc.ShouldNotFind[0]
+	if negative.CWE != "CWE-306" || negative.Path != "app.py" || !slices.Equal(negative.Evidence, []string{"safe_account", "abort(401)"}) {
+		t.Errorf("negative assertion = %+v", negative)
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: Updated the negative assertion test the same way, so reordering YAML evidence terms does not change test behavior.

evals/security-deep-dive-auth-omission.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - session_cookie
+      - serve_account_data
+    required: true
+should_not_find:
+  - finding: session
+    cwe: CWE-306
+    path: app.py
+    evidence_contains:
+      - safe_account
+      - abort(401)
```
andrew: `abort(401)` appears in both endpoints (`/account` line 19 and `/account-safe` lines 27/29), so it doesn't distinguish a `/account-safe` false positive from a correct `/account` finding. That leaves `safe_account` as the only real discriminator, and a trace that contrasts with the safe endpoint would trip this negative. Either split the fixture into `account.py` + `account_safe.py` so `path:` disambiguates, or drop `abort(401)` here.

internal/evals/evals_test.go
```diff
@@ -30,6 +31,39 @@ func TestLoadScenarios(t *testing.T) {
...
+
+func hasEvidenceTerms(evidence []string, terms ...string) bool {
+	if len(evidence) != len(terms) {
+		return false
+	}
+	for _, term := range terms {
+		if !slices.Contains(evidence, term) {
+			return false
+		}
+	}
```
andrew, line 45: This re-asserts the YAML's field values in Go; `TestLoadScenarios` already proves the file parses and the fixture exists. If you want a scenario-specific test, drive `HeuristicJudge{}.Judge(sc, fakeReport)` and assert the positive matches and the negative doesn't, so it proves the assertions behave rather than mirroring the file.

skills/security-deep-dive/SKILL.md
```diff
@@ -130,6 +130,29 @@ Even where the input is attacker-controlled, check whether the project has alrea
 
 Even where the input is attacker-controlled, check the precondition does not subsume the conclusion. If reaching the sink requires the attacker to already hold a capability equal to or stronger than what the sink grants — write access to a directory documented as holding executable hooks, MITM position on a connection the finding claims to let them influence — the finding is circular. The attack path's first step already arrives at its last. Write "precondition subsumes conclusion" and move to the next sink.
 
+### Step 2a: Security-decision omission checks
```
andrew: Wedging `Step 2a` between the existing numbered steps reads as a workaround for not renumbering. This content fits naturally as a trailing paragraph under Step 2 (where the trust-boundary and precondition-subsumes rules already live) without a new `###` heading.

andrew (changes requested):
The SKILL.md content covers everything #660 asked for, and the fixture code is clean (no answer-giving comments). A few things to address.

Heads up: #671 also targets #660 with a two-fixture layout; we'll want to converge on one of them.

evals/security-deep-dive-auth-omission.yaml
abhinavgautam01: Dropped `abort(401)` from the negative assertion. `safe_account` is now the sole discriminator, so the negative assertion only rejects a finding that identifies the fail- closed endpoint.

internal/evals/evals_test.go
abhinavgautam01: Replaced the field-mirroring test with a `HeuristicJudge` test using a representative fail-open report. It now proves the required omission finding matches while the safe-endpoint negative assertion remains unmatched.

skills/security-deep-dive/SKILL.md
abhinavgautam01: Folded the omission-check guidance into Step 2 and removed the separate `Step 2a` heading, keeping the audit procedure’s existing numbered structure intact.

abhinavgautam01:
Thanks @andrew , addressed the review points...

> Heads up: #671 also targets #660 with a two-fixture layout; we'll want to converge on one of them.

I compared #671. It overlaps rather than complements this PR: it places the same omission check in Step 1 and uses separate fail-open API-key and fail-closed session fixtures, while this PR places it in Step 2 and uses paired endpoints.

I agree only one implementation should land. I’m happy to fold the two-fixture layout into this PR or close this PR in favour of #671, depending on your preference.

andrew (approved):

PR #671, opened 2026-07-18
Audit missing and empty security inputs for fail-open validation bypasses

### LINKED ISSUE #660: Audit missing and empty security inputs for fail-open validation bypasses

Dodothereal:
# Audit missing and empty security inputs for fail-open validation bypasses

## Summary

The `security-deep-dive` skill's Phase 2 traced what hostile value reaches a sink but never asked what the code does when a security input is *absent*. A credential, session, CSRF, rate-limit, or authz check that runs its validation only when a value is present fails open the moment an attacker omits the input — before they have to forge one. This PR adds the omission check to the per-sink checklist and adds two eval fixtures (fail-open and fail-closed) that exercise it.

Fixes #660

## Changes

- `skills/security-deep-dive/SKILL.md`: new paragraph in Phase 2 Step 1 (Trace the input). After the existing dead-end rule, the trace now also names the read primitive (header accessor, key lookup, cookie read), the sentinel value a miss returns (`None`, empty string, missing key, unset cookie, `0`/`-1`), and which control-flow branch each sentinel falls into. If the absent case reaches the authorised or serviced path, the sink fails open and the bypass is carried into Step 2 as its own finding rather than ruled out as harmless. The check is scoped to security decisions (authn, authz, session, CSRF, rate-limit, credential) and explicitly narrows away from optional non-security inputs and from dev-only / trusted entry points that the threat model already scopes out.
- `evals/fixtures/fail-open-missing-apikey/`: tiny Flask app whose `/admin` endpoint runs its `X-API-Key` comparison only when the header is present, so an omitted header reaches the authorised branch (the `should_find` case).
- `evals/security-deep-dive-fail-open.yaml`: scenario expecting an authentication-bypass finding (CWE-287, High) tracing `X-API-Key` / `EXPECTED_KEY`, with `must_not_contain` guardrails.
- `evals/fixtures/fail-closed-missing-session/`: tiny Flask app whose `/profile` endpoint rejects with 401 when `session["user_id"]` is missing (the negative case).
- `evals/security-deep-dive-fail-closed.yaml`: scenario asserting no authentication-bypass finding is filed for the fail-closed handler.

No finding-schema or ingestion changes; the new scenarios use the existing `security-deep-dive` findings output shape and field set, and the existing sink-class and disposition vocabulary.

## Testing

Verification target was the eval harness (`go test -tags evals ./internal/evals/...`), which validates scenario loading (`TestLoadScenarios` asserts every `*.yaml` loads and its `fixture` directory exists) and the deterministic judge. This VM has no Docker daemon, so the required sandbox (`sandbox-run`, the only permitted path per the contributor rules to run the repo's commands) could not execute — `docker info` reports unavailable and `sandbox-run` fails with "cannot execute docker". I therefore did not run the Go test suite directly on this machine (the repo is untrusted and the rules forbid that).

Instead I verified the new scenarios against the exact contract the harness enforces: parsed both YAML files with `KnownFields`-equivalent strict parsing (no unknown scenario or assertion fields) and checked `given`/`fixture`/`skill` are non-empty, every assertion is non-empty, each references a real `evals/fixtures/<dir>` that exists on disk, and each matches `internal/evals/types.go`'s validate() rules (non-blank assertions, non-blank `must_not_contain` terms, structural fields only). Both fixtures are also syntactically valid Python. The harness's load test is purely YAML + filesystem checks against these same files, with no model call in the default path, so the new scenarios load the same way the existing three do.

---
### AI assistance disclosure

This contribution was produced by an autonomous AI coding agent (Claude Code) that @Dodothereal operates and monitors. @Dodothereal is accountable for it, will address review feedback promptly, and will close this PR immediately if this kind of contribution is unwelcome in this project. Commits carry an `Assisted-by: Claude Code` trailer.


evals/fixtures/fail-open-missing-apikey/app.py
```diff
@@ -0,0 +1,20 @@
...
+
+EXPECTED_KEY = "supersecret"
+
+
+@app.get("/admin")
+def admin():
+    supplied = flask.request.headers.get("X-API-Key")
+    # Only reject when a key _is_ given and differs from the expected one.
+    # A request that omits the header entirely falls through to the
+    # authorised branch, so the check fails open on a missing credential.
```
andrew, line 13: This comment tells the model the answer (`the check fails open on a missing credential`). The existing `sqli-basic/app.py` fixture has no explanatory comments; the README covers intent for humans and the code is left bare so the eval measures the prompt, not comment-reading. Drop the inline comment here and in `fail-closed-missing-session/app.py`.

evals/fixtures/fail-closed-missing-session/app.py
```diff
@@ -0,0 +1,19 @@
...
+app = flask.Flask(__name__)
+app.secret_key = "dev"
+
+
+@app.get("/profile")
+def profile():
+    user_id = flask.session.get("user_id")
+    # No session means no principal: reject the request before any
+    # authorised branch runs. The credential is required, so the
+    # check fails closed on a missing value.
```
andrew, line 12: Same as the fail-open fixture: drop this inline comment so the model has to work it out from the code.

evals/security-deep-dive-fail-open.yaml
```diff
@@ -0,0 +1,17 @@
...
+fixture: fixtures/fail-open-missing-apikey
+skill: security-deep-dive
+should_find:
+  - finding: authentication bypass
+    severity: High
+    cwe: CWE-287
+    path: app.py
+    evidence_contains:
+      - X-API-Key
+      - EXPECTED_KEY
```
andrew, line 11: #660 asks the positive assertion to require evidence for both the conditional check and the protected operation. `X-API-Key` and `EXPECTED_KEY` are both the check side. Add a term for the protected path, e.g. `secret admin payload`.

andrew (changes requested):
The SKILL.md paragraph is well placed and covers everything #660 asked for, and splitting fail-open/fail-closed into separate fixtures is the right call. Two things to fix before this lands.

Heads up: #664 also targets #660.

andrew:
Closing in favour of #664, which covers #660 and has now merged. Thanks for the work here — happy to look at anything from this branch that isn't covered by #664 as a follow-up.

PR #672, opened 2026-07-18
Audit bulk-bound request bodies for mass assignment

### LINKED ISSUE #661: Audit bulk-bound request bodies for field-level authorization and mass assignment
## Summary

Add a focused mass-assignment check to `security-deep-dive` for application code that bulk-binds a request body into a persistence model, domain object, or downstream service payload.

The existing deep-dive covers object-level authorization, but it does not explicitly distinguish format/schema validation from field-level authorization. A body can be perfectly valid according to its DTO or JSON schema while still allowing a caller to set fields such as `role`, `owner_id`, `status`, `permissions`, or an authoritative price.

This can be implemented as a prompt and eval change. It does not need a new skill or output type.

## Scope

Update `skills/security-deep-dive/SKILL.md` so that, when all of the following are true, the audit performs a field-level authorization check:

- The target is application/service code with authenticated or differently privileged actors.
- An externally supplied object/body is deserialized or bound in bulk.
- The resulting object is persisted, used for an authorization decision, or forwarded to a privileged downstream service.

For those sites:

1. Identify fields copied by bulk update, mapper, spread/catch-all, ORM binding, or equivalent framework behavior.
2. Identify fields whose authority belongs to the server or to a more privileged actor: ownership, role/permission, workflow status, security flags, authoritative monetary values, tenant identifiers, and similar project-specific fields.
3. Look for an explicit allow-list, field stripping, server-side overwrite, or per-field authorization before the object reaches the sensitive operation.
4. Do not treat type checks, required-field checks, JSON Schema, DTO annotations, or other shape validation as field-level authorization.
5. When the same body type is accepted by multiple privilege levels, compare the lower-trust binding with the higher-trust path if that comparison is needed to establish intended field ownership.

Keep the output compact: one finding per vulnerable binding/root cause, listing the affected sensitive fields. Do not emit one finding per field when a single allow-list fix closes them together.

Use `CWE-915` when it fits.

## Token and scope constraints

This should not require enumerating every field in every request type.

Trigger the deeper check only at bulk-binding sites that feed persistence, authorization, or privileged downstream calls. Simple request structs whose fields are consumed individually do not need a cross-endpoint field inventory unless the code reveals a suspicious pass-through.

## Evals

Add a minimal fixture and scenario under `evals/`:

1. **Should find:** a lower-privilege update endpoint bulk-copies a request DTO containing a server-owned field such as `role` or `owner_id` into a stored object.
2. **Should not find:** the same endpoint explicitly allow-lists ordinary editable fields or overwrites the sensitive field from the authenticated principal.

The positive assertion should require evidence for the body source, the bulk-copy/binding operation, and the sensitive field.

## Acceptance criteria

- `security-deep-dive` distinguishes schema validation from field-level authorization.
- The check is limited to bulk-bound bodies reaching a sensitive operation.
- Findings group fields by vulnerable binding/root cause.
- A positive mass-assignment eval passes.
- A paired allow-listed/server-overwritten negative eval does not emit a matching finding.
- No schema, worker, database, or UI changes are introduced.

## Non-goals

- General API schema review.
- Treating every writable field as security-sensitive.
- Replacing object-level ownership checks.
- Creating framework-specific reference packs; those can remain part of the broader work in #106.

abhinavgautam01:
Closes #661

## Summary

Adds a focused field-level authorization check to `security-deep-dive` for mass-assignment risks in application and service code.

The check is limited to bulk-bound request bodies that reach persistence, authorization decisions or privileged downstream operations. It distinguishes input-shape validation from authorization over server-owned fields.

## Changes

- Add mass-assignment guidance for bulk-bound request bodies.
- Gate the deeper check on authenticated actors with distinct privileges and a sensitive bulk-binding path.
- Instruct the audit to identify server-owned fields such as:
  - roles and permissions
  - ownership and tenant identifiers
  - workflow or security state
  - authoritative prices and balances
- Require explicit allow-lists, field stripping, server-side overwrites, or per-field authorization before sensitive operations.
- Use `CWE-915` where applicable.
- Group fields sharing one vulnerable binding into a single root-cause finding.
- Add a paired eval fixture:
  - vulnerable member endpoint uses `account.update(body)`
  - safe profile endpoint allow-lists editable fields and overwrites `owner_id`
- Add eval coverage that requires evidence of the request body, bulk update and sensitive `role` field while rejecting a matching finding for the safe endpoint.

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`


skills/security-deep-dive/SKILL.md
```diff
@@ -92,6 +92,7 @@ Sink classes to enumerate. The classes are conceptual; the language you are audi
...
..., or a privileged downstream call. At each such binding, identify every field copied by an ORM binder, mapper, spread/catch-all, `update`/merge, or equivalent framework operation. Separate ordinary caller-editable fields from authority the server or a more privileged actor owns: ownership and tenant identifiers, roles and permissions, workflow status, security flags, authoritative prices or balances, and project-specific equivalents. Look for an explicit allow-list, stripping of forbidden fields, a server-side overwrite, or a per-field authorization check before the sensitive operation. Type checks, required-field checks, DTO annotations, JSON Schema, and other shape validation are not field-level authorization. Where one body type is accepted by lower- and higher-privilege paths, compare them when that establishes intended ownership. Report one finding per vulnerable binding/root cause, listing all affected sensitive fields; do not emit one finding per field when one allow-list fix addresses them together. Use `CWE-915` when it fits. Do not build a repository-wide field inventory: request structs consumed field-by-field are out of scope unless code shows a suspicious pass-through.
```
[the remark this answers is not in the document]
abhinavgautam01: Narrowed the guidance to determine whether a bulk binding can copy sensitive server-owned fields. It now explicitly says not to enumerate ordinary caller-editable fields.

evals/security-deep-dive-mass-assignment.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - role
+    required: true
+should_not_find:
+  - finding: mass assignment
+    cwe: CWE-915
+    path: app.py
+    evidence_contains:
+      - update_profile
+      - account.update(editable)
+      - owner_id
```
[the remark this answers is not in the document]
abhinavgautam01: Added optional exact location matching for eval assertions, since path deliberately normalizes away line numbers. The negative assertion now targets app.py:42; the regression test makes the real finding mention the safe endpoint and verifies it still passes.

internal/evals/evals_test.go
```diff
@@ -342,6 +342,42 @@ func TestRunnerRejectsSchemaInvalidReport(t *testing.T) {
...
+  "location":"app.py:42",
+  "trace":"update_profile uses account.update(editable) and overwrites owner_id."
+}]}`
+	got, err = (HeuristicJudge{}).Judge(sc, safeReport)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got[1].Matched {
+		t.Fatalf("allow-listed endpoint unexpectedly passed should_not_find: %+v", got[1])
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: Added a result-count check and asserted that the inspected result is should_not_find before accessing got[1].

evals/security-deep-dive-mass-assignment.yaml
```diff
@@ -0,0 +1,21 @@
...
+    evidence_contains:
+      - request.get_json()
+      - account.update(body)
+      - role
+    required: true
+should_not_find:
+  - finding: mass assignment
+    cwe: CWE-915
+    path: app.py
+    location: app.py:42
```
andrew: `app.py:42` is the `"display_name": body.get(...)` dict entry, not a location a model would report for a `/profile` finding (that'd be line 37, 46, or 36). With exact-match `location` plus three required evidence terms, this `should_not_find` will effectively never trip in a real run; the unit test only passes because the fake `safeReport` hard-codes `app.py:42`.

Exact-line matching on a negative assertion is fragile whichever line you pick. Suggest splitting the fixture into `account.py` and `profile.py` so `path: profile.py` disambiguates the negative case and `location:` can be dropped from this scenario. That also removes the need for the new `Assertion.Location` field in this PR; it can land separately if there's a `should_find` use for it.

internal/evals/evals_test.go
```diff
@@ -342,6 +344,48 @@ func TestRunnerRejectsSchemaInvalidReport(t *testing.T) {
 	}
 }
 
+func TestMassAssignmentScenario(t *testing.T) {
+	sc := mustLoadScenario(t, "../../evals/security-deep-dive-mass-assignment.yaml")
+	report := `{"findings":[{
+  "title":"Mass assignment in update_account",
+  "cwe":"CWE-915",
+  "location":"app.py:27",
+  "trace":"request.get_json() supplies body, and account.update(body) copies role without an allow-list. Unlike update_profile, it does not use account.update(editable) or overwrite owner_id."
```
andrew: This trace quotes `update_profile`, `account.update(editable)`, and `owner_id` specifically so that evidence-only matching would misfire on the correct finding, which is what motivates the `location` field. Real deep-dive traces don't typically quote the safe path's code verbatim. If the fixture is split by file (see the scenario comment) this test simplifies and the contrived contrast text can go.

andrew (changes requested):
The SKILL.md guidance and fixture look good. The `should_not_find` guard needs rework before this is useful in a real eval run.

evals/security-deep-dive-mass-assignment.yaml
abhinavgautam01: Reworked the fixture into separate account.py and profile.py route modules. The negative assertion now matches path: profile.py, so it no longer depends on an exact source line. Removed the temporary Assertion.Location support from this PR.

internal/evals/evals_test.go
abhinavgautam01: Simplified the positive fake finding to describe only the vulnerable account.py path. The separate-file fixture now provides the required disambiguation without embedding safe-path code into the vulnerable finding’s trace.

andrew (approved):

PR #673, opened 2026-07-18
Validate security-deep-dive sink dispositions

### LINKED ISSUE #662: Validate that every security-deep-dive sink has a consistent disposition
## Summary

Add deterministic semantic validation for the coverage invariant already required by `security-deep-dive`:

> Every inventory sink must appear either in `findings[].sinks` or in `ruled_out[].sinks`.

The skill states this requirement in `skills/security-deep-dive/SKILL.md`, and the repository report renders missing dispositions as `unresolved`. JSON Schema cannot enforce cross-array references, however, so a schema-valid report can silently omit a sink after fan-out consolidation and still be ingested as a completed scan.

This issue makes that existing contract machine-enforced. It does not add model work and should usually reduce wasted scans by letting the existing repair loop correct incomplete output immediately.

## Current locations

- Contract: `skills/security-deep-dive/SKILL.md`, Output section.
- IDs and references: `skills/security-deep-dive/schema.json`.
- Schema validation/repair flow: `internal/worker/schema_validate.go` and the validation calls in `internal/worker/skill.go`.
- Eval validation: `internal/evals/runner.go`.
- Existing unresolved rendering: `internal/web/repo_report.go` (`buildSinkDispositions`).

## Required invariants

For a `security-deep-dive` report:

1. Every `inventory[].id` is unique.
2. Every inventory ID is referenced by at least one finding or ruled-out entry.
3. Every value in `findings[].sinks` and `ruled_out[].sinks` refers to an existing inventory ID.
4. An ID must not be both a finding sink and ruled out. Contradictory outcomes should be repaired rather than relying on UI precedence.
5. Repeating an ID within the same finding or ruled-out entry should be rejected or normalized consistently.

An empty inventory remains valid for a genuinely empty scoped/diff scan if the current schema permits it.

## Suggested implementation

Add a small semantic validator alongside schema validation, for example:

```go
func ValidateReportSemantics(skillName, report string) string
```

Only `security-deep-dive` needs a validator initially. Parse a minimal local struct containing `inventory[].id`, `findings[].sinks`, and `ruled_out[].sinks`; do not tighten the main findings ingestion struct or start rejecting unrelated model-output fields.

Feed semantic validation failures through the same one-resume repair path used for JSON Schema failures. The repair message should identify exact IDs, for example:

- `inventory sink S17 has no disposition`
- `finding F3 references unknown sink S99`
- `sink S4 appears in both findings and ruled_out`

Run the same semantic validation in the eval runner so fixtures cannot pass with incomplete coverage.

Preserve the existing strictness behavior:

- If repair succeeds, continue normally.
- If repair still fails, follow the same warning-versus-failure policy as remaining schema errors and `schema_strict`.

## Tests

Add table-driven unit tests covering:

- complete report;
- unresolved inventory ID;
- unknown referenced ID;
- duplicate inventory ID;
- finding/ruled-out conflict;
- repeated reference;
- empty inventory;
- malformed JSON remains handled by the schema/JSON path rather than panicking.

Add or update a worker test proving semantic failures are included in the repair prompt and are rechecked after repair. Add an eval-runner test proving incomplete coverage fails validation.

## Acceptance criteria

- Schema-valid deep-dive reports with unresolved, unknown, duplicate, or contradictory sink references are detected.
- Validation participates in the existing repair and strictness flow.
- Both normal scans and model-backed evals apply the invariant.
- Other skills are unaffected.
- No database migration or UI change is required.

## Non-goals

- Adding an input/source inventory.
- Changing how sink dispositions are displayed.
- Enforcing semantic contracts for every output kind in this issue.
- Reworking findings ingestion or fingerprinting.

abhinavgautam01:
Fixes #662

## Summary

Adds skill-specific semantic validation for `security-deep-dive` reports so every inventoried sink has one consistent disposition.

## Changes

- Require unique inventory sink IDs.
- Require each inventory sink to be referenced by either a finding or a `ruled_out` entry.
- Reject unknown, repeated, and conflicting sink references.
- Apply the semantic checks during normal parsing, repair attempts, the report-validation API and eval runs.
- Keep all non-deep-dive skills on their existing JSON-Schema-only validation path.
- Add coverage for each semantic invariant and eval rejection of an incomplete deep-dive report.

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`

### DIFF
--- a/internal/web/api.go
+++ b/internal/web/api.go
@@ -307,7 +307,7 @@ func (s *Server) apiValidateReport(w http.ResponseWriter, r *http.Request) {
 		writeAPIError(w, http.StatusRequestEntityTooLarge, "could not read body (max 1 MB)")
 		return
 	}
-	if detail := worker.ValidateReportSchema(skill.SchemaJSON, string(body)); detail != "" {
+	if detail := worker.ValidateSkillReport(skill.Name, skill.SchemaJSON, string(body)); detail != "" {
 		writeJSON(w, http.StatusOK, map[string]any{"valid": false, "errors": detail})
 		return
 	}
--- a/internal/web/api.go
+++ b/internal/web/api.go
@@ -277,7 +277,7 @@ func (s *Server) apiGetScan(w http.ResponseWriter, r *http.Request) {
 // its skill's schema without installing a JSON Schema library inside the runner
 // container. The request body is the candidate report; the response is
 // {"valid":true} or {"valid":false,"errors":"..."} using the exact same
-// validator (worker.ValidateReportSchema) the harness runs after the scan, so
+// validator (worker.ValidateSkillReport) the harness runs after the scan, so
 // an in-container pass guarantees the harness will not send a repair prompt.
 //
 // The body is already capped at apiMaxBody (1 MB) by apiAuth's MaxBytesReader;
--- a/internal/worker/exposure.go
+++ b/internal/worker/exposure.go
@@ -228,7 +228,7 @@ func (w *Worker) recordExposureResult(skill *db.Skill, scan *db.Scan, depID uint
 func (w *Worker) parseExposureOutput(skill *db.Skill, scan *db.Scan, depID uint, report string, emit func(Event)) error {
 	if skill.SchemaJSON != "" {
 		if detail := ValidateSkillReport(skill.Name, skill.SchemaJSON, report); detail != "" {
-			emit(Event{Kind: KindError, Text: "schema: report.json does not validate against schema.json:\n" + detail})
+			emit(reportValidationEvent(skill, detail))
 			if w.SchemaStrict {
 				return &SchemaValidationError{Skill: skill.Name, Detail: detail}
 			}
--- a/internal/worker/exposure.go
+++ b/internal/worker/exposure.go
@@ -227,7 +227,7 @@ func (w *Worker) recordExposureResult(skill *db.Skill, scan *db.Scan, depID uint
 // back to under_investigation; invalid justification labels are dropped.
 func (w *Worker) parseExposureOutput(skill *db.Skill, scan *db.Scan, depID uint, report string, emit func(Event)) error {
 	if skill.SchemaJSON != "" {
-		if detail := ValidateReportSchema(skill.SchemaJSON, report); detail != "" {
+		if detail := ValidateSkillReport(skill.Name, skill.SchemaJSON, report); detail != "" {
 			emit(Event{Kind: KindError, Text: "schema: report.json does not validate against schema.json:\n" + detail})
 			if w.SchemaStrict {
 				return &SchemaValidationError{Skill: skill.Name, Detail: detail}
--- a/internal/worker/schema_validate.go
+++ b/internal/worker/schema_validate.go
@@ -3,7 +3,6 @@ package worker
 import (
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strings"
 
 	"github.com/santhosh-tekuri/jsonschema/v6"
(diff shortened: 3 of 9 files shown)


internal/worker/exposure.go
```diff
@@ -227,7 +227,7 @@ func (w *Worker) recordExposureResult(skill *db.Skill, scan *db.Scan, depID uint
 // back to under_investigation; invalid justification labels are dropped.
 func (w *Worker) parseExposureOutput(skill *db.Skill, scan *db.Scan, depID uint, report string, emit func(Event)) error {
 	if skill.SchemaJSON != "" {
-		if detail := ValidateReportSchema(skill.SchemaJSON, report); detail != "" {
+		if detail := ValidateSkillReport(skill.Name, skill.SchemaJSON, report); detail != "" {
 			emit(Event{Kind: KindError, Text: "schema: report.json does not validate against schema.json:\n" + detail})
```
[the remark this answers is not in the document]
abhinavgautam01: Updated exposure-report validation failures to use the shared validation: ... does not pass Scrutineer report validation event, so the operator-facing message remains correct for both schema and skill-specific validation failures.

internal/web/api.go
```diff
@@ -304,7 +304,7 @@ func (s *Server) apiValidateReport(w http.ResponseWriter, r *http.Request) {
 		writeAPIError(w, http.StatusRequestEntityTooLarge, "could not read body (max 1 MB)")
 		return
 	}
-	if detail := worker.ValidateReportSchema(skill.SchemaJSON, string(body)); detail != "" {
+	if detail := worker.ValidateSkillReport(skill.Name, skill.SchemaJSON, string(body)); detail != "" {
```
andrew, line 310: The doc comment above (line 277) still names `worker.ValidateReportSchema` as the validator the harness runs. It's now `ValidateSkillReport`; the guarantee in the next clause still holds, just update the function name so the comment stays accurate.

internal/worker/schema_validate.go
```diff
@@ -87,3 +113,94 @@ func formatValidationError(ve *jsonschema.ValidationError) string {
...
+		dispositions[id] = struct{}{}
+	}
+	return errs
+}
+
+func formatSemanticValidationErrors(errs []string) string {
+	if len(errs) == 0 {
+		return ""
+	}
+	sort.Strings(errs)
```
andrew: Non-blocking: `sort.Strings` reorders the repair prompt so reference errors can appear before the inventory errors that cause them. The walk order (inventory → findings → ruled_out → coverage) is already deterministic, so you could drop the sort and keep that order for the model.

andrew (changes requested):
One stale reference and one optional suggestion, otherwise this looks good.

internal/web/api.go
abhinavgautam01: Updated the `apiValidateReport` comment to reference `worker.ValidateSkillReport`, which is the validator the endpoint now calls.

internal/worker/schema_validate.go
abhinavgautam01: Removed the final sort so semantic validation errors retain their deterministic causal order: inventory errors first, then dependent reference errors, then disposition coverage. Added a focused test to pin this ordering for repair prompts.

andrew (approved):

andrew:
Approved, but this now conflicts with main in `internal/evals/evals_test.go` and `internal/web/api.go` (churn from #652/#665). Please rebase and I'll merge.

andrew (approved):

PR #685, opened 2026-07-21
Add opt-in model-backed eval judge

### LINKED ISSUE #65: Eval harness for skills: should_find / should_not_find fixtures
We tune `security-deep-dive` and other model-backed skills by hand and have no regression signal when SKILL.md changes. Add a fixture-driven eval harness so prompt edits can be measured.

Fixture format, one YAML per scenario under `evals/`:

    given: a handler that concatenates user input into a SQL string
    fixture: fixtures/sqli-basic
    skill: security-deep-dive
    should_find:
      - finding: SQL injection in buildQuery
        severity: high
        required: true
    should_not_find:
      - the unused import on line 3

Runner executes the real skill against `fixture` via the existing `worker.RunSkill` path, then a judge call (depends on the auxiliary model helper) scores each assertion: was every `should_find` met, was any `should_not_find` violated. Required misses fail the test; optional misses are reported.

- [ ] `evals/` directory with YAML scenarios and `evals/fixtures/*` mini-repos
- [ ] `internal/evals` package: loader, runner, judge
- [ ] `go test -tags evals ./internal/evals/...` entrypoint, off by default in CI
- [ ] Seed scenarios for `security-deep-dive`, `semgrep`, `zizmor`
- [ ] Per-scenario cost reporting so we know what a full eval run costs

abhinavgautam01:
Closes #65

## Summary

Adds an opt-in, model-backed judge for fixture-driven skill evaluations.

The existing deterministic judge remains the default for local and CI checks. Live eval runs can now request a semantic verdict from the Anthropic Messages API when assertion matching needs model judgment.

## Changes

- Add `ModelJudge` backed by structured Anthropic Messages API output
- Require exactly one validated verdict for every eval assertion
- Preserve assertion ordering and reject incomplete, duplicate, or invalid verdict responses
- Add judge usage and pricing to each scenario's total cost reporting
- Keep the deterministic `HeuristicJudge` as the default
- Add `SCRUTINEER_EVAL_JUDGE=model` opt-in for live fixture runs
- Document required environment variables and model-judge usage

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`

### DIFF
--- a/evals/README.md
+++ b/evals/README.md
@@ -43,5 +43,20 @@ criteria, not evidence.
 The default judge matches findings by title substring plus optional severity,
 CWE, path, and evidence. These assertions define a minimum bar: additional
 findings do not fail an eval unless they match `should_not_find` or the report
-contains a `must_not_contain` term. Model-backed judging can be plugged in by
-implementing `evals.Judge`.
+contains a `must_not_contain` term.
+
+For a semantic, model-backed verdict during a live run, opt in explicitly. The
+judge uses the Anthropic Messages API and its cost is included in each
+scenario's reported cost:
+
+```sh
+SCRUTINEER_RUN_EVALS=1 \
+  SCRUTINEER_EVAL_MODEL=claude-sonnet-5 \
+  SCRUTINEER_EVAL_JUDGE=model \
+  SCRUTINEER_EVAL_JUDGE_MODEL=claude-haiku-4-5 \
+  ANTHROPIC_API_KEY=sk-ant-... \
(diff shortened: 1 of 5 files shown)


andrew (approved):
LGTM. Token normalisation matches the `worker.CostFromUsage` contract, verdict validation is sound, and the default path is unchanged.

Tiny nit for a follow-up: the two `ModelJudge` tests set `Endpoint` inconsistently (one appends `/v1/messages`, one doesn't).

Issue #106, opened 2026-04-30
Expand and sharpen the skill set


andrew:
I cloned and read through the security skill repos doing the rounds (Trail of Bits, Sentry Warden, Latio, raptor, Phoenix, netresearch, AgentSecOps, Anthropic's reviewer, and a few of the awesome-lists). Most of what's out there is either a thin wrapper around an existing CLI (`run trivy`, `run gitleaks`, `run nmap`) or bulk-generated filler. After deduping against what we already ship, there are about eight ideas actually worth taking, four of which I think are clear wins.

For reference, we currently have 16 skills: nine that build context (metadata, repo-overview, packages, dependencies, dependents, advisories, maintainers, sbom, subprojects), three that audit code (security-deep-dive, semgrep, zizmor), and four workflow skills (triage, verify, patch, disclose). The gaps are mostly on the audit side.

### Narrow per-class audit skills (source: [getsentry/warden-skills](https://github.com/getsentry/warden-skills))

Sentry split their audit into five single-purpose skills (`code-execution`, `data-exfil`, `authz`, `pii`, `gha-workflows`) instead of one broad one. Each ships a `references/` directory of framework-specific notes: the authz skill carries twelve files covering how Django, DRF, FastAPI, NestJS, Next.js and GraphQL each express guards, so the model knows that a missing decorator in DRF might be covered by a viewset mixin three files away. The code-execution skill is version-aware (PyYAML >=5.1 changed the default loader, `vm2` is abandoned, jsonwebtoken >=9.0 fixed alg confusion) which kills a whole class of false positives our current deep-dive emits.

Our `security-deep-dive` already does sink-class enumeration in Phase 1, so we're halfway there. The proposal is to keep deep-dive as the inventory pass and add four narrow skills that each take one sink class from the inventory and work it with a dedicated reference pack:

- `audit-injection` (command/code exec, deserialisation, SSTI)
- `audit-exfil` (SSRF, path traversal, XXE, response leakage)
- `audit-authz` (IDOR, tenant isolation, fail-open guards, JWT claims)
- `audit-pii` (real identifiers in code/logs/fixtures, with synthetic-data allowlisting)

[...]

PR #686, opened 2026-07-21
Add finding-scoped variants skill

abhinavgautam01:
Part of #106

## Summary

Adds an opt-in, finding-scoped `variants` skill for identifying distinct, high-confidence sibling instances of a confirmed vulnerability's root cause within the same repository.

The skill is intentionally narrow and read-only. It does not run during triage, does not scan dependents and returns an empty findings report when no independently actionable variant is verified.

## Changes

- Add bundled `variants` skill with:
  - Finding-scoped API reads for the source finding and existing repository findings
  - Static-only, repository-local variant analysis
  - Explicit checks for distinctness, reachability, missing controls, currentness and impact
  - High-confidence-only reporting
  - Required source-finding attribution through `prior_art`
- Add a strict findings schema for variant reports.
- Add metadata and schema validation coverage, including invalid confidence and missing attribution cases.
- Document the new skill and its finding-scoped context.

## Scope

This implements the intra-repository variant-analysis portion of #106. Cross-dependent or downstream variant analysis remains outside this skill's scope.

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`


andrew (changes requested):
A few things to fix before this lands:

**`reachability` enum diverges from the project vocabulary.** `skills/variants/schema.json` uses `["reachable", "conditional", "unknown"]`, but every other findings schema (`vuln-scan`, `security-deep-dive`, `advisory-deep-dive`, `reachability`) and the DB doc at `internal/db/db.go:632` use `["reachable", "harness_only", "unclear"]`. Ingest doesn't validate this field, so `conditional`/`unknown` would land in the indexed `findings.reachability` column and not match any existing filter or export. Given SKILL.md makes traced reachability a hard validation gate, `["reachable"]` alone may be the right enum here.

**`confidence` schema enum vs `min_confidence` metadata.** `min_confidence: high` drops below-threshold findings at ingest and keeps the rest; the schema `enum: ["high"]` hard-fails the whole report if one entry comes back `medium`. With both, one stray value loses an otherwise-good run. I'd keep `min_confidence` and relax the schema to the standard confidence enum.

**SKILL.md**: "Do not analyse a source finding that is rejected, duplicate, fixed, or published" should say what to write — `{"findings":[]}`, matching the missing-context and failed-request cases.

No UI button is wired, so this is only reachable via the generic finding-skill API endpoint. That matches `forensics` so I'm assuming it's intentional for a first cut of #106.

abhinavgautam01:
Thanks, fixed.

- Relaxed the schema to the shared `high` / `medium` / `low` confidence vocabulary. The existing `scrutineer.min_confidence: high` gate now filters below-threshold rows during ingestion without allowing one medium row to invalidate the whole report.
- Restricted `reachability` to `reachable`: this skill only emits candidates after a verified source-to-sink trace. Added schema coverage that rejects `unclear`.
- Made the closed-source-finding path explicit: rejected, duplicate, fixed, or published sources now write `{"findings":[]}` and exit successfully.

The API-only launch surface is intentional for this first cut and matches the existing `forensics` skill.

andrew (approved):

PR #687, opened 2026-07-21
feat: share retry primitives and extend bounded git retries

### LINKED ISSUE #683: Adopt gitRetry in the remaining remote git callers
Follow-up to #682, which added `gitRetry` for `cloneOrFetch`, `fetchRef`, and `ListRemoteBranches`. Two other places still make one remote git attempt and give up.

`internal/worker/upstream.go`:

- `ResolveRemoteHead` runs `ls-remote` once. Same shape as `ListRemoteBranches`, so the tight picker budget fits.
- `SyncUpstream` runs `clone --bare`, `fetch`, and `push --force` once each. The clone and fetch can take the scan policy directly. The force push needs a decision rather than a blanket retry: repeating a push that half-completed is not obviously idempotent, and a transient failure after the ref updated would retry a no-op at best.

Both already inherit `WaitDelay` via `gitWithEnv`, so this is only about the retry loop.

`internal/skills/git.go` `CloneOrPull` has its own `git()` helper and a direct `exec.CommandContext` for the clone, so it gets neither the retry nor `WaitDelay`. It needs the destination-reset hook for the same reason `cloneOrFetch` did: a partial clone leaves a non-empty target that the next attempt refuses.

The classifier and policy in `internal/worker/git_retry.go` are self-contained and could move somewhere both packages can reach if `internal/skills` should not import `internal/worker`.

### LINKED ISSUE #684: Dedupe backoff helpers between internal/httpx and internal/worker
#682 added `gitBackoffDelay`, `gitJitter`, and `gitSleep` in `internal/worker/git_retry.go`. `internal/httpx/retry.go` already has `backoffDelay` with the same doubling-to-a-cap shape and its own context-aware sleep.

The git version adds jitter; the httpx version reads `Retry-After`. A shared helper could carry both and each caller keep its own transient/permanent decision.

If #683 moves the git retry policy out of `internal/worker` so `internal/skills` can reach it, that is a natural place for the shared bits to land.

p-linnane:
Closes #683 and closes #684. #682 added a bounded retry policy for the scan clone and fetch paths, but left its backoff, jitter, and context-aware sleep duplicated against `internal/httpx`, and kept its git runner and transient/permanent classifier inside `internal/worker` where `internal/skills` could not reach them.

This extracts the transport-agnostic timing primitives into a new `internal/retry` package and the git runner, WaitDelay handling, failure classifier, and clone-destination cleanup into a new `internal/gitx` package, then routes `internal/httpx` through the shared timing helpers. The move is behaviour-preserving: the backoff shape, the fail-closed classification, and every transient and permanent marker are carried over verbatim, and the HTTP path keeps its own Retry-After handling and 429/5xx decision.

With the policy reachable from both packages, the remaining single-attempt callers adopt it: `ResolveRemoteHead` on the tight branch-picker budget, the `SyncUpstream` bare clone and upstream fetch on the scan budget, and the skills `CloneOrPull` clone, cached fetch, and explicit-ref fetch (which also gain the WaitDelay bound they never had). Local-only work such as reset, rev-parse, and symbolic-ref stays outside the policy, since repeating it cannot fix what a retry budget is meant to fix.

The force push in `SyncUpstream` gets a decision rather than a blanket retry, because a transient failure after the destination ref already moved would re-push a no-op at best. On an ambiguous failure it inspects the destination ref and treats the push as already done when the ref holds the SHA that was just fetched, otherwise falling back to the push's own bounded retry budget. Credential hygiene is unchanged: retry notices still carry neither the remote URL nor git's output, and the reconciliation lookup discards its own error rather than surfacing it.

Highlights:
- new `internal/retry` (backoff, jitter, context-aware sleep) and `internal/gitx` (runner, WaitDelay, classifier, clone-dest cleanup), shared by the worker, skills, and HTTP callers
- bounded retries added to `ResolveRemoteHead`, the `SyncUpstream` clone and fetch, and the skills `CloneOrPull` clone, cached fetch, and explicit-ref fetch
- force-push reconciliation: an ambiguous push that already installed the fetched SHA is not repeated
- classifier, retry budgets, environment hardening, cancellation semantics, and fail-closed behaviour preserved from #682; local-only git operations are never retried

### DIFF
--- a/internal/gitx/retry.go
+++ b/internal/gitx/retry.go
@@ -1,0 +1,304 @@
+// Package gitx runs Git commands and applies a bounded retry policy to
+// network-facing invocations.
+package gitx
+
+import (
+	"context"
+	"errors"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	retryx "scrutineer/internal/retry"
+)
+
+const (
+	DefaultAttempts  = 3
+	DefaultBaseDelay = 500 * time.Millisecond
+	DefaultMaxDelay  = 4 * time.Second
+	DefaultWaitDelay = 10 * time.Second
+)
+
+// Runner runs one Git invocation and returns its combined output.
+type Runner func(ctx context.Context, dir string, env []string, args ...string) (string, error)
+
+// Notice describes a transient failure before the next attempt.
+type Notice struct {
+	Label    string
+	Attempt  int
+	Attempts int
+	Delay    time.Duration
+}
+
+// Command is one remote Git invocation plus operation-specific hooks.
+type Command struct {
+	Label string
+	Dir   string
+	Env   []string
+	Args  []string
+	// Reset runs after a failed attempt, before either another attempt or a
+	// terminal error return. It must only clean command-owned state.
+	Reset func() error
+	// Confirm may recognize that an operation succeeded despite an ambiguous
+	// transient error. A confirmation error does not replace the original Git
+	// error; the normal retry budget continues unless the context ended.
+	Confirm func(context.Context) (bool, error)
+}
+
+// Retry bounds how a remote Git invocation is retried. Its zero value is the
+// production policy; fields are exposed so callers can use a tighter budget
+// or deterministic runners and sleepers in tests.
+type Retry struct {
+	Attempts  int
+	BaseDelay time.Duration
+	MaxDelay  time.Duration
+	Run       Runner
+	Sleep     func(context.Context, time.Duration) error
+	Notify    func(Notice)
+}
+
+// Resolved fills zero-valued options with the production defaults.
+func (r Retry) Resolved() Retry {
+	if r.Attempts <= 0 {
+		r.Attempts = DefaultAttempts
+	}
+	if r.BaseDelay <= 0 {
+		r.BaseDelay = DefaultBaseDelay
+	}
+	if r.MaxDelay <= 0 {
+		r.MaxDelay = DefaultMaxDelay
+	}
+	if r.Run == nil {
+		r.Run = Run
+	}
+	if r.Sleep == nil {
+		r.Sleep = retryx.Sleep
+	}
+	return r
+}
+
+// Do runs cmd, retrying only transient failures while the budget, context,
+// cleanup hook, and optional success confirmation allow another attempt.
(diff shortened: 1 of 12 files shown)


andrew (changes requested):
The extraction is clean and the new call sites are well covered, but the move drops reasoning comments that #682 established and that a shared package needs more, not less.

`internal/gitx/retry.go`:
- The `permanentFailures` / `transientFailures` lists lost every inline comment explaining the non-obvious entries: why permanent markers are checked first (mixed output like `No space left on device` followed by `the remote end hung up unexpectedly`), why `service temporarily unavailable` is narrowed from `temporarily unavailable` (local EAGAIN vs nginx 503), what the Cloudflare 520-524 range is. The next person adding a marker won't have that context.
- `CloneDestReset` now does an exported recursive delete with a two-line godoc. The original explained why `os.RemoveAll(dst)` is safe here (the clone path is only reached when dst had no `.git`, so anything present was put there by this call). Please carry that over.

Minor: `containsString` in `internal/skills/git_test.go` can be `slices.Contains`.

Everything else looks right: `Confirm` ordering, `branchPickerRetry` preserving injected `run`/`sleep`, local ops staying off the retry path, and `remoteRefMatches` making one direct attempt.

p-linnane:
Carried all of that over. The classifier lists now keep the permanent-first rationale, the mixed-output examples, the Cloudflare 520-524 note, and the `temporarily unavailable` narrowing, and `CloneDestReset` keeps the why-RemoveAll-is-safe reasoning. Also switched the test helper to `slices.Contains`. Marker set is unchanged from #682 (verified identical), so it's comments only.

andrew (changes requested):
Please clean clone-owned destinations when retrying ends. `Retry.Reset` currently runs only before another attempt, so context cancellation or an exhausted retry budget leaves the last partial clone behind. The new persistent `skills-cache/...` and `upstream-sync.git` paths can then remain unusable across later invocations. Run the safe reset before terminal error returns, join any cleanup error, and add coverage for exhaustion or cancellation followed by a fresh invocation.

andrew (approved):
Terminal cleanup now runs on cancellation, permanent failure, and exhaustion, with the cleanup error joined and both caller-level fresh-invocation tests in place.

PR #691, opened 2026-07-22
Add Capslock capability prefilter for Go reachability analysis

abhinavgautam01:
Part of #4, as this PR delivers the initial Go implementation; capability prefilters for other languages remain follow-up work under #4.

## Summary

Adds a Go-only, deterministic Capslock capability-analysis skill that provides package-level privileged capability data for later reachability review.

The prefilter narrows call-graph investigation without treating absent capability data as proof that a dependency is safe.

## Changes

- Install pinned `github.com/google/capslock/cmd/capslock@v0.3.2` in the Go runner profile.
- Add a bundled `capslock` skill with:
  - Go-profile requirement
  - Offline local-module analysis via `-force_local_module`
  - Deterministic JSON package-to-capability output
  - Explicit error reporting without model-authored fallback data
- Run Capslock from triage only when the repository has a root `go.mod`.
- Stage the latest validated successful Capslock report for `reachability`.
  - Prefer the current scan group.
  - Fall back only to matching repository, ref, and subpath results.
  - Ignore malformed, missing-skill, and tool-error reports.
- Update reachability guidance to use capability data for call-graph prioritization while preserving normal safety checks.
- Document the skill, Go runner tool, and `context.json` contract.
- Add schema, metadata, context-staging, group-isolation, ref/subpath, and error-path tests.

## Verification

- Ran the pinned Capslock binary through the bundled wrapper against this repository and confirmed wrapped package-capability JSON output.
- `go test ./...`
- `go test -tags evals ./...`
- `go test -race ./internal/worker ./internal/web`
- `go test -race -tags evals ./internal/worker ./internal/web`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`


andrew (changes requested):
1. **Continue past unusable Capslock reports**

   The selector parses only the newest row. A newer malformed, missing-skill, or tool-error report returns `nil`, hiding older valid data and preventing the documented fallback.

   **Suggested change:** Iterate through ordered candidates until a valid, successful report is found. Try the current scan group first, then compatible repository/ref/subpath results.

2. **Clarify or enforce the offline contract**

   `force_local_module` only disables Capslock's temporary-module fallback. Its `packages.Load` call can still trigger normal Go downloads for uncached `go.mod` dependencies.

   **Suggested change:** Either enforce offline operation with a controlled cache and disabled `GOPROXY`, or remove the offline/local-only claims.

abhinavgautam01:
Thanks, fixed.

- The Capslock context selector now iterates through ordered candidate scans instead of parsing only the newest row.
- Same-group reports are still preferred, but malformed reports, missing-skill/schema reports and Capslock error reports are skipped so older valid data can be used.
- If no usable same-group report exists, it falls back to compatible repo/ref/subpath results with the same skip-over-unusable behavior.
- Updated the Capslock skill wording to remove the full offline/local-only claim. It now clarifies that `-force_local_module` disables Capslock's temporary-module fallback, while normal Go package loading may still consult the configured module cache/proxy for declared dependencies.
- Added focused tests for grouped and fallback candidate selection.

andrew:
Closing this along with #4 — not a problem with the implementation, which is solid and addressed both review rounds cleanly.

Revisiting #4 itself: the reachability guidance here (correctly) says the capability map can't rule anything out, only reorder where to look first. Reachability already orders by severity and opens each candidate with a call-site grep, so the net effect is a search-order hint. That hint only applies where a repo has a root `go.mod` *and* gets a reachability run, and reachability targets Ruby/JS apps — the overlap is narrow. Against that, it's a pinned tool in the Go image plus ~150 lines of cross-scan context staging to maintain.

If reachability later grows a Go-app path and we find it's spending real time on call-graph expansion that a deterministic prefilter would cut, this is the right shape to bring back.

Issue #12, opened 2026-04-20
Zombie package detection and migration


andrew:
Detect abandoned packages that still have significant dependent usage, and help consumers migrate.

- [ ] `Repository.health` (active, stale, abandoned, zombie)
- [ ] Health scoring job (dependent count, pushed_at, maintainer activity)
- [ ] Alternatives table (purl, kind: fork, successor, equivalent)
- [ ] Migration guide on the finding
- [ ] Per-dependent campaign tracking (notified, acked, migrated, declined, silent)
- [ ] Re-fetch dependent counts over time and plot on the finding

andrew:
All checked items landed. The last one — re-fetching dependent counts over time and plotting on the finding — is superseded by #808's per-dependent campaign status, which is the direct migration signal rather than a lagging ecosyste.ms proxy. #817 closed without merging for that reason.

PR #692, opened 2026-07-22
Add repository health scoring

abhinavgautam01:
Part of #12

## Summary

Adds repository health scoring so Scrutineer can classify repositories as active, stale, abandoned or zombie based on existing metadata, maintainer and downstream dependency signals.

This gives operators a durable health signal for prioritizing zombie-package and abandoned-maintainer risk without changing scan behavior yet.

## Changes

- Add repository health fields:
  - `health`
  - `health_score`
  - `health_summary`
  - `health_checked_at`
- Add deterministic health assessment logic using:
  - repository archive status
  - last pushed timestamp
  - maintainer activity status
  - maximum downstream dependent count
- Refresh repository health after metadata, package, and maintainer ingestion
- Show health status on repository list and repository detail pages
- Expose health fields through the repository API
- Document the new database fields and OpenAPI schema
- Add tests for health classification and refresh behavior

## Tests

- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `go test -race ./...`
- `git diff --check`


andrew (changes requested):
Thanks for picking this up. The classification logic in `AssessRepositoryHealth` is right and the tests are good — keep both. The storage and refresh side needs rework before this lands.

#12 specs a single `Repository.health` column plus a periodic scoring job. This PR adds four columns and refreshes them from three parser hooks instead. That shape causes two concrete problems:

**The extra three columns are derived data with no reader.** `health_score` isn't consumed by any of the downstream #12 items (alternatives, campaigns, migration guides all key on the enum, not a number). `health_summary` is regenerable at render time from the same inputs — call `AssessRepositoryHealth` in the handler and print `.Summary`. `health_checked_at` is only meaningful because refresh is event-driven; with a periodic job it's just "when the job last ran". Drop all three and keep only `Health RepositoryHealth `gorm:"index"``.

**Parser hooks can't keep a time-dependent value fresh.** Health depends on `now - PushedAt`: a repo pushed 11 months ago is `active` today and should become `stale` next month with zero writes to any input. Hooking refresh into `parseRepoMetadataOutput` / `parsePackagesOutput` / `parseMaintainersOutput` means the stored value only moves when one of those skills happens to re-run, so it drifts. There's also already a missed hook — `internal/worker/ecosystems.go:342` writes `DependentRepos` without calling `RefreshRepositoryHealth`, so a dependent-count change from that path never updates health. A periodic job (item 2 in #12) avoids both: one refresh path, no invalidation sites to miss, and time-based transitions happen on schedule.

Concretely:

- keep `AssessRepositoryHealth`, its tests, and the `RepositoryHealth` type
- keep only the `Health` column on `Repository`; drop `HealthScore`, `HealthSummary`, `HealthCheckedAt` from the struct, `docs/database.md`, `openapi.yaml`, and `api.go`
- drop the three `RefreshRepositoryHealth` calls from `skill.go` / `skill_parsers.go` and their test assertions
- change `RefreshRepositoryHealth` to update only `health`
- in `repo_show.html`, compute the summary at render time (load the repo's packages and maintainers in the handler — they're already loaded for that page — and call `AssessRepositoryHealth`) rather than reading a stored string
- either add the periodic refresh job in this PR (it's #12 item 2) or leave `Health` unpopulated here and land the job next — don't ship the parser hooks as an interim

The `repo_list.html` column can stay as-is reading the stored `Health`.

abhinavgautam01:
Thanks, reworked this to match #12’s storage shape.

- Kept only the stored `Repository.Health` enum and removed `health_score`, `health_summary` and `health_checked_at` from the model, API, OpenAPI, docs, templates and tests.
- Removed the parser-hook refresh calls from metadata, packages, and maintainers parsing.
- Changed `RefreshRepositoryHealth` so it only persists `health`.
- Added a periodic repository health scorer that runs once at startup and then hourly, so time-based classifications can age without requiring another parser run.
- Updated `repo_show` to derive the explanatory summary at render time from the repository, package rows, and maintainers already loaded for the page.
- Left the repo list column reading the stored `Health` enum.

andrew (changes requested):
The rework does what the last review asked - single `health` column, periodic job, render-time summary, and the aging test covers the drift case. Two small things before this lands:

- The method `repositoryHealthTick` shares its name with the package const holding the interval. Rename one of them, e.g. the const to `repositoryHealthInterval`.
- The tick rewrites `health` for every repo every hour even when the value is unchanged. Skip the `Update` when the stored value already matches.

abhinavgautam01:
Thanks, fixed both

- Renamed the interval constant to `repositoryHealthInterval`, leaving `repositoryHealthTick` as the method name.
- Updated `RefreshRepositoryHealth` to skip the `health` update when the stored value already matches the recomputed classification.
- Added regression coverage for the unchanged-health case so the periodic scorer does not rewrite rows unnecessarily.

andrew (approved):

PR #698, opened 2026-07-23
feat(advisory-deep-dive): per-advisory fix-audit verdicts, certificate and regression watch

### LINKED ISSUE #13: Fix audits for existing advisories
For each existing advisory, check whether the advertised fix actually fixed the vulnerability. Run the original repro, attempt bypasses, hunt variants.

- [ ] Fix-audit skill per advisory (SKILL.md + schema.json in its own directory under `skills/`)
- [ ] Output `{status: fixed | bypass | variant | regressed, evidence, finding?}`
- [ ] On `status=fixed` emit a public certificate; otherwise open a new finding
- [ ] Re-run on new releases (regression watch)
- [ ] Use results as a security-deep-dive calibration corpus

alexandre-daubois:
Fix #13

This records a per-advisory verdict for advisory-deep-dive, backs a public certificate when the fix held, and re-audits automatically on new releases.

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -419,6 +419,22 @@ Known security advisories from the `advisories` skill. Replaced each run.
 | withdrawn_at | datetime | Non-null if the advisory was withdrawn. |
 | created_at | datetime | |
 
+## advisory_audits
+
+Fix-audit verdicts from the `advisory-deep-dive` skill: for each published advisory, whether its advertised fix still holds at the audited commit. One row per advisory per run; the newest row per `(repository_id, advisory_uuid)` is authoritative. Keyed by advisory UUID rather than `advisories.id` because the `advisories` skill replaces its rows wholesale each run, which would orphan a foreign key.
+
+| Column | Type | Notes |
+|--------|------|-------|
+| id | integer PK | |
+| repository_id | integer FK | |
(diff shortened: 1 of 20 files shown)


andrew (changes requested):
`parseAdvisoryAuditOutput` returns early when `ingestFindings` errors, but `ingestFindings` returns `FailOnThresholdError` *after* persisting the findings, and the rest of the codebase treats that error as "finalized with findings committed" (see the comment on `maybeFireScanFinalized`). So with `fail_on` set on this skill, the run that finds a High-severity bypass persists the findings but never writes the `AdvisoryAudit` rows - the verdicts are lost. Knock-on: the scan isn't `done`, so the regression-watch baseline never advances and every subsequent `packages` scan re-enqueues the audit.

Fix: on `FailOnThresholdError`, write the audit rows first, then propagate the error.

Minor, take or leave: duplicate `advisory_uuid` entries in one report aren't rejected (last row wins via `id desc`), and `latestAdvisoryAuditStatuses` reads every historical audit row for the page's advisories rather than just the newest per pair.

alexandre-daubois:
Comments addressed!

andrew (approved):

Issue #14, opened 2026-04-20
Interchange and federation


andrew:
Federation between scrutineer instances and non-scrutineer tools.

- [x] In-toto Statement v1 envelope with a predicateType URI per record kind
- [x] `interchange.schema.json` shipped alongside scrutineer: certificate, claim, optout, route predicate schemas
- [x] Public feed (plain git repo): optout, route, clean certificates only
- [x] Members-only feed (SOPS/age-encrypted git): nonzero certificates
- [x] Claim-check endpoint: `POST /claim-check {hash} -> {match, contact}`
- [x] `finding_hash = sha256(shared_salt + repo + normalised_location + cwe)`
- [x] Export-feed and import-feeds jobs
- [ ] Outbound claim-check before status moves to reported
- [x] Never publish finding bodies, severity, CVSS, or health scores
- [ ] Sigstore signing for the public tier (later)

PR #708, opened 2026-07-24
feat(interchange): in-toto record envelope, salted finding hash and claim-check endpoint

alexandre-daubois:
Part of #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,7 @@ See [SECURITY.md](SECURITY.md) for the reporting policy and [threatmodel.md](thr
 - [docs/backup.md](docs/backup.md) -- backing up and restoring the database (built-in `scrutineer backup`/`restore`, `sqlite3`, Litestream)
 - [docs/development.md](docs/development.md) -- project layout, regenerating embedded data, running tests
 - [docs/encrypted-sharing.md](docs/encrypted-sharing.md) -- encrypted findings sharing between contributors (age + SSH keys, team keyring management)
+- [docs/interchange.md](docs/interchange.md) -- federation interchange format: in-toto record envelope, salted finding hashes, the claim-check endpoint
 - [docs/codex.md](docs/codex.md) -- the codex backend: what differs from claude, sandbox interaction, adding another harness
 - [docs/opencode.md](docs/opencode.md) -- the opencode backend: provider-agnostic credentials and egress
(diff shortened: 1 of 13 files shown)


andrew (changes requested):
Nice work — the anti-fingerprinting 404, `MaxBytesReader`, salt kept out of argv, `AssertFormat()`, and closed predicate schemas with tests that reject smuggled `severity`/`cvss_score` are all the right calls.

A few things before this lands:

1. The PR body says `Fix #14` but this covers ~5 of the 10 checkboxes (feeds, export/import jobs, outbound claim-check before `reported`, sigstore are all deferred). GitHub will auto-close #14 on merge. Change to `Part of #14`.

2. `canonicalCWE` doesn't dedupe: `"CWE-79,CWE-79"` and `"CWE-79"` hash differently. This is a wire contract shared across instances, so once hashes are published it can't change without a coordinated salt rotation. Add `slices.Compact` after the sort in `internal/interchange/hash.go` now while nothing depends on it, plus a test case in `TestFindingHashCWEList`.

3. `claimHashSet` scans every non-rejected/non-duplicate finding and computes a SHA256 per row, holding `claimIndex.mu` for the whole rebuild. That's fine at tens of thousands of rows with the 1-minute TTL; at much more the first request after expiry stalls every concurrent claim-check. @andrew — how many finding rows should this assume on a mature instance? If "small for now", a note in the `claimHashSet` doc comment that the strategy assumes a bounded table is enough.

4. Not blocking: the schema pins certificate `status` to `const: "fixed"`. #14's members-only feed carries nonzero certificates (bypass/variant/regressed from `advisory-deep-dive`), so the first non-clean certificate forces a `certificate/v2` predicate bump. If that's the plan, nothing to change.

alexandre-daubois:
About point 4: that's the plan. The const stays until the members-only feed actually emits nonzero certificates, at which point `certificate/v2` widens status to the full bypass/variant/regressed set. Widening it now would publish a shape nothing produces, and consumers would have to handle values they'll never see.

andrew (approved):
Thanks — all four points addressed. `slices.Compact` after the sort with the matching test case and doc update is exactly right, and the `claimHashSet` note names the replacement strategy so whoever hits the wall knows what to build.

Issue #118, opened 2026-04-30
Experiment: shorter audit prompts


andrew:
`security-deep-dive/SKILL.md` is ~140 lines of methodology. There's a reasonable hypothesis that past a certain length the model starts pattern-matching the checklist instead of reading the code, and that a terser prompt ("here is the repo, here is the threat model, find vulnerabilities in {focus_area}, validate each one before reporting") with the detailed sink taxonomy moved to a `references/` file it can pull on demand would do better.

Once #65 lands we can A/B this properly. Filing so it isn't forgotten.

andrew:
[openai/codex-security](https://github.com/openai/codex-security/tree/f22d4a36f26d16287bcdfd707b369116e02a08c3) is a partial data point for this. Its top-level `security-scan/SKILL.md` is 42 lines and hands off to `references/repository-wide-scan.md` (48 lines) and `references/scan-artifacts-and-ledger.md` (85 lines); the ledger, schema validation, dedup, and state mechanics are 30 Python helper scripts the skill calls rather than prose the model has to follow. `attack-path-analysis/SKILL.md` is 117 lines with the 153-line severity policy in `references/`. Not all of theirs are short though — `triage-finding` is 342 lines and `vulnerability-writeup` is 263.

The split that seems to hold across their set: the SKILL.md carries intent and phase order, `references/` carries policy tables and taxonomies pulled on demand, and anything that can be checked deterministically (schema shape, ledger reconciliation, receipt counts) is a script the skill runs rather than an instruction it might skip.

`security-deep-dive/SKILL.md` is now 276 lines, up from the ~140 when this was filed, so the A/B has more headroom than it did. The sink taxonomy and the report-shape rules are the obvious candidates to move out; the report shape in particular could be a `schema.json` + validator the skill runs at the end rather than 40 lines of "the report must contain...".

andrew:
#793 (reference-driven candidate + paired-experiment reporting) and #804 (deterministic eval tests in CI) both merged. Production `skills/security-deep-dive/SKILL.md` is still 276 lines and untouched.

Next step is the live A/B run and a promotion decision:

```
SCRUTINEER_RUN_EVALS=1 SCRUTINEER_EVAL_MODEL=claude-sonnet-5 go test -tags evals ./internal/evals/... -run TestRunFixtures -v
```

then commit the aggregate report and either promote the short variant or record why not.

PR #716, opened 2026-07-26
Shorten security deep dive prompt

abhinavgautam01:
Part of #118

## Summary

Shortens the always-loaded `security-deep-dive` prompt while preserving its core audit contract.

The detailed sink taxonomy, six-step checklist and inventory completeness guidance now live in a bundled reference file that the skill can consult when needed.

## Changes

- Reduce `skills/security-deep-dive/SKILL.md` to the core workspace, API, scope, method, finding, concurrency, fan-out and output rules
- Add `skills/security-deep-dive/references/audit-taxonomy.md` for detailed sink classes and per-sink audit procedure
- Keep mandatory report/schema requirements in the main skill prompt
- Preserve the inventory-first audit model and high-signal finding rules

## Tests

- `git diff --check`
- `go test ./internal/skills ./internal/worker -run 'TestBundledSchemas|TestSecurityDeepDive|TestParseBundled'`
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/skills/security-deep-dive/SKILL.md
+++ b/skills/security-deep-dive/SKILL.md
@@ -18,259 +18,167 @@ metadata:
 
 # security-deep-dive
 
-Audit the first-party source for security vulnerabilities. The target is this codebase's own code; do not report that a dependency has a CVE. A finding is valid only if the vulnerable logic lives here. If the same vulnerable code exists in a fork, a sibling project, or a vendored copy, note it; the finding follows the code.
-
-The audit has two phases. Phase 1 produces an inventory of every sink in the codebase. Phase 2 works through the inventory and decides on each entry. The inventory is part of the report, not scratch work — two runs against the same commit should produce the same inventory regardless of which sinks catch attention first.
-
-Workspace layout:
-- `./src` — the cloned repository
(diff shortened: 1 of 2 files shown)


andrew:
Closing rather than requesting changes because the shape of the PR doesn't match what #118 asks for, and iterating on this diff won't get there.

#118 describes an experiment: add a shorter prompt as a variant, A/B it against the current one using the #65 eval harness, and see which does better. This PR replaces the production prompt in place with no eval run, so there's no baseline left to compare against and no data showing the rewrite doesn't regress.

It also doesn't test the #118 hypothesis. The new SKILL.md tells the model to read `references/audit-taxonomy.md` before inventorying sinks, so both files load on every run — 384 lines in context vs the current 251. The idea in #118 was that the taxonomy sits in `references/` for the model to pull *when a sink class is unclear*, not always.

Separately, the rewrite drops content the rest of the system depends on:
- the `ruled_out` disposition-label vocabulary (`out_of_model_trusted_input`, `by_design_disclaimed`, etc.) that the threat workbench parses from `ruled_out[].reason` — see `internal/web/threat_workbench_test.go`
- the `boundaries[]` fill-from-threat-model mapping
- `provenance: inferred` vs `documented` handling
- the library-installed-gadget boundary case, agentic system-vs-user-role detail, the same-sink-different-boundary consolidation rule, and the `artefact` field format

If you want to pick this up again, the PR to open is one that adds the short prompt as a second variant alongside the current SKILL.md, runs both through the #65 `should_find`/`should_not_find` fixtures, and puts the comparison in the PR body. Happy to point at where the variant should live if that's useful.

PR #722, opened 2026-07-27
feat(chat): chat with model about a repo or finding

### LINKED ISSUE #16: Chat with claude about a repo or finding
Conversational view on a repo or finding, backed by claude with access to the clone and the database.

- [ ] Per-repo chat
- [ ] Per-finding chat
- [ ] Persisted conversations

alexandre-daubois:
Adds a section at the very bottom of findings to chat with a model about a finding:

Capture d’écran 2026-07-27 à 11 43 32

Mulitple conversations history support is implemented.

Fix #16


andrew (changes requested):
This is great, thank you. Merges cleanly onto current main and everything passes on the merged tree (build, vet, tests, golangci-lint, govulncheck, deadcode). The security posture is right: T5 strip on the chat clone with `threatmodel.md` updated, `IsolationKey` so a hardened turn stays off scan-N's network, `acceptEdits` dropped when there's no output file so `Read,Grep,Glob` is the actual boundary, `KindEgress` split out so proxy log lines can't become the answer on a killed turn, and the argv budget accounted for on both fresh and resume prompts.

Three things before merge:

- Rebase onto `main`. The branch is 18 commits behind; the merge is clean and tests pass on it, but rebasing keeps `internal/web/server.go` history linear through #700/#707 and lands the `docs/codex.md` note next to the newer harness wording.
- `chatSlots` is a separate pool sized from the queue rather than shared with it, so a host with `-concurrency 4` can end up running 4 scans plus 4 chat turns for 8 containers. Either cap it (something like `max(1, q.Concurrency()/2)`) or document the doubled ceiling wherever concurrency is covered.
- Conversations can't be deleted. Each one holds a full clone under `<datadir>/chat-<id>/src` plus a harness state dir, and the only cleanup path today is deleting the whole repository. A `DELETE /conversations/{id}` that calls `RemoveChatArtifacts` and drops the rows would close that.

alexandre-daubois:
Comments addressed. Endpoint added to delete conversations, cap and documentation added for point 2

andrew (approved):
Rebase, chat-pool cap, and per-conversation delete all landed as requested; green on current main (build, vet, tests, golangci-lint, govulncheck, deadcode). Thanks!

PR #746, opened 2026-07-27
add allow_remote config to permit non-localhost access

walterhpearce:
The web server enforces a localhost-only host-header check (DNS-rebinding protection): requests are served only when Host is 127.0.0.1, localhost, or ::1. That blocks deployments behind a reverse proxy or on a LAN.

Add an `allow_remote` config option (default false) that, when true, skips the host check so the UI + skill API can be reached for any Host. Defaults to false preserves the existing behavour.

- config: Config.AllowRemote (yaml: allow_remote)
- web: Server.AllowRemote gates securityHeaders' host check
- cmd/scrutineer: wire config through to the server (config-only, no flag)
- document in scrutineer.sample.yaml; add securityHeaders bypass test

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -119,6 +119,7 @@ type flags struct {
 	autoRejectMissedCount int
 	federationSalt        string
 	federationContact     string
+	allowRemote           bool
 	skillLocal            skillDirs
 
 	// set records which flags were passed on the command line so merge
@@ -277,6 +278,9 @@ func (f *flags) merge(cfg *config.Config) {
 	if cfg.FederationContact != "" && !f.set["federation-contact"] {
 		f.federationContact = cfg.FederationContact
 	}
+	// AllowRemote is config-only (no CLI flag); the zero value keeps the
+	// localhost-only host-header check on (remote access denied).
+	f.allowRemote = cfg.AllowRemote
 
 	// Seed the model pick list from the active harness's own defaults,
 	// so a fresh install of any backend has a working list with correct
@@ -560,6 +564,7 @@ func run(log *slog.Logger) error {
 	srv.SetDefaultEffort(f.effort)
 	srv.FederationSalt = f.federationSalt
 	srv.FederationContact = f.federationContact
+	srv.AllowRemote = f.allowRemote
 
 	if f.recipientsFile != "" {
 		recs, err := loadRecipients(f.recipientsFile)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -146,6 +146,15 @@ type Config struct {
 	// hash match. Required when FederationSalt is set: startup refuses a
 	// salt without a contact.
 	FederationContact string `yaml:"federation_contact"`
+	// AllowRemote permits remote (non-localhost) access to the web server. It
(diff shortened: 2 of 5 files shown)


andrew (changes requested):
Thanks for the PR. I want to be cautious here because this removes a documented mitigation rather than adjusting it.

A boolean `allow_remote` drops the DNS-rebinding defence entirely. For the reverse-proxy deployment — proxy on the same host forwarding to `127.0.0.1:8080` with a fixed `Host` — an `allowed_hosts: [scrutineer.internal]` list would serve the proxy and keep the rebinding check for everything else. The blanket bypass is only strictly needed when clients hit scrutineer directly by IP over a LAN, which is the deployment I'd rather not encourage: there's no authentication on the UI or the `/api/v1/*` JSONL exports, and "trusted network" is doing all the work.

If the goal is the reverse-proxy case, could this be reworked as `allowed_hosts` (list of extra hostnames accepted alongside the loopback set) instead of a boolean? That's a change I'd be comfortable merging.

Either way, before this can land:

- `threatmodel.md` needs updating. Boundary 1 (line 48) and T3 both state the Host check unconditionally, and `server.go` points readers at that file. An opt-out of a mitigation has to appear in the threat model.
- `scrutineer.sample.yaml`: the entry says "on a LAN" but `addr` still defaults to `127.0.0.1:8080`, and it doesn't mention that the unauthenticated `/api/v1/*` exports open up alongside the UI.
- `warnIfNonLoopbackListenAddr` will still fire when the operator has explicitly opted in; it should probably go quiet in that case.

Small: "behavour" in the PR description.

walterhpearce:
Would you be comfortable with the allowed_hosts expanding CIDRs as well, if I went with an allow list approach instead? 

My case is behind a GCP IAP proxy and load balancer, as well as some sporadic other systems calling the APIs, so I'd need to allow the full internal network(s). I wanted to avoid also having to deploy another reverse proxy locally kn the system just to bounce to it, when I'm already deploying an isolated system for all this with proxy rules on the LB. But if we can support full address expansion, I'm good with limiting it.

andrew:
Thanks, that context helps. `allowed_hosts` would match the request's `Host`, rather than the caller's source address. For the IAP load balancer, listing the hostname sent to the backend should be enough; GCP preserves the original Host and can also rewrite it to a fixed value. Direct internal callers could use an explicit Scrutineer hostname or IP.

If the CIDRs are intended to restrict which client networks may connect, I'd keep that in the GCP firewall/IAP configuration. Allowing source CIDRs in `allowed_hosts` would mix two separate controls.

Which endpoints are the other systems calling? `/api/` already skips this Host check and uses scan bearer tokens. `/api/v1/` is unauthenticated, so I'd want those calls routed through IAP.

I'd be comfortable with exact hostnames and IP literals in `allowed_hosts`, alongside the threat-model, sample-config, and startup-warning updates. I wouldn't add CIDR matching in this PR.

andrew:
@walterhpearce checking in on this — still keen to land it as `allowed_hosts` (exact hostnames and IP literals, no CIDR) plus the `threatmodel.md`/sample-config/startup-warning updates from the earlier review. Are you planning to pick it back up, or would you rather someone else take it from here?

PR #752, opened 2026-07-28
feat(interchange): certificate/v2 records and the on-disk record layout

alexandre-daubois:
Part of #14

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -29,6 +29,7 @@ import (
 	"scrutineer/internal/bundledassets"
 	"scrutineer/internal/config"
 	"scrutineer/internal/db"
+	"scrutineer/internal/interchange"
 	"scrutineer/internal/queue"
 	"scrutineer/internal/skills"
 	"scrutineer/internal/web"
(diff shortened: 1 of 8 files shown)


andrew (changes requested):
- `internal/interchange/feed.go:138`: encrypted files are retained solely when their plaintext matches. Changing `Recipients` therefore does not re-encrypt unchanged records. Added recipients cannot read the existing feed, while removed recipients retain access. Track recipient-set changes and force re-encryption, with add/remove rotation tests.

- `internal/interchange/feed.go:74`: an unknown tier with an empty record set bypasses every `Carries` check, is treated as plaintext, and proceeds to delete every managed record. Reject tiers other than `public` and `members` before inspecting or mutating the directory.

- `internal/interchange/feed.go:130`: reads and writes follow symlinks in the feed checkout. A symlinked kind directory or expected record file can make pruning or `os.WriteFile` delete or overwrite files outside the feed. Reject symlinked managed paths and keep all mutations beneath the resolved feed root.

andrew (changes requested):
The follow-up commit addresses the earlier static cases, but two issues remain:

1. `internal/interchange/feed.go:176-183`: recipient rotation rejects SSH recipients because the age SSH recipient types do not implement `fmt.Stringer`. Scrutineer supports SSH recipients and documents them as the default. Preserve a canonical fingerprint during parsing or wrap the recipients, then add a mixed SSH and X25519 rotation test.

2. `internal/interchange/feed.go:325-330`: record-shaped symlinks are silently skipped. `ReadFeed` can therefore omit a record without reporting an error, despite the documented symlink rejection. Surface these entries as errors and add a read-path regression test.

andrew (approved):
Both points addressed. The `Recipient` wrapper carrying the canonical SSH key (with `WrapWithLabels` forwarded and comments stripped) fixes rotation for the documented default; dropping the `IsRegular()` filter so record-shaped symlinks are listed lets `readRegular` refuse them and `os.Remove` unlink the entry on prune. Regression tests cover both.

PR #753, opened 2026-07-28
feat(web): honour a maintainer's request not to scan their repository

alexandre-daubois:
Part of #14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Entries are grouped by release, newest first. Each entry is a summary written fo
 
 ## Unreleased
 
+- A repository whose maintainer asks not to be scanned can now be marked as such, and no scan runs against it again: neither a manual run, nor the scheduler, nor an automatic follow-up. ([#754](https://github.com/alpha-omega-security/scrutineer/pull/754), [@alexandre-daubois](https://github.com/alexandre-daubois))
 - Scans can be scheduled to repeat automatically and are skipped when the code has not changed since the last run. ([#633](https://github.com/alpha-omega-security/scrutineer/pull/633), [@alexandre-daubois](https://github.com/alexandre-daubois))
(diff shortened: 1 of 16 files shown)


andrew (changes requested):
Changes needed before merge:

1. Setting the federation opt-out only updates the repository fields. Existing queued scans still run, running scans continue, and paused scans can be resumed through the direct queue paths in `internal/web/scans.go`. Stop existing work when the opt-out is set, and reject opted-out repositories in worker dispatch and resume paths to close races.

2. The scheduler checks the repository snapshot loaded at the start of the tick. If the opt-out is recorded after that query, `runScheduledScan` can still perform an upstream sync or remote HEAD request. Recheck and coordinate the current opt-out state before any network access.

3. The Unreleased changelog entry links to PR `#754`; this entry belongs to `#753`.

andrew (changes requested):
The follow-up commits address cancellation state and the recorded reason, but three races remain:

1. `internal/worker/worker.go:596-637`: the opt-out check and queued-to-running transition are not atomic. The sweep can cancel the queued row after the check, then `startScan` saves the stale row as running after the sweep has finished. Claim the row conditionally while the repository remains opted in, and add a regression test for this interleaving.

2. `internal/web/server.go:2791-2869`: enqueue checks the flag separately from creating the scan. A scan can be created after the opt-out sweep and reported as queued. Coordinate the repository check and scan creation with the opt-out update.

3. `internal/web/schedule.go:127-153`: the live check only narrows the race. An opt-out can commit immediately after the read, then the scheduler still contacts the upstream or resolves remote HEAD. Serialize this check-and-network section with recording the opt-out, and add a concurrent regression test.

andrew (approved):
All three races closed with the requested regression tests. The conditional-UPDATE claim in `startScan`, the insert-then-recheck in `enqueueSkillWith`, and the per-repo `lockRepoFederation` across the firing and the opt-out handler each match the interleaving they were asked to fix. Merges cleanly with the harness extraction on main.

PR #755, opened 2026-07-28
fix(web): reject non-boolean ?active= with 400 on GET /api/skills

### LINKED ISSUE #733: `GET /api/skills?active=` treats unparseable values as `false`
`apiListSkills` in `internal/web/api.go:497` reads `?active=` with `strconv.ParseBool` and discards the error:

```go
if v := r.URL.Query().Get("active"); v != "" {
    active, _ := strconv.ParseBool(v)
    q = q.Where("active = ?", active)
}
```

So `?active=yes` or `?active=typo` silently filters to inactive skills instead of returning 400. Return `400 Bad Request` when the value is present but not a valid bool.

charlie-morrison:
Closes #733.

`apiListSkills` parsed `?active=` with `strconv.ParseBool` and discarded the error, so `?active=yes` became `false` and returned the *inactive* skills — the opposite of what the caller asked for, with a 200.

```go
active, err := strconv.ParseBool(v)
if err != nil {
    writeAPIError(w, http.StatusBadRequest,
        fmt.Sprintf("active: must be true or false, got %q", v))
    return
}
```

The message mirrors `importRevalidate`'s `revalidate: must be true or false, got %q`, since it's the same situation and that comment already spells out the reasoning ("a malformed value is a 400 rather than a silent fall-back"). An empty `?active=` keeps its current meaning: no filter.

**Test.** `TestAPIListSkills_activeFilter` covers `true`, `false`, empty, and `yes`/`typo`/`2`. I ran it against the unfixed handler first to confirm it actually pins the bug — it fails there, and the failure output shows the reported symptom exactly (`?active=yes: status 200 ... rows=[... name:retired-skill]`).

One thing worth flagging for anyone writing similar tests: `Skill.Active` is `not null;default:true`, so `db.Skill{Name: "x", Active: false}` inserts **true** — GORM omits Go zero values and the column default wins. The retired fixture needs an explicit `Update("active", false)`. That cost me a confusing first run.

**Spec.** `openapi.yaml` documented only a `200` for this route, so I added the `400` (`$ref: BadRequest`, matching the other routes) and noted the empty-value behaviour on the parameter.

- `go test ./internal/web/ -count=1` — passes
- `gofmt -l`, `go vet ./internal/web/` — clean

Unrelated, spotted while validating my edit and **not touched in this PR**: `openapi.yaml` doesn't parse as YAML on `main`. Line 331 has an unquoted description containing `: ` —

```yaml
description: Optional baseline scan to use for `rescan_mode: diff`; omitted lets Scrutineer choose the latest compatible completed scan.
```

`rescan_mode: diff` makes it a mapping value, so `yaml.safe_load` fails there. Quoting the string fixes it. Happy to open a separate issue or PR if useful — it seemed like it might matter to #751.

### DIFF
--- a/internal/web/api.go
+++ b/internal/web/api.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"strconv"
@@ -495,7 +496,14 @@ func (s *Server) apiListCNAs(w http.ResponseWriter, r *http.Request) {
 func (s *Server) apiListSkills(w http.ResponseWriter, r *http.Request) {
 	q := s.DB.Order("name")
 	if v := r.URL.Query().Get("active"); v != "" {
-		active, _ := strconv.ParseBool(v)
+		// A malformed value is a 400 rather than a silent fall-back to false,
+		// so ?active=yes never quietly returns the inactive skills instead.
+		active, err := strconv.ParseBool(v)
+		if err != nil {
+			writeAPIError(w, http.StatusBadRequest,
+				fmt.Sprintf("active: must be true or false, got %q", v))
+			return
+		}
 		q = q.Where("active = ?", active)
 	}
 	var rows []db.Skill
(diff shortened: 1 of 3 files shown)


andrew (approved):
The invalid active filter now returns 400, with regression coverage and the documented API response.

PR #761, opened 2026-07-28
Add host-only API delete endpoints

### LINKED ISSUE #735: Add repository and finding delete under `/api/v1/`
Nothing on either API can delete a repository or a finding; deletion only exists as `POST /repositories/{id}/delete` on the web UI. That is deliberate for the scan-token API (a scan running against a hostile repo should not be able to delete records), but it also means a cron job or CI script cannot clean up without form-posting to the browser routes.

Add `DELETE /api/v1/repositories/{id}` and `DELETE /api/v1/findings/{id}` on the host-only `/api/v1/` group, reusing the same code paths the web UI handlers already call. The scan-token API stays read/update only.

abhinavgautam01:
Fixes #735

## Summary

Adds host-only `/api/v1` delete endpoints so local automation can clean up repositories and findings without form-posting to browser routes.

The scan-token API remains read/update only; these delete routes live on the localhost-only `/api/v1` surface.

## Changes

- Add `DELETE /api/v1/repositories/{id}`
- Add `DELETE /api/v1/findings/{id}`
- Reuse the existing repository delete cleanup path for API repository deletion
- Add finding deletion cleanup for:
  - child notes, communications, references, history and reviews
  - exposure rows
  - label join rows
  - finding-scoped chat conversations
- Preserve scans when deleting a finding by clearing `scan.finding_id`
- Update OpenAPI documentation
- Add tests for both delete endpoints

## Tests

- `go test ./internal/web -run 'TestAPIv1Delete|TestRepoDelete|TestExportRejectsBadHost'`
- `go test ./...`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/internal/web/api_export.go
+++ b/internal/web/api_export.go
@@ -47,6 +47,10 @@ func (s *Server) apiDeleteRepository(w http.ResponseWriter, r *http.Request) {
 	}
 	deleted, err := s.deleteRepository(repo)
 	if err != nil {
+		if errors.Is(err, errRepositoryDeleteInFlight) {
+			writeAPIError(w, http.StatusConflict, err.Error())
+			return
+		}
 		writeAPIError(w, http.StatusInternalServerError, err.Error())
 		return
 	}
--- a/internal/web/api_export.go
+++ b/internal/web/api_export.go
@@ -41,7 +41,7 @@ func (s *Server) exportHandler() http.Handler {
 }
 
 func (s *Server) apiDeleteRepository(w http.ResponseWriter, r *http.Request) {
-	repo, ok := loadExportRowByID[db.Repository](s, w, r, "repository")
+	repo, ok := s.loadExportRepositoryByID(w, r)
 	if !ok {
 		return
 	}
@@ -55,7 +55,7 @@ func (s *Server) apiDeleteRepository(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *Server) apiDeleteFinding(w http.ResponseWriter, r *http.Request) {
-	finding, ok := loadExportRowByID[db.Finding](s, w, r, "finding")
+	finding, ok := s.loadExportFindingByID(w, r)
 	if !ok {
 		return
 	}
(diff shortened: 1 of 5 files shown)


andrew (changes requested):
Please address these issues before merging:

1. `deleteFinding` clears `finding_id` for queued, running, and paused scans. Queued scans can then run without their required finding context. A running worker retains the old ID in memory and later attempts to save it after the finding has been deleted, causing an FK failure and potentially leaving the scan stuck as running. Reject deletion while linked scans are in flight, or safely cancel and finish them first. Add regression tests for queued and running scans.

2. `loadExportRowByID` fails the required `ireturn` lint check. Change it to populate a typed destination pointer, or use concrete loaders, instead of returning generic `T`.

abhinavgautam01:
Fixed both points.

`deleteFinding` now refuses to delete a finding while any linked finding-scoped scan is queued, running or paused and the API maps that to `409 Conflict` instead of detaching the scan. Completed linked scans are still preserved by clearing `finding_id`.

I added regression coverage for queued and running linked scans to assert the delete is rejected and both the finding row and scan `finding_id` remain intact. The previous generic-loader lint issue is also resolved and the CI golangci-lint version reports `0 issues`.

andrew (changes requested):
Both earlier points are addressed (concrete loaders, in-flight check + 409 + tests). Three remaining:

- The in-flight count runs on `s.DB` before `s.DB.Transaction(...)` (`server.go:2680-2687`). A scan enqueued between the check and the `Update("finding_id", nil)` gets its `finding_id` cleared while queued — the case the guard exists to prevent. Move the `Count(&inFlight)` (and the `Pluck` for `ConversationIDs`) inside the transaction and return `errFindingDeleteInFlight` from there.
- `DELETE /v1/findings/{id}` in `openapi.yaml` is missing the `409` response the handler now returns.
- `deleteRepository` is extracted here and exposed at `DELETE /api/v1/repositories/{id}` for automation, but has no in-flight guard — a cron hitting it while a scan is running deletes the scan row from under the worker. Add the same queued/running/paused check → 409, inside the transaction, and document 409 on the repository delete in openapi too.

abhinavgautam01:
Thanks, fixed

- Moved the finding delete in-flight count and conversation ID pluck inside the delete transaction.
- Added the same queued/running/paused guard to repository delete inside its transaction.
- Repository delete now returns `409 Conflict` instead of deleting while repo-linked or finding-scoped scans are in flight.
- Documented the `409` responses for both delete endpoints in OpenAPI.
- Added regression coverage for queued/running repository scans and a paused finding-scoped scan.

andrew (approved):
All three from the last round addressed: in-flight checks and plucks moved into the transaction on both delete paths, 409 documented on both endpoints, and the repository guard covers finding-scoped scans on other repos. Browser `repoDelete` picking up the guard is a fix. Thanks for the thorough test coverage.

PR #762, opened 2026-07-28
Add finding migration guide

abhinavgautam01:
Part of #12

## Summary

Adds a finding-page migration guide for zombie or abandoned package repositories.

This builds on repository health scoring and package alternatives by showing operators the packages consumers should migrate away from, curated replacement targets and dependent exposure rows that can drive migration outreach.

## Changes

- Add a Migration guide panel on finding pages for:
  - abandoned repositories
  - zombie repositories
  - repositories with curated package alternatives
- Show repository health and health evidence
- List published packages ordered by downstream reach
- List operator-curated forks, successors, or equivalent package alternatives
- Show affected or under-investigation dependents for consumer prioritization
- Hide the panel for healthy repositories without alternatives
- Add focused tests for rendering and non-rendering cases

## Tests

- `go test ./internal/web -run 'TestFindingShowMigrationGuide|TestRepoPackageAlternatives|TestFindingExposure'`
- `go test ./...`
- `go test -tags evals ./...`
- `go vet ./...`
- `go vet -tags evals ./...`
- `git diff --check`


internal/web/finding_migration.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func loadFindingMigrationGuide(gdb *gorm.DB, f db.Finding, repo db.Repository) (*findingMigrationGuide, error) {
+	alternatives, err := loadPackageAlternatives(gdb, repo.ID)
+	if err != nil {
+		return nil, err
+	}
+	show := len(alternatives) > 0 ||
+		repo.Health == db.RepositoryHealthAbandoned ||
+		repo.Health == db.RepositoryHealthZombie
```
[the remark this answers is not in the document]
abhinavgautam01: The visibility check now uses the current `AssessRepositoryHealth` result before deciding whether to render the migration guide, so archived repositories with an empty persisted `repo.Health` still render as abandoned/zombie without requiring a prior health refresh. I added coverage by leaving the archived test repo’s stored health empty.

internal/web/finding_migration.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+
+	var dependents []db.Dependent
+	if err := gdb.Where("id IN ?", dependentIDs).
+		Order("dependent_repos desc, downloads desc, id asc").
+		Limit(migrationGuideRowLimit).
+		Find(&dependents).Error; err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: The migration guide now normalizes an empty exposure status to `under_investigation`, filters out `known_not_affected` rows before applying the top-N dependent limit, and only prioritizes actionable dependents. I added a regression case where the highest-reach rows are all known-not-affected and lower-reach affected/under-investigation rows still appear in the guide.

internal/web/finding_migration.go
```diff
@@ -0,0 +1,130 @@
...
+	if len(exposureRows) == 0 {
+		return nil
+	}
+
+	actionableIDs := make([]uint, 0, len(exposureRows))
+	statusByDependent := make(map[uint]db.FindingDependent, len(exposureRows))
+	for _, row := range exposureRows {
+		row.Status = migrationExposureStatus(row.Status)
+		statusByDependent[row.DependentID] = row
+		if row.Status == db.ExposureKnownNotAffected {
```
alexandre-daubois, line 80: `actionableIDs` only excludes `known_not_affected`, so a dependent the exposure skill returned as fixed lands in Consumer prioritization, while the copy right below claims that table only holds affected or under-investigation rows. You should exclude `db.ExposureFixed` too, and decide whether it feeds `KnownSafeCount` or gets its own number in that sentence.

internal/web/templates/finding_show.html
```diff
@@ -306,6 +306,104 @@ <h3 class="flex flex-1 items-center justify-between gap-4 py-4 text-sm font-medi
   </details>
   {{end}}
 
+  {{if .MigrationGuide}}{{$g := .MigrationGuide}}
```
alexandre-daubois, line 309: This adds a user-facing panel with no `## Unreleased` line in CHANGELOG.md and no mention in the README's Findings bullet (line 194) that lists what a finding page shows. Please add both here, and re-check the changelog PR link now that the number is known.

internal/web/finding_migration.go
```diff
@@ -0,0 +1,130 @@
...
+			guide.KnownSafeCount++
```
alexandre-daubois, line 81: `KnownSafeCount` / `TotalExposureRows` and the `{{else if $g.TotalExposureRows}}` sentence have no test: the existing case always takes the `{{if $g.PriorityDependents}}` arm. Add a case where every exposure row is non-actionable so that sentence and its two numbers get asserted; it's the case the fixed handling above gets wrong.

internal/web/finding_migration.go
abhinavgautam01: `fixed` exposure rows are now treated as non-actionable and excluded from Consumer prioritization, alongside `known_not_affected`. I kept them separate from `KnownSafeCount` and added a dedicated fixed count so the summary copy stays accurate.

internal/web/templates/finding_show.html
abhinavgautam01: Added an Unreleased changelog entry for #762 and updated the README Findings bullet to mention migration guidance on finding pages.

internal/web/finding_migration.go
abhinavgautam01: Added a regression case where every exposure row is non-actionable: one `known_not_affected` and one `fixed`. The test asserts the no-actionable summary sentence and both counts.

andrew (changes requested):
The panel is the right shape for #12; the data loading needs rework so it doesn't re-query and recompute what `findingShow` already has.

- `loadFindingMigrationGuide` recomputes health on every GET via `db.AssessRepositoryHealth`, fed by a `Limit(10)` packages query and an unbounded maintainers query. `RefreshRepositoryHealth` (the health job) feeds it *all* packages, so this recomputed classification can disagree with the stored `repo.Health`. Use `repo.Health` directly. If the evidence `Summary` string is wanted, derive it in `internal/db` from the already-loaded `repo` (or store it alongside `Health`); don't add two queries per finding-page view.
- `finding_dependents` and `dependents` are re-queried at `finding_migration.go:78-80,104-108` but `server.go:1744-1756` already loaded both for this request. Pass `fdRows` and the resolved dependent map in from the handler.
- The `show :=` gate at `finding_migration.go:57-60` duplicates `showPackageAlternatives` (`package_alternatives.go:120-122`). Call that instead.
- `finding_migration.go:113-115` re-checks `known_not_affected`/`fixed` inside the second loop, but `actionableIDs` already excluded those — the branch is dead. Drop it.
- Drop the README edit; the Findings bullet lists the page's core sections and this panel is conditional. The CHANGELOG entry covers it.

abhinavgautam01:
Thanks, fixed

- `loadFindingMigrationGuide` now uses the persisted `repo.Health` instead of recomputing health from a limited packages query plus a maintainer query.
- Reused the `findingShow` exposure rows and dependent lookup map instead of querying `finding_dependents` / `dependents` again.
- Switched the panel gate to `showPackageAlternatives`.
- Removed the dead fixed / known-not-affected branch from the prioritized dependent loop.
- Dropped the README edit; the CHANGELOG entry remains.

andrew (changes requested):
Thanks, all five points from the last round are addressed. Two more before this lands:

- `finding_migration.go:48-56` — the `packages` query runs before the `showPackageAlternatives` gate. That gate only needs `repo` and `alternatives`, so every finding page on an active/stale repo with no alternatives pays a packages query for nothing. Move `if !showPackageAlternatives(repo, alternatives) { return nil, nil }` above the packages query.
- `HealthSummary` is now dead — nothing sets it after the recompute was removed. Drop the struct field and the `{{if $g.HealthSummary}}` branch in `finding_show.html`.

abhinavgautam01:
Thanks, fixed

- Moved the `showPackageAlternatives` gate before the packages query so active/stale repos without alternatives avoid the extra package lookup.
- Removed the dead `HealthSummary` field and the unused template branch.

andrew (approved):
The two requested updates are addressed. The alternatives gate now avoids the unused package query, and the dead health-summary path is removed.

PR #770, opened 2026-07-29
Bound repository dependency API results

### LINKED ISSUE #713: Add `limit` and `offset` pagination to repository read APIs
### Summary

Several repository-scoped read APIs return all matching rows in one response. Add optional `limit` and `offset` query parameters to keep responses bounded for large repositories.

### Details

In `internal/web/api_reads.go`, the following handlers currently load all matching rows:

- `apiListMaintainers`
- `apiListPackages`
- `apiListAdvisories`
- `apiListDependents`
- `apiListDependencies`

For example:

```go
s.DB.Where("repository_id = ?", id).Order("dependent_repos desc").Find(&rows)
```

Desired usage:

```text
GET /api/repositories/1/dependencies?limit=100&offset=0
GET /api/repositories/1/dependents?limit=100&offset=200
```

### Impact

Large repositories can have thousands of dependencies, dependents, advisories, packages or maintainers. Returning all rows at once increases memory use in the web server and forces skill clients to process more JSON than needed.

### Suggested Fix

Add shared query parsing for pagination parameters:

- `limit` optional, with a conservative maximum cap.
- `offset` optional, defaulting to `0`.
- Reject negative or non-numeric values with `400 Bad Request`.
- Preserve current behavior when parameters are absent, or define a default cap if maintainers prefer bounded responses by default.

Apply the helper to repository read endpoints and update tests/OpenAPI docs.

snowyukitty:
## Summary

- add a shared `parseLimit` helper for repository dependency reads, defaulting to 200 rows and capping requests at 1000
- apply the bound to the dependents and dependencies queries, returning `400 Bad Request` for empty, non-numeric, or negative values
- document the query parameter and cover default, explicit, capped, zero, and invalid limits

## Testing

The Linux-target checks below ran against an LF snapshot of the exact candidate tree:

- `go mod tidy -diff`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go vet ./...`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go vet -tags podman ./...`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/web`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 golangci-lint run ./...` (0 issues)

The three targeted runtime tests also passed locally on Windows with a temporary process-control compatibility shim for upstream's Unix-only worker code; the shim is not part of this patch.

Closes #713.

Disclosure: Prepared with OpenAI Codex assistance; the resulting patch and validation evidence were reviewed before submission.

### DIFF
--- a/internal/web/api_reads.go
+++ b/internal/web/api_reads.go
@@ -11,10 +11,31 @@ import (
 	"scrutineer/internal/worker"
 )
 
+const (
+	repositoryReadDefaultLimit = 200
+	repositoryReadMaxLimit     = 1000
+)
+
 // The read endpoints below expose the structured rows scrutineer already
 // populates from prior skill scans. Skills that need context for a repo
 // (verify/patch/disclose, security-deep-dive's reach and prior-art steps)
 // call these instead of re-parsing the original scan reports.
+
+func parseLimit(r *http.Request, defaultN, maxN int) (int, error) {
+	query := r.URL.Query()
+	if !query.Has("limit") {
+		return defaultN, nil
+	}
+	raw := query.Get("limit")
+	n, err := strconv.Atoi(raw)
+	if err != nil || n < 0 {
+		return 0, errors.New("limit must be a non-negative integer")
+	}
+	if n > maxN {
+		n = maxN
+	}
+	return n, nil
+}
 
 // repoScopedID parses the path id and enforces the scan-owns-repo auth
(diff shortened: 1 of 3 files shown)


andrew (approved):
Matches the rescoped #713 exactly — helper, two endpoints, default/cap, tests, OpenAPI. Thanks for the thorough test coverage.

PR #775, opened 2026-07-29
Serve openapi.yaml at GET /api/openapi.yaml

### LINKED ISSUE #731: Serve openapi.yaml over HTTP
`openapi.yaml` is only in the repo. Serve it at `GET /api/openapi.yaml` so a skill or an external tool can fetch the spec from a running server without a checkout. Embed the file (like the templates and static assets) so it ships in the binary.

charlie-morrison:
Fixes #731.

Embeds the repo-root `openapi.yaml` and serves it from a running server, so a skill or an external tool can read the API surface without a checkout.

## Where it is registered, and why not in `apiHandler`

`apiHandler` wraps every route in `apiAuth`, which only accepts a bearer token belonging to a **currently running scan**. A tool discovering the API has no scan to take a token from, so mounting the spec there would have defeated the point of serving it — the first version of the test showed exactly that, a `401` on `GET /api/openapi.yaml`.

So it is registered on the root mux as `GET /api/openapi.yaml`, which is more specific than the existing `"/api/"` pattern and therefore wins the match. It is wrapped in `securityHeaders`, so it keeps the **localhost-only host check** the browser UI and the `/api/v1` export surface rely on. Reading it is unauthenticated, which discloses nothing new: the document is already public in this repository. This mirrors how `/claim-check` sits on the root mux while still being documented in the spec.

## The embed needed a root package

`//go:embed` cannot reach a parent directory, and `openapi.yaml` lives at the repo root while the server is in `internal/web`. Rather than move the file — which would break the `../../openapi.yaml` read in the new coverage test from #751, plus docs references — this adds a tiny root package (`openapi.go`, `package scrutineer`) that embeds it and exposes `OpenAPISpec`.

## Spec and coverage test

- Documented the new path in `openapi.yaml` with `security: []`, and bumped `info.version` to `0.4.0`. The server base is `/api`, so the path key `/openapi.yaml` resolves to `/api/openapi.yaml`.
- Added the route to `registeredOpenAPIRoutes` in the #751 coverage test, next to the existing `/claim-check` entry, since root-mux registrations are not auto-discovered from `apiHandler`/`exportHandler`.

## Tests

Written against the unfixed code first and confirmed failing (`401`/`403` mismatch), then passing:

- `TestOpenAPISpecServedOverHTTP` — 200, `Content-Type: application/yaml`, parses as YAML, documents its own path.
- `TestOpenAPISpecMatchesRepoFile` — served bytes are identical to `openapi.yaml` on disk, so a stale embed cannot silently serve a spec that disagrees with the repo.
- `TestOpenAPISpecRejectsNonLocalHost` — a non-local `Host` still gets 403.
- `TestOpenAPISpecNeedsNoBearerToken` — pins the reason it is not behind `apiAuth`.

`go build ./...`, `go vet ./...` and the full `go test ./internal/web` suite pass locally.

Happy to move it behind `apiAuth` or change the content type if you would rather have it a different way.

### DIFF
--- a/internal/web/openapi.go
+++ b/internal/web/openapi.go
@@ -6,17 +6,45 @@ import (
 	"scrutineer"
 )
 
-// openAPISpec serves the embedded openapi.yaml.
+// openAPISpecHandler serves the embedded openapi.yaml under two access rules,
+// because the two callers the document exists for reach the server very
+// differently.
 //
-// It is registered on the root mux rather than inside apiHandler because
-// apiAuth only accepts a bearer token belonging to a RUNNING scan: a tool
-// discovering the API has no scan to borrow a token from, so serving the spec
-// from behind that middleware would defeat the point of serving it at all.
-// Reading it is unauthenticated but still host-only via securityHeaders, the
-// same boundary the /api/v1 export surface relies on (see threatmodel.md).
-// The document is already public in the repository, so this discloses nothing
-// a reader could not fetch from GitHub.
-func (s *Server) openAPISpec(w http.ResponseWriter, r *http.Request) {
+// A caller on the host -- the browser UI, curl, a tool discovering the API --
+// gets it unauthenticated. It has no running scan to borrow a bearer token
+// from, so requiring one would defeat the point of serving the spec at all;
+// the Host check is what keeps a DNS-rebound browser out, the same boundary
+// the /api/v1 export surface relies on (see threatmodel.md).
+//
+// A skill inside the runner container cannot satisfy that check. context.json
+// advertises the runtime's host endpoint (host.docker.internal for
+// docker/podman, the default gateway IP under Apple's container), and the
+// egress proxy rewrites the dial target to loopback but forwards the original
+// Host, so the request arrives with a non-local Host and would 403. Those
+// callers instead present the per-scan bearer token they already hold for the
+// rest of /api. That is not a weaker boundary: a rebound browser cannot mint a
+// running scan's token, and it cannot attach an Authorization header
+// cross-origin without a preflight this server never answers.
+//
+// The document is already public in the repository either way, so neither path
+// discloses anything a reader could not fetch from GitHub.
+func (s *Server) openAPISpecHandler() http.Handler {
+	spec := http.HandlerFunc(s.openAPISpec)
+	fromHost := securityHeaders(spec)
+	fromContainer := s.apiAuth(spec)
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if localHost(r.Host) {
+			fromHost.ServeHTTP(w, r)
+			return
+		}
+		w.Header().Set("Content-Security-Policy", cspPolicy)
+		fromContainer.ServeHTTP(w, r)
+	})
+}
+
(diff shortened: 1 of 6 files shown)


andrew (changes requested):
The new route is wrapped only in `securityHeaders`, which rejects the `host.docker.internal` host advertised in `context.json` to the default containerised runner. The egress proxy rewrites the dial target to loopback but preserves the original HTTP `Host`, so a skill fetching `{api_base}/openapi.yaml` receives 403. Please allow bearer-authenticated container access while retaining unauthenticated localhost access, and add a regression test using the container API host.

charlie-morrison:
Thanks — you're right, and the 403 was reachable exactly where the issue wants the endpoint to work. `securityHeaders` was the only wrapper, so the `host.docker.internal` (or gateway-IP, under Apple's `container`) `Host` that `api_base` advertises got rejected before the handler ran.

Pushed 43ad9ff, which picks the access rule from the caller rather than widening the host check:

- **Local `Host`** — unchanged: unauthenticated, host-only, same boundary as `/api/v1`.
- **Non-local `Host`** — goes through `apiAuth` and must present a bearer token belonging to a **currently running scan**, i.e. the credential the container already holds for the rest of `/api`. I preferred this over allowlisting `host.docker.internal` because the Apple runtime advertises a gateway IP, so a name allowlist wouldn't cover it — and an allowlist would hand the document to a rebound browser, where a token can't be.

Rebinding is still covered on that path: a browser can't mint a running scan's token, and it can't attach `Authorization` cross-origin without a preflight this server never answers.

Host parsing moved into a `localHost` helper so both paths share one definition of "local"; `securityHeaders` is otherwise untouched.

**Regression test** (`TestOpenAPISpecServedToContainerWithScanToken`) uses both container API hosts — `host.docker.internal:8080` and `192.168.64.1:8080` — with a running scan token, and asserts the served bytes are the spec. I ran it against the pre-fix handler first and it fails with your symptom on both:

```
--- FAIL: TestOpenAPISpecServedToContainerWithScanToken
    Host host.docker.internal:8080 with a running scan token: status = 403, want 200; body=forbidden: invalid host
    Host 192.168.64.1:8080 with a running scan token: status = 403, want 200; body=forbidden: invalid host
```

`TestOpenAPISpecNonLocalHostNeedsValidToken` pins the other side — no token, a garbage token, and a *finished* scan's token all get 401 from a non-local host and never see the document.

**One behaviour change worth flagging:** a tokenless non-local request now gets **401** instead of 403, since the request is no longer refused on host grounds alone. `openapi.yaml` documents the endpoint as optional-auth (`security: [{}, {bearerAuth: []}]`) with the 401 in place of the old 403. Happy to keep it a 403 instead if you'd rather the response not distinguish the two cases.

Also took the Copilot nit — the served/on-disk comparison is `bytes.Equal` now, no string conversion.

`go build ./...`, `go vet ./...` and the full `go test ./internal/web` (117s) all pass.

andrew (approved):
Container-access gap closed: skills reach the spec with their per-scan bearer via `host.docker.internal` / the Apple gateway IP, localhost stays unauthenticated, and a rebound browser without a token gets 401. `localHost` extraction leaves `securityHeaders` unchanged. Tests cover both container hosts and the no/garbage/done-scan token cases.

PR #778, opened 2026-07-29
feat(interchange): export and import the federation feeds over git

alexandre-daubois:
Part of #14


andrew (changes requested):
Three issues need changes:

- `internal/web/federation.go:292`: An applied opt-out is skipped permanently. If its repository is deleted and re-added, the new row remains scannable despite the active peer opt-out. Track which repository row received the application, or reopen matching records during deletion.
- `internal/web/federation.go:568`: Existing export clones always use their original `origin`. Changing a configured feed remote and restarting continues publishing to the old remote while leaving the new one empty. Check and update the clone origin before fetching.
- `internal/web/federation.go:412`: A corrected route clears `applied_at`, but the old imported channel makes this branch skip the update. The corrected record is archived and stamped without taking effect. Replace prior unstamped routes from the same feed while preserving analyst-owned values.

andrew (approved):
All three review points addressed:

- **Opt-out survives repo delete/re-add** — `InterchangeRecord.AppliedRepositoryID` records which row the stamp acted on; `reopenRepoInterchangeRecords` clears it inside the delete transaction so the next import re-applies. `TestRepoDelete_reopensTheRecordsAppliedToIt` walks the full delete→re-add→re-import cycle.
- **Export clone follows a changed remote** — `retargetOrigin` runs before `fetch --prune` and rewrites `origin` when it differs. `TestExportFeed_followsAChangedRemote` exports to `old` then `moved` and asserts only `moved` keeps receiving.
- **Corrected route replaces prior unstamped route** — `replaceableRoute` allows overwrite when the channel is empty or an unstamped same-feed hint; the write is conditional on the read value and only closes when `RowsAffected > 0`. `TestImportFeed_peerCorrectsItsOwnRoute` and `TestImportFeed_routeDoesNotClobberAConcurrentLocalEdit` cover both the correction and the race.

Needs a rebase before merge though — main moved a lot today (#784/#785/#788/#789/#790/#794/#795) and 11 of the 18 files here now conflict (CHANGELOG, README, main.go + test, config.go, server.go, skill.go, db.go, docs/database.md, threatmodel.md, sample.yaml). Should all be adjacent-addition trivial, but there are enough sites that I'd rather you resolve them than me guess.

alexandre-daubois:
Here you go!

PR #788, opened 2026-07-30
Reject a labels body with no array instead of clearing all labels

### LINKED ISSUE #710: `PUT /api/findings/{id}/labels` treats omitted or null labels as clear-all
### Summary

The labels API endpoint accepts a request body where `labels` is omitted or explicitly set to `null` and treats that the same as an intentional empty label list. This can clear all labels from a finding without the client explicitly sending `"labels": []`.

### Details

In `internal/web/api_finding_writes.go`, `apiSetFindingLabels` decodes the request body into:

```go
var body struct {
    Labels []string `json:"labels"`
}
```

If a caller sends `{}` or `{"labels": null}`, `body.Labels` remains `nil`. The handler then calls:

```go
db.SetFindingLabels(s.DB, id, body.Labels)
```

`db.SetFindingLabels` treats the provided slice as the complete replacement set, so a nil slice clears the association. The helper also trims label names and silently skips blank entries, so payloads such as `{"labels":["wontfix", ""]}` are accepted even though one item is invalid.

This behavior is different from a more explicit API contract where:

- Missing `labels` should be rejected as malformed input.
- `labels: null` should be rejected because the endpoint expects an array.
- Blank label names should be rejected instead of silently ignored.
- `labels: []` should remain the explicit way to clear all labels.

### Impact

API clients can accidentally clear labels by sending an incomplete payload. This makes label updates less predictable and can hide client-side bugs because the endpoint returns `204 No Content` even when the request did not explicitly provide a valid label set.

### Suggested Fix

Validate the decoded request body before calling `db.SetFindingLabels`:

- Distinguish omitted/null labels from an intentionally empty array.
- Require `labels` to be a JSON array.
- Reject blank labels with `400 Bad Request` or `422 Unprocessable Entity`.
- Keep `labels: []` as the explicit clear-all operation.

charlie-morrison:
Fixes #710.

`PUT /api/findings/{id}/labels` decoded `{}` and `{"labels": null}` to a nil slice and passed it straight to `db.SetFindingLabels`, which replaces the finding's label set with whatever it receives. So a malformed body wiped every analyst-set label **and answered 204** — the client bug looked like a successful update.

Scoped to exactly what @andrew laid out in the issue thread:

- Reject `body.Labels == nil` with **400**, reusing the decoder's existing `"body must be JSON with a labels array"` message.
- `{"labels": []}` still clears, so callers keep the deliberate way to empty the set.
- **Blank-label handling left alone** — `SetFindingLabels` trimming and skipping blanks is forgiving-input behaviour and out of scope, as you said.

**The browser UI is unaffected.** Its "Save labels" form posts to `/findings/{id}/labels` and is served by `finding_forms.go`, not this JSON endpoint, so unchecking every box still clears as before. Worth stating explicitly since "reject the empty case" could plausibly have broken the one UI path that legitimately sends nothing.

**Verification.** `TestAPISetFindingLabels_rejectsMissingLabelsArray` seeds two labels, then sends `{}`, `{"labels":null}` and `{"other":"field"}`, asserting 400 *and* that the two labels survive — the silent data loss is the actual bug, so the test pins the data, not just the status. It ends by confirming the explicit `[]` clear-all still returns 204 and empties the set. Run against the unfixed handler first, it fails with the reported symptom on all three:

```
--- FAIL: TestAPISetFindingLabels_rejectsMissingLabelsArray
    PUT {}: status = 204, want 400
    PUT {} cleared labels: have 0, want the 2 already set
    PUT {"labels":null}: status = 204, want 400
    PUT {"labels":null} cleared labels: have 0, want the 2 already set
    PUT {"other":"field"}: status = 204, want 400
    PUT {"other":"field"} cleared labels: have 0, want the 2 already set
```

Also documented the 400 in `openapi.yaml` and made the `labels` description state that `[]` is the clear-all and that omitting the field is rejected — it previously only said "Missing labels are created on demand", which reads as being about label rows, not the field.

`go build ./...`, `go vet ./...`, and the full `go test ./internal/web ./internal/db` all pass.

### DIFF
--- a/internal/web/api_finding_writes.go
+++ b/internal/web/api_finding_writes.go
@@ -160,6 +160,14 @@ func (s *Server) apiSetFindingLabels(w http.ResponseWriter, r *http.Request) {
 	if !ok {
 		return
 	}
+	// {} and {"labels": null} both decode to a nil slice, which
+	// SetFindingLabels would apply as an intentional replacement with the
+	// empty set: a malformed body would silently wipe analyst-set labels and
+	// still answer 204. An explicit [] stays the way to clear (#710).
+	if body.Labels == nil {
+		writeAPIError(w, http.StatusBadRequest, "body must be JSON with a labels array")
+		return
+	}
 	if err := db.SetFindingLabels(s.DB, id, body.Labels); err != nil {
 		writeAPIError(w, http.StatusInternalServerError, err.Error())
 		return
--- a/internal/web/api_finding_writes.go
+++ b/internal/web/api_finding_writes.go
@@ -163,7 +163,9 @@ func (s *Server) apiSetFindingLabels(w http.ResponseWriter, r *http.Request) {
 	// {} and {"labels": null} both decode to a nil slice, which
 	// SetFindingLabels would apply as an intentional replacement with the
 	// empty set: a malformed body would silently wipe analyst-set labels and
-	// still answer 204. An explicit [] stays the way to clear (#710).
+	// still answer 204. Clearing stays available to callers that ask for it
+	// with a present array — an explicit [], or one whose names are all
+	// blank, since SetFindingLabels trims and skips blank names (#710).
 	if body.Labels == nil {
 		writeAPIError(w, http.StatusBadRequest, "body must be JSON with a labels array")
 		return
--- a/internal/web/api_finding_writes_test.go
+++ b/internal/web/api_finding_writes_test.go
@@ -263,6 +263,43 @@ func TestAPISetFindingLabels(t *testing.T) {
 	}
 }
 
+// A body that omits labels, or sends null, must not be read as "replace with
+// the empty set": that let a malformed request silently wipe analyst-set
+// labels and still answer 204 (#710). Only an explicit [] clears.
+func TestAPISetFindingLabels_rejectsMissingLabelsArray(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+	f, tok, _ := seedFindingForAPI(t, s)
+	path := fmt.Sprintf("/api/findings/%d/labels", f.ID)
+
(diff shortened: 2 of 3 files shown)


charlie-morrison:
Copilot is right on all three, and it's the same point in three places — thanks, that was a real inaccuracy in the PR rather than a style nit.

I checked it rather than taking it on faith: against this branch, `{"labels":[""]}`, `{"labels":[" "]}` and `{"labels":["  ","\t"]}` each return **204 and leave the finding with 0 of its 2 labels**. `db.SetFindingLabels` trims every name and skips the blanks, so an all-blank array reaches `Association("Labels").Replace` as the empty set. Saying "an explicit `[]` stays the way to clear" claimed an exclusivity the code doesn't have — in a PR whose entire point is that the docs disagreed with what the endpoint did to your data, that's the wrong sentence to leave in.

Pushed `80847ae`, which corrects the handler comment, the test doc comment and the OpenAPI description to describe trim-and-skip, and adds `TestAPISetFindingLabels_blankNamesAreIgnored` so the prose is now pinned by a test instead of a claim. It also asserts that a blank sent *alongside* a real name drops only the blank (`[" ","triage"]` → just `triage`), since "any blank clears" would be the wrong lesson to take from the first half.

No behaviour change: rejecting blank names is still out of scope per your comment on #710, and this only documents what already happens. `go build ./...`, `go vet`, and the full `internal/web` + `internal/db` suites pass locally.

andrew (approved):
Scoped exactly to #710: `body.Labels == nil` → 400 with the existing decoder message, `{"labels":[]}` still clears, blank-name handling left alone. The test asserts both the status and that the seeded labels survive, verified against the unfixed handler. Nice self-catch on the clear-all wording in the second commit — pinning the trim-and-skip behaviour keeps the openapi text honest.

PR #789, opened 2026-07-31
Buffer template rendering before writing HTTP responses

### LINKED ISSUE #573: Buffer template rendering before writing HTTP responses
### Summary

`internal/web/server.go` renders HTML templates directly into the `http.ResponseWriter` in `(*Server).render`.

### Problem

`ExecuteTemplate` writes output incrementally. If a template execution error happens after partial HTML has already been written, Go has already committed the response headers as `200 OK`. The later `http.Error(w, ..., http.StatusInternalServerError)` cannot reliably change the status to `500` and can append a plain-text error onto a partially-rendered HTML page.

This creates broken responses and can also trigger `http: superfluous response.WriteHeader` logs.

### Suggested Fix

Render into a `bytes.Buffer` first. Only set the final response headers and copy the buffer to `w` after template execution succeeds. If execution fails, log the error and send a clean `500 Internal Server Error` before any template bytes reach the client.

### Impact

Clients receive correct HTTP status codes and avoid corrupted mixed HTML/error responses when template rendering fails.

charlie-morrison:
Closes #573.

`render()` executed the template straight into the `ResponseWriter`. Because `ExecuteTemplate` writes incrementally, a mid-template error arrives after the 200 and part of the page are already committed — the `http.Error` that follows can't change the status and only appends plain text to half-rendered HTML.

Built to the approach you laid out in the issue: a `sync.Pool` of `*bytes.Buffer`, reset on Get, execute into it, and only on success set the headers and `WriteTo(w)`; on error, log and `http.Error` exactly as before, with nothing written yet.

**Verified against the unfixed code first**, so this states a fact rather than a hunch — on `main` the new test reports:

```
render_test.go:35: status = 200, want 500
render_test.go:38: response carries partially rendered HTML; body starts:
    "<!doctype html><title>t</title><div>PARTIAL-OUTPUT-MARKER</div>template: boom_test.html:1:66: ..."
```

Three tests: the error-after-partial-output case, the success path, and one that drives a **failed** render and then a successful one, to prove the failed page can't reach the next request out of the pool.

_Updated after review (c793677, tests only)._ The pooled-buffer test originally went success -> success, which asserts nothing: `WriteTo` drains the buffer as it writes, so that pair passes with `buf.Reset()` deleted — as it did. The reset is load-bearing on the **error** path, where the failed render returns its buffer to the pool with the partial page still in it; without the reset the next request is served 4122 bytes of the previous page under a matching `Content-Length`. The test now drives that path and fails on exactly it. I also dropped the 8 KiB of padding in the error test: the comment credited `net/http`'s buffering, but `render()` writes to an `httptest.ResponseRecorder`, whose `Write` appends straight to `Body` — no `bufio` in the path. Unpadded, the test still fails on pre-fix `render()` with `status = 200` and the marker in the body.

Two things I added beyond the issue, both easy to drop if you'd rather not have them:

- **Oversized buffers are dropped instead of recycled** (`maxPooledRenderBuf`, 1 MiB). Without it one long findings list pins that capacity in the pool for the process lifetime.
- **`Content-Length` is now set**, since buffering means the whole body is known before the first write. This is the one externally visible change: responses that were previously chunked now carry a length.

`render()` is the only place that changes; as you noted, the SSE endpoints don't go through it, so there's no streaming to preserve. `go build ./...`, `go vet`, `gofmt` and the full `internal/web` suite pass locally.

### DIFF
--- a/internal/web/render_test.go
+++ b/internal/web/render_test.go
@@ -1,0 +1,98 @@
+package web
+
+import (
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+// A template error raised after some output has already been produced must
+// still yield a clean 500. Executing straight into the ResponseWriter commits
+// a 200 plus the partial HTML before the error is known, so the http.Error
+// that follows cannot change the status and only appends plain text to a
+// half-rendered page (#573).
+func TestRenderTemplateErrorAfterPartialOutput(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+
+	const marker = "PARTIAL-OUTPUT-MARKER"
+	_, err := s.tmpl.New("boom_test.html").
+		Funcs(map[string]any{
+			"boom": func() (string, error) { return "", errors.New("boom") },
+		}).
+		// The padding matters: bufio in net/http only flushes to the recorder
+		// once enough bytes accumulate, so a short prefix could pass by luck
+		// rather than because the fix works.
+		Parse(`<!doctype html><title>t</title><div>` + marker +
+			strings.Repeat("x", 8<<10) + `</div>{{ boom }}`)
+	if err != nil {
+		t.Fatalf("parse test template: %v", err)
+	}
+
+	w := httptest.NewRecorder()
+	s.render(w, localReq("GET", "/"), "boom_test.html", nil)
+
+	if w.Code != http.StatusInternalServerError {
+		t.Errorf("status = %d, want 500", w.Code)
+	}
+	if got := w.Body.String(); strings.Contains(got, marker) {
+		t.Errorf("response carries partially rendered HTML; body starts: %q", got[:min(len(got), 120)])
+	}
+	if ct := w.Header().Get("Content-Type"); strings.Contains(ct, "text/html") {
+		t.Errorf("Content-Type = %q, want the plain-text error type, not HTML", ct)
+	}
+}
+
+// The success path is unchanged apart from now being able to declare its
+// length, since the whole body is known before anything is written.
+func TestRenderSuccessSetsHTMLHeaders(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+
+	const body = "<!doctype html><title>ok</title><p>hello</p>"
+	if _, err := s.tmpl.New("ok_test.html").Parse(body); err != nil {
+		t.Fatalf("parse test template: %v", err)
+	}
+
+	w := httptest.NewRecorder()
+	s.render(w, localReq("GET", "/"), "ok_test.html", nil)
+
+	if w.Code != http.StatusOK {
+		t.Fatalf("status = %d, want 200", w.Code)
+	}
+	if got := w.Body.String(); got != body {
+		t.Errorf("body = %q, want %q", got, body)
+	}
+	if ct := w.Header().Get("Content-Type"); ct != "text/html; charset=utf-8" {
+		t.Errorf("Content-Type = %q", ct)
+	}
+	if cl := w.Header().Get("Content-Length"); cl != strconv.Itoa(len(body)) {
+		t.Errorf("Content-Length = %q, want %d", cl, len(body))
+	}
+}
+
+// Buffers are pooled and reused, so a large page must not leave its bytes
+// visible to the next render.
+func TestRenderPooledBufferDoesNotLeakBetweenPages(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+
+	const leak = "LEAK-FROM-FIRST-PAGE"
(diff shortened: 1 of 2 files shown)


internal/web/server.go
```diff
@@ -565,13 +566,45 @@ func (s *Server) render(w http.ResponseWriter, r *http.Request, name string, dat
...
 	data["Sorter"] = sortCtx{path: r.URL.Path, query: sorterQuery}
-	w.Header().Set("Content-Type", "text/html; charset=utf-8")
-	if err := s.tmpl.ExecuteTemplate(w, name, data); err != nil {
+	// Render into a buffer first. ExecuteTemplate writes incrementally, so
+	// executing straight into the ResponseWriter commits a 200 and part of the
+	// page before a mid-template error is known; the http.Error below then
+	// cannot change the status and only appends plain text to half-rendered
+	// HTML. Buffering keeps the failure path able to send a clean 500.
+	buf := renderBufPool.Get().(*bytes.Buffer)
+	buf.Reset()
```
alexandre-daubois, line 575: `TestRenderPooledBufferDoesNotLeakBetweenPages` goes success -> success, and `WriteTo` already resets the buffer on its way out, so the pooled buffer is always empty by the time the second render Gets it: delete `buf.Reset()` from `server.go` and all three tests still pass. The path where Reset is load-bearing is error -> success, where the failed render's partial page stays in the buffer, and driving the erroring template first brings back 4 KB of the previous request's page with a matching Content-Length.

internal/web/render_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestRenderTemplateErrorAfterPartialOutput(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+
+	const marker = "PARTIAL-OUTPUT-MARKER"
+	_, err := s.tmpl.New("boom_test.html").
+		Funcs(map[string]any{
+			"boom": func() (string, error) { return "", errors.New("boom") },
+		}).
+		// The padding matters: bufio in net/http only flushes to the recorder
```
alexandre-daubois: The comment on the padding isn't right: the test calls `s.render` directly with an `httptest.ResponseRecorder`, whose Write appends straight into `rw.Body`, so there's no `net/http` `bufio.Writer` anywhere in the path. Dropping the 8 KiB of x still fails on main with the same status = 200, want 500 and the marker in the body, so either reword the reason or drop both.

internal/web/server.go
charlie-morrison: You're right, and it's worse than an uncovered line — I deleted `buf.Reset()` from `server.go` and all three tests stayed green, so that test was asserting nothing about the reset.

The error -> success path reproduces exactly as you described. Driving the failing template first, then a small one, against the code with `Reset` removed:

```
body = "LEAK-FROM-FAILED-PAGEyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", want "small"
Content-Length = "4122", want 5
```

4122 bytes of the previous request's page under a matching `Content-Length` — a well-formed response carrying someone else's page, which is the part that would never look broken downstream, so I asserted on the header too rather than just the body.

`TestRenderPooledBufferDoesNotLeakBetweenPages` is now `TestRenderPooledBufferDoesNotLeakFailedPageIntoNextRender` and drives error -> success; it fails without the `Reset` and passes with it. The comment says why success -> success can't show this, so it doesn't get "simplified" back later.

In c793677. No production code changed.

internal/web/render_test.go
charlie-morrison: Correct — the reason I gave was wrong. `render()` writes to the `httptest.ResponseRecorder` here, and its `Write` appends straight into `rw.Body`; there's no `net/http` `bufio.Writer` anywhere in that path, so nothing was being flushed past.

Checked it rather than just reasoning about it: with the 8 KiB dropped, the unpadded template still fails on pre-fix `render()` with both original symptoms —

```
status = 200, want 500
response carries partially rendered HTML; body starts: "<!doctype html><title>t</title><div>PARTIAL-OUTPUT-MARKER</div>template: boom_test.html:1:66: ..."
```

So the padding was buying nothing. Dropped both it and the comment — the template is now one readable line, which is what it should have been. The 4 KiB in the leak test is doing real work though (it's the payload that shows up in the next response), so that one stays.

Also in c793677, and I've fixed the same claim in the PR description, which repeated it.

andrew (approved):
Matches #573 exactly: pooled buffer, Reset on Get, execute into it, headers + `WriteTo` only on success, clean 500 on error. The oversized-drop and `Content-Length` extras are both sensible. Alexandre's review caught the success→success test not covering `Reset()`; the error→success rewrite in c793677 was verified against the mutated code, which is how it should be done.

PR #790, opened 2026-07-31
feat(ecosystems): allow disabling the enrichment and skip it when unreachable

### LINKED ISSUE #212: Allow disabling packages.ecosyste.ms enrichment and degrade gracefully when blocked
Scans currently call `packages.ecosyste.ms` to enrich SBOM packages with repository URLs, download counts, and dependent-repo data. This is opportunistic — the call is cheap and the data is package-names-only, so it's not a serious exfiltration concern — but operators running in restricted-egress environments need two things:

- A config option to disable the enrichment entirely (no outbound calls to `packages.ecosyste.ms`), so the network policy can deny the domain without producing noise or partial state.
- Graceful degradation when the domain is reachable in config but blocked at the network layer: timeouts and DNS failures should not break the scan or leave half-enriched rows, just skip the enrichment and log it.

Surfacing this in the config alongside the egress allowlist work would let an operator say "no third-party enrichment" in one place.

alexandre-daubois:
Fix #212


andrew (approved):
Delivers both #212 asks cleanly. `DisableEcosystems()` neuters both seams as well as setting the flag, the 503/422 split keeps the no-PURL reason honest, SBOM re-resolve preserves an earlier enabled run's more specific error, and `unreachable()` correctly excludes response-level timeouts so one slow host can't starve the sources after it. `depScan` going through `s.resolvePURL` fixes that handler bypassing the seam.

Two observations, neither blocking: `unreachable()` matches any `*net.OpError` (including `Op: "read"`, e.g. a mid-response RST) which is broader than the comment's "the dial itself" — arguably still correct for a firewall that RSTs post-connect, just undersold in the comment. And the stray-argument guard at `main.go:48` is a CLI behaviour change beyond #212, though well-motivated for the first default-true bool.

PR #794, opened 2026-08-01
Use magic classification in code browser

andrew:
The code browser now uses `clone.InspectBlob` and `magic` to classify content before converting it to a string. It renders UTF-8 text and leaves binary, unknown, and UTF-16 content behind a neutral notice, while preserving the 2 MiB limit and truncation banner. This catches signed formats without NUL bytes and malformed UTF-8 that the old `clone.Blob` check could display.


p-linnane (approved):
Nice!

PR #797, opened 2026-08-03
Align audit finding location schemas

abhinavgautam01:
Follow-up to https://github.com/alpha-omega-security/scrutineer/pull/784#pullrequestreview-4842148313
Part of #106

## Summary

Aligns the location validation contract across the bundled `audit-injection`, `audit-exfil` and `audit-authz` skills.

This resolves the pre-existing schema drift Andrew noted during review of #784 and prepares the three audit schemas for a future shared `$defs` extraction.

## Changes

- Update `audit-injection`’s location pattern from:
  - `^[^:]+:[1-9][0-9]*$`
  - to `^.+:[1-9][0-9]*$`
- Match the existing `audit-exfil` and `audit-authz` location contract.
- Add a structural parity test for the three audit schemas.
- Ignore only the skill-specific schema title during comparison so any future contract drift fails tests.

The shared `$defs` extraction remains intentionally out of scope for this follow-up.

## Tests

- `go test ./internal/worker -run 'Test(AuditFindingSchemasStayAligned|BundledSchemas)' -count=1`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `golangci-lint run ./cmd/... ./docker/... ./internal/... ./skills/...`
- `git diff --check`

### DIFF
--- a/internal/worker/schema_bundled_test.go
+++ b/internal/worker/schema_bundled_test.go
@@ -1,10 +1,40 @@
 package worker
 
 import (
+	"encoding/json"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 )
+
+func TestAuditFindingSchemasStayAligned(t *testing.T) {
+	paths := []string{
+		"../../skills/audit-injection/schema.json",
+		"../../skills/audit-exfil/schema.json",
+		"../../skills/audit-authz/schema.json",
+	}
+
+	var baseline map[string]any
+	for _, path := range paths {
+		raw, err := os.ReadFile(path)
+		if err != nil {
+			t.Fatalf("read %s: %v", path, err)
+		}
+		var schema map[string]any
+		if err := json.Unmarshal(raw, &schema); err != nil {
+			t.Fatalf("decode %s: %v", path, err)
+		}
+		delete(schema, "title")
+		if baseline == nil {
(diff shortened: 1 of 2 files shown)


andrew (changes requested):
The pattern change and parity test are both right — after this the three schemas are byte-identical modulo `title`, which is where #784 said they should be before the `$defs` collapse.

One thing before merge:

- `schema_bundled_test.go:33` — `t.Errorf("%s differs from %s beyond title", path, paths[0])` says drift happened but not where. The other tests in this file print the mismatch; this one should too, since whoever trips it during the `$defs` work will otherwise re-run `diff` by hand. `cmp.Diff(baseline, schema)` (go-cmp is already a dependency) or marshal both back to JSON and print the two strings.

abhinavgautam01:
Thanks, fixed

The parity test now canonicalizes each title-stripped schema using `json.MarshalIndent`. When schemas differ, the failure prints both the baseline and actual normalized schemas, making the exact drift visible in CI.

I used the JSON approach instead of `cmp.Diff` because `go-cmp` is not currently present in this branch's `go.mod` or `go.sum`, avoiding an additional test-only dependency.

andrew (approved):
The requested schema-diff diagnostic is addressed.

PR #799, opened 2026-08-03
worker: have stageContext write context.json to workRoot and skill dir

### LINKED ISSUE #499: Have stageContext write context.json to both workRoot and skill dir
#470 made `stageSkill` mirror `workRoot/context.json` into the skill directory so `./context.json` resolves from there. That works, but it introduced an implicit ordering dependency: `stageSkill` must run after `stageContext`, and the failure when they're swapped is silent (the read returns `ErrNotExist` and the mirror is skipped). #470 fixed the one violation in `internal/worker/exposure.go` and documented the constraint on `stageSkill`, but a future caller can still get it wrong with no signal.

The more durable shape is to have the producer write to both places, the same way `stageSkill` writes `schema.json` to both `dst` and `workRoot` because it owns that data. `stageContext` owns `context.json`, so it should write to both `workRoot` and the skill dir:

```go
func stageContext(workRoot, skillDir, apiBase, ... ) error {
    // ... build ctx ...
    for _, p := range []string{workRoot, skillDir} {
        if p == "" { continue }
        if err := os.WriteFile(filepath.Join(p, "context.json"), data, filePerm); err != nil {
            return err
        }
    }
    return nil
}
```

Then `stageSkill` drops its read-back of `workRoot/context.json` and the ordering constraint goes away. Two call sites to update (`internal/worker/skill.go`, `internal/worker/exposure.go`); the test from #470 stays.

charlie-morrison:
Closes #499.

`stageContext` now writes `context.json` to both `workRoot` and the skill directory, and `stageSkill` drops its read-back of `workRoot/context.json` — the producer owns the data and writes every copy, as the issue describes.

## The ordering constraint doesn't go away — it inverts, and I couldn't make it disappear without changing who owns `skillDir`

The issue expects `stageSkill`'s ordering dependency to disappear once `stageContext` writes both copies. It doesn't, because `stageSkill` opens with `os.RemoveAll(dst)`: it owns the skill directory's lifecycle so a restage can't leave stale files behind. So with the naive shape — `stageContext` first, as both call sites had it — the skill-dir copy is written and then **deleted milliseconds later**, and `./context.json` stops resolving from the skill dir with no error anywhere. That is the same silent failure the issue is trying to remove, just moved.

I verified it rather than assuming it. With the new `stageContext` in place and only the call order flipped back:

```
--- FAIL: TestStageWorkspace_skillStagingDoesNotEatContextJSON
    skill_test.go:743: ./context.json does not resolve from the skill dir:
    stat .../001/.claude/skills/s/context.json: no such file or directory
```

So both call sites now stage the skill **first**, and the doc comment on `stageSkill` says why. The constraint is inverted, not removed, and I'd rather say that plainly than claim the issue's stated benefit.

**The alternative, if you want the constraint genuinely gone:** hoist the `os.RemoveAll(dst)` out of `stageSkill` into an explicit reset at the top of `stageWorkspaceWithInputs` and `doExposure`. Then neither function owns `skillDir`, both just write into an already-cleared directory, and the order really is irrelevant. I didn't take it because it changes `stageSkill`'s contract for the six tests that call it directly and for any future standalone caller that relies on it clearing stale files (`TestStageSkill_remirrorClearsStaleScripts` is the one that would notice). Happy to switch if you'd prefer that shape — it's a small diff.

## Tests

- `TestStageWorkspace_skillStagingDoesNotEatContextJSON` — pins the hazard above from outside, through `StageWorkspace`, so a future reorder fails in CI instead of silently in production. This is the test that fails on the flipped order.
- `TestStageContext_writesToWorkRootAndSkillDir` — both copies exist and are **byte-identical**, so they can't drift apart.
- `TestStageContext_emptySkillDirWritesWorkRootOnly` — callers that stage no skill (diff rescan, evals, the `stageContext` unit tests) pass `skillDir == ""` and keep exactly the old single-file behaviour.
- `TestStageSkill_mirrorsContextJSONToSkillDir` is removed: it asserted the read-back that this PR deletes. Its intent — `./context.json` resolves from the skill dir — is what the first test above now covers end to end, which is the better place for it since that was always a property of the *staging sequence*, not of `stageSkill` alone.

The `#470` test stays as-is and passes.

## Notes

- `stageContext` `MkdirAll`s each destination, so it no longer depends on the skill dir already existing.
- Signature change is internal (`stageContext`, `stageContextWithInputs`); the exported `StageWorkspace` is untouched, so eval harnesses don't need updating.
- No behaviour change for `skillDir == ""` callers.

### DIFF
--- a/internal/worker/diff_rescan_test.go
+++ b/internal/worker/diff_rescan_test.go
@@ -76,7 +76,7 @@ func TestPrepareDiffRescanStagesDiffInputs(t *testing.T) {
 		t.Fatalf("threat model scan id = %v, want %d", stored.DiffThreatModelScanID, tm.ID)
 	}
 
-	if err := stageContext(workRoot, "http://api", "", DefaultMetadataDir, &stored, &repo); err != nil {
+	if err := stageContext(workRoot, "", "http://api", "", DefaultMetadataDir, &stored, &repo); err != nil {
 		t.Fatal(err)
 	}
 	var ctx skillContext
--- a/internal/worker/exposure.go
+++ b/internal/worker/exposure.go
@@ -167,11 +167,13 @@ func (w *Worker) doExposure(ctx context.Context, scan *db.Scan, emit func(Event)
 	}
 
 	skillDir := w.Runner.SkillDir(workRoot, skill.Name)
-	if err := stageContext(workRoot, w.APIBase, w.ForkOrg, w.metadataDir(), scan, &scan.Repository); err != nil {
-		return "", fmt.Errorf("stage context: %w", err)
-	}
+	// stageSkill clears skillDir, so it runs before stageContext, which
+	// writes context.json into that directory (#499).
 	if err := stageSkill(&skill, workRoot, skillDir); err != nil {
 		return "", fmt.Errorf("stage skill: %w", err)
+	}
+	if err := stageContext(workRoot, skillDir, w.APIBase, w.ForkOrg, w.metadataDir(), scan, &scan.Repository); err != nil {
+		return "", fmt.Errorf("stage context: %w", err)
 	}
 
 	depRepo := db.Repository{URL: dep.RepositoryURL, Name: dep.Name}
--- a/internal/worker/novelty_test.go
+++ b/internal/worker/novelty_test.go
@@ -49,7 +49,7 @@ func TestNoveltyContextStagesBoundedChangedFileEvidence(t *testing.T) {
 
 	scan := fixture.scan(head)
 	if err := stageContextWithInputs(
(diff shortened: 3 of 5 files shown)


andrew (approved):
The inverted-constraint shape is the one I want, for the reason you gave: `stageSkill` staying self-clearing means a future standalone caller doesn't inherit stale scripts. Hoisting the `RemoveAll` would move that trap to every call site, which is the same "silent, easy to get wrong" pattern #499 was filed against.

#499's real complaint was silent failure on reorder; `TestStageWorkspace_skillStagingDoesNotEatContextJSON` fixes that. Good catch on the `RemoveAll` — the issue was wrong about the constraint disappearing.

Thanks for this.

PR #803, opened 2026-08-04
Share audit finding schema definitions

abhinavgautam01:
Part of #106

## Summary

Consolidates the identical finding contracts used by `audit-injection`, `audit-exfil` and `audit-authz` into one shared JSON Schema.

This follows the schema-alignment work in #784 and #797 while preserving the self-contained `schema.json` that workers store and stage for each scan.

## Changes

- Add `skills/_shared/audit-findings.schema.json` as the canonical audit finding contract.
- Replace the three duplicated audit schemas with small skill-specific wrappers referencing the shared contract.
- Resolve local schema references while loading skills and bundle them under `$defs`.
- Store and stage the resulting schema as a self-contained document.
- Restrict local references to the skill collection boundary.
- Resolve symlinks before containment checks to prevent reference-path escapes.
- Include bundled schema contents in `SourceHash`, ensuring shared-schema changes bump affected skill versions.
- Update bundled-schema and ingestion tests to use the production schema-loading path.
- Add tests covering:
  - Reference bundling and internal `$ref` rewriting.
  - Shared-schema source-hash changes.
  - Path traversal and symlink containment.
  - Audit wrapper consistency.
  - Good and invalid audit reports.
  - Finding-reference ingestion.

## Tests

- `go test -race ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `golangci-lint run`
- `git diff --check`


andrew (changes requested):
Direction's right and the bundler is careful (symlink-resolved containment, cycle-safe, `SourceHash` folded in). A few things before merge:

- `shouldSkipDir` in `internal/skills/load.go` doesn't skip `_shared`, so `LoadDirectoryExcept` walks into it. Harmless while it only holds `.schema.json`, but a stray `SKILL.md` there would be upserted as a real skill. Add underscore-prefixed dirs to the skip set.
- `TestAuditFindingSchemasStayAligned` is now dead weight. #797 added it to catch drift between three copy-pasted files; this PR removes the copy-paste. `TestAuditFindingSchemasReferenceSharedContract` already pins each wrapper to `{\$schema, title, \$ref: <shared>}`, so bundling and comparing modulo `title` can only fail if the bundler is non-deterministic, and `TestParseFile_bundlesLocalSchemaReferences` already covers that. Delete it.
- The `len(wrapper) != 3` check in `TestAuditFindingSchemasReferenceSharedContract` will trip on any future `\$id`, `\$comment`, or `description` in a wrapper. If the intent is "validation keywords belong in the shared file", assert absence of `type`/`properties`/`\$defs` rather than an exact key count.
- `decodeSchemaObject`: `strings.NewReader(string(raw))` copies the `[]byte`; use `bytes.NewReader(raw)`.

abhinavgautam01:
Thanks, fixed all four points.

- Skill discovery now skips underscore-prefixed directories, with a regression test proving a `_shared/SKILL.md` is not loaded.
- Removed `TestAuditFindingSchemasStayAligned`, since the shared contract makes the old copy-drift check redundant.
- Relaxed the wrapper test so it rejects validation keywords (`type`, `properties` and `$defs`) instead of requiring an exact top-level key count.
- Changed `decodeSchemaObject` to use `bytes.NewReader(raw)` and avoid the byte-to-string copy.

andrew (approved):

PR #806, opened 2026-08-04
feat: Monorepo sub-package support: scoped scans, attribution, and per-package identity

dkw-oss:
# Monorepo sub-package support: scoped scans, attribution, and per-package identity

## Summary

Scanning a monorepo (e.g. `rails/rails`, which ships activesupport, actionpack,
… as separate gems) previously rolled everything up under one repository entry:
findings, packages, advisories, maintainers, and disclosure all keyed to the
repo URL. This PR makes a monorepo sub-package a first-class thing you can submit,
scan, attribute, disclose, and export on its own.

Submit `owner/repo#sub/dir` (or paste a `.../tree/<branch>/<sub/dir>` URL) and the
whole scan pipeline scopes to that sub-folder, which gets its own page with its
own findings, published packages, advisories, and disclosure channel.

## Background — what already existed

A subprojects scaffold was already present (subprojects skill, `Subproject`
table, `Scan.SubPath`, "scan all subprojects"), but it was half-built: `#subdir`
only tagged the triage scan, the pipeline still ran repo-wide, sub-path scoping
was a soft prompt hint, and there was no link between a published package and the
sub-folder it came from. This PR completes that scaffold rather than reinventing
it.

## What's new

- **Submission + pipeline scoping.** `#subdir` eagerly creates a first-class
  `Subproject`; `triage` forwards `scan_subpath` to every pipeline child; the run
  API accepts an optional validated `sub_path`; the scan dedup/409 key is
  `sub_path`-aware so two sub-packages can run the same skill concurrently.
- **Hard scoping + fallback.** A hard-scoped scan prunes sibling packages from
  the workspace but keeps the sub-path at `./src/<subPath>` and preserves `.git`,
  so profile detection, `scan_subpath`, and finding attribution are identical to
  a soft scan — only the physical tree differs. If a sub-package can't resolve its
  dependencies in isolation, the scan widens to the whole tree and re-runs once,
  recording the mode on `Scan.ScopeMode`. The trigger watches the run's streamed
  output (where a failed `bundle install` actually surfaces), not just the final
  report. Finding-scoped runs (verify/patch/disclose) and diff rescans stay soft
  — the latter because a pruned worktree would diff as all-siblings-deleted.
- **Registry attribution.** A Go-side reconcile links each `Package`/`Advisory`
  to the `Subproject` it belongs to, matched by manifest name (directory-basename
  fallback), self-healing to repo-level when unmatched. The maintainers skill may
  emit an optional per-sub-package disclosure channel.
- **Per-sub-package identity.** A subproject detail page (findings, attributed
  packages/advisories, disclosure channel), a per-sub-package disclosure-channel
  setter, and a `sub_path` export filter so a monorepo can be exported/encrypted
  one package at a time.
- **Stable Subproject identity.** The subprojects parser now upserts on
  `(repository_id, path)` instead of delete-and-recreate, so attribution
  references survive skill re-runs.

## Toggles

Both default to the pre-existing behaviour for single-package repos and revert it
fully when flipped:

- `subproject_scope: hard|soft` (config default + per-scan `Scan.ScopeMode`
  override) — the filesystem scoping above.
- `monorepo_attribution: on|off` — the registry/disclosure split. Independent of
  the scope toggle, so you can attribute per-package but still audit the whole
  tree, or vice versa.

## Interface impact

Additive only:

- New **nullable** columns: `Package.SubprojectID`, `Advisory.SubprojectID`,
  `Subproject.DisclosureChannel`, `Scan.ScopeMode`.
- New **optional** fields: `sub_path` on the skill-run request body and the
  findings-export query; an optional `subprojects[]` disclosure block in the
  maintainers skill schema.
- New routes: the subproject detail page + its disclosure-channel setter.

Not changed: `Repository.url` stays the unique key (no second repo row per
subdir); the clone cache is still keyed on the bare URL (one shared clone);
existing API shapes gain fields but drop none. A single-package repository
behaves identically end-to-end.

The one place an *existing* path changes: subproject scans now hard-scope by
default (they were whole-tree/soft before). This is gated — `subproject_scope:
soft` restores the prior behaviour exactly — and root/single-repo scans are
untouched.

Maintainer *rows* deliberately stay repo-wide (a maintainer is a person, keyed by
login across repos, with no clean per-package key); per-sub-package disclosure
routing is done through the sub-package's disclosure channel instead.

## Testing

`go build ./...`, `go vet ./...`, `gofmt`, and `go test ./...` all pass. New tests
cover the sub-path sanitiser, Subproject stable-id upsert, submission +
traversal rejection, sub_path-aware dedup, the attribution reconcile (incl.
self-heal), per-sub-package disclosure channel, hard-scope prune (top-level +
nested), scope-mode resolution, the dependency-resolution classifier, the
hard→soft fallback — triggered both from the final report and from the streamed
output — retry reproducibility of `ScopeMode`, the diff-rescan-stays-soft guard,
and the subproject page/handlers.

## Backward compatibility

Additive, nullable schema; unique key unchanged; both toggles default to prior
behaviour; single-package repositories are unaffected.


andrew (changes requested):
Nice work, this closes off the monorepo scaffold properly and the test coverage is thorough. One bug that needs fixing before merge, plus a couple of small things.

### `subprojects` skill run under hard scope prunes the repo's Subproject table

Triage now forwards `sub_path` to every child, including `subprojects` (line 71 of its enqueue list). Under the new `subproject_scope: hard` default that child runs against a workspace pruned to just the one sub-folder. `subprojects` enumerates whatever is under `./src` (its own SKILL.md says it is "not sub-path-scoped... this skill *produces* sub-paths; it does not consume one"), so it now sees one directory. `parseSubprojectsOutput` then upserts what it saw and deletes everything else at repo level (`skill_parsers.go`, the `path NOT IN ?` prune).

Concrete case: submit `rails/rails#activesupport` on a fresh repo. `EnsureSubproject` creates the `activesupport` row, triage enqueues `subprojects` with `sub_path=activesupport`, hard-scope prunes `./src` to `activesupport/` + `.git`, the skill reports at most `[{path:"activesupport"}]`, and the parser deletes every other Subproject row for the repo. If the repo had already been fully scanned, submitting one sub-package wipes the rest and the reconcile that follows nulls every `Package.SubprojectID` / `Advisory.SubprojectID` that pointed at them. And because `hasOpenRepoScopedScan` now keys on `sub_path`, a repo-root `subprojects` run and a scoped one no longer dedupe against each other, so they can race last-writer-wins on the same table.

I think the cleanest fix is to add the repo-wide-projection output kinds (`subprojects` at minimum) to the `scanScopeHard` opt-out alongside finding-scoped and diff-rescan, since those skills describe the whole repository regardless of which sub-folder was requested. A guard in `parseSubprojectsOutput` that skips the prune when `scan.SubPath != ""` would also stop the data loss if a scoped run slips through some other way.

### Minor

- `internal/worker/worker.go` `SubprojectScope` field comment says `"hard"/"" stages only the sub-folder`, but `scanScopeHard` returns `mode == "hard"` and its own doc (and `TestScanScopeHard` "subpath + unset default is soft") say an unset default is soft. Should be `"soft"/""` on the whole-clone side.
- `internal/db/db.go` runs the `DELETE FROM subprojects WHERE id NOT IN (SELECT MIN(id) ...)` on every `Open()`. It's a no-op once the unique index exists and the table is tiny so probably fine, just flagging that it's a full-table scan per boot rather than a one-shot.
- The streamed-output fallback trigger is a lowercase substring match over the whole agent narration. An audit agent quoting one of the marker strings from the project's own error-handling code (a project that itself emits `"could not find gem"`, say) would trip a spurious whole-tree re-run. Cost-only and confined to hard-scoped sub-package scans, so I'm not too worried, but worth keeping in mind if we see doubled runs.

Happy to help with any of this if useful, just let me know.

dkw-oss:
Thanks @andrew - addressed:

1. **Repo-wide skills opt out of hard scope** — `scanScopeHard(scan, outputKind)` returns soft for the nine repo-wide projection kinds (subprojects, dependencies, packages, advisories, maintainers, repo_metadata, repo_overview, posture, finding_dedup), so a sub-path scan never prunes them to a fragment; fixes the reported `subprojects` Subproject-table wipe *and* the identical latent data loss in `dependencies`/`maintainers`, on every enqueue path. 

2. **subprojects prune backstop** — `parseSubprojectsOutput` skips its wholesale prune when `scan.SubPath != ""`, turning any run that reaches the parser mis-scoped into an upsert-only no-op instead of a table wipe 

3. **Scope-mode comment fixed** — `Worker.SubprojectScope` now documents an empty field as `"soft"`, not `"hard"` (the CLI `-subproject-scope` flag supplies the hard default). **[minor 1]**

4. **Dedup migration is one-shot** — the `DELETE FROM subprojects` + unique-index build in `db.Open()` is gated on `Migrator().HasIndex`, so it runs once instead of a full-table scan on every boot. **[minor 2]**

5. **Fallback false-positive documented** — recorded that the dependency-resolution soft-fallback matches the agent's streamed *narration* by necessity (the harness discards raw tool output), so a spurious re-run from an audit quoting a marker string is cost-only and accepted rather than masked by a heuristic that could miss real resolver failures. **[minor 3]**

Also:

1. `parseMaintainersOutput` skips repo-wide writes (`Association.Replace` + repo disclosure channel) on a sub-path-scoped run while keeping the per-subproject channel, and `parseDependenciesOutput` gained a comment recording its whole-repo-view assumption; both defend against a future edit that makes the skill honour `scan_subpath`. 

2. **repo-overview live fix + contract** — `repo-overview` no longer honours `scan_subpath` (always `brief ./src`): it was the one repo-wide skill self-narrowing, silently overwriting the repository's languages/license with a sub-package's on every scoped submission; the `scan_subpath` contract was sharpened in all three canonical spots (`worker/skill.go`, `db.go`, `docs/skills.md`) so no future repo-wide skill repeats it.

andrew (approved):
Thanks — the repo-wide-projection opt-out plus the write-side backstops cover the table-wipe cleanly, and the classification test means a new output kind can't quietly regress it. The repo-overview de-subpath was a good catch.

PR #807, opened 2026-08-04
Have metadata, packages and advisories skills use cached ecosyste.ms data

### LINKED ISSUE #520: Have metadata, packages and advisories skills use cached ecosyste.ms data
The repository model now stores cached ecosyste.ms payloads, and the internal API exposes them through:

`GET /api/repositories/{id}/ecosystems/{source}/raw`

However, the bundled `metadata`, `packages` and `advisories` skills still instruct the model to fetch ecosyste.ms directly.

### Problem

These skills still spend turns and context on external fetches and raw JSON handling even when Scrutineer already has the upstream payload cached on the repository row.

### Suggested change

Update the bundled skills to try the internal cached endpoint first:

- `metadata` -> `repo`
- `packages` -> `packages`
- `advisories` -> `advisories`

If the endpoint returns `404`, the skills can fall back to the current direct ecosyste.ms request so behavior remains robust when the cache is absent.

### Benefits

- Reduces model context bloat from large upstream JSON payloads.
- Avoids unnecessary external fetches during skill execution.
- Reuses the cached ecosyste.ms data introduced by the server-side prefetch work.
- Keeps current fallback behavior for uncached repositories.

charlie-morrison:
Closes #520.

The three bundled ecosyste.ms skills now try `GET {api_base}/repositories/{repository_id}/ecosystems/{source}/raw` with the scan's bearer token before falling back to the direct upstream URL on 404, using the mapping from the issue (`metadata` → `repo`, `packages` → `packages`, `advisories` → `advisories`).

I took the correction in your comment as the framing: the PR does not claim a context-size win, since `/raw` returns the verbatim upstream payload either way. The wins as written are `--hardened`, offline, and letting `advisories` skip re-paginating what the prefetcher already collated.

The `--hardened` claim is pinned by tests already in the repo, so it is stated as a fact rather than a hunch — `cmd/scrutineer/main_test.go:474` asserts default mode contains `*.ecosyste.ms` and `:489` asserts hardened mode does not. Under `--hardened` the direct fetch is refused by the egress proxy, so the cached endpoint is not an optimisation for these three skills, it is the only path that works.

### One thing the issue did not mention, and the reason this is not purely docs

Adding the fallback creates a state that could not be reached before: **both sources unavailable at once.** Under `--hardened` on a repository with nothing cached, the endpoint 404s *and* the direct fetch is blocked.

That matters because of what the parsers do. `parsePackagesOutput` and `parseAdvisoriesOutput` replace the whole row set — `DELETE WHERE repository_id = ?` and then insert whatever the skill produced. The old instructions told the skill to write `{"packages": []}` when it had nothing, and called it "a valid result, not an error". In the new unreachable state that empty array is reported as success and **deletes packages and advisories a previous scan recorded**, which is the same hazard your comment above `buildEgressAllow` describes for `ecosystems_enrichment`.

So each skill now separates the two cases explicitly:

- a source **answered** and upstream genuinely has nothing → write the empty array, unchanged behaviour
- **no source answered** → do not write `./report.json`; exit non-zero

I checked that the harness honours that before writing it in: in `internal/worker/skill.go`, a `RunSkill` error returns before the parse (only a `MaxTurnsReachedError` gets the partial path), and the parse is additionally guarded by `if report != ""`. So both refusing to write and exiting non-zero leave the existing rows untouched.

`metadata` gets the same rule for consistency, though it is the mildest of the three: `parseRepoMetadataOutput` guards each field with `if m.X != ""`, so `{}` mostly leaves the row alone. It does still blank the stored `metadata` blob and set `stars`/`forks` to zero unconditionally, which is why an unreachable run should not report one.

**If you would rather keep the skills always-writes and treat this as out of scope, say so and I will cut it back to the cached-first step alone** — it is a behaviour change to the failure path, and #520 as filed only asked for the endpoint swap.

### Notes

- The fallback is documented as a first-class path, not an error path, since `ecosystems_enrichment: false` makes 404 the steady state — matching the wording already in `apiGetEcosystemsRaw`'s doc comment.
- `compatibility:` in the frontmatter was updated on all three; it previously promised network access to ecosyste.ms as a hard requirement. No other copies of those strings exist in the tree, so nothing drifted.
- No test added. The change is prose inside `SKILL.md`, and a test asserting on instruction wording would break on the next reword without protecting anything. I did verify all three still parse with `skills.ParseFile` — correct names, `requires_remote` preserved, **zero warnings** — and `go build ./...`, `go vet`, `go test ./internal/skills/... ./internal/bundledassets/...` are green.

### DIFF
--- a/skills/advisories/SKILL.md
+++ b/skills/advisories/SKILL.md
@@ -2,7 +2,7 @@
 name: advisories
 description: Fetch published GHSA and CVE advisories affecting any package this repository produces, via advisories.ecosyste.ms.
 license: MIT
-compatibility: Needs network access to advisories.ecosyste.ms.
+compatibility: Prefers the scrutineer API's cached ecosyste.ms payload; needs network access to advisories.ecosyste.ms only when nothing is cached.
 metadata:
   scrutineer.version: 1
   scrutineer.output_file: report.json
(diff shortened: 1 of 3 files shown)


andrew (approved):
Keeping the exit-non-zero-when-neither-source-answered rule — the cached-first step is what creates the both-unreachable state, and without it hardened+uncached would wipe rows via the delete-then-insert parsers. Thanks for flagging it rather than silently including it.

PR #808, opened 2026-08-05
Add per-dependent migration campaign tracking

abhinavgautam01:
Part of #12

## Summary

Adds operator-managed migration campaign tracking for downstream consumers affected by findings in abandoned or zombie package repositories.

Each finding/dependent pair can now record its current outreach state, an operator note and when the campaign information last changed.

## Changes

- Add migration campaign states:
  - `notified`
  - `acked`
  - `migrated`
  - `declined`
  - `silent`
- Store campaign status, note and update timestamp on `FindingDependent`.
- Add campaign controls to each actionable consumer in the finding migration guide.
- Validate campaign states and reject notes without a selected status.
- Scope updates to the requested finding/dependent pair.
- Keep campaign timestamps separate from exposure-verdict timestamps.
- Preserve campaign data when exposure scans upsert new verdicts.
- Keep completed tracked campaigns visible after a consumer becomes fixed or known not affected.
- Continue prioritizing actionable consumers by downstream reach.
- Document the new database fields.
- Add persistence, validation, scoping, rendering and regression tests.

This leaves dependent-count history and plotting as the remaining follow-up in #12.

## Tests

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `golangci-lint run`
- `git diff --check`

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -77,8 +77,8 @@ change history for findings. `payload` is JSON stored portably as text.
 ## package_alternatives
 
 Operator-curated migration targets for repositories classified as abandoned or
-zombie. Later #12 migration-guide and campaign tracking work can join on this
-table instead of reparsing notes or reports.
+zombie. The finding migration guide and dependent campaign tracking join on
+this table instead of reparsing notes or reports.
 
 | Column | Type | Notes |
 |--------|------|-------|
@@ -419,6 +419,9 @@ One row per (finding, dependent) pair the `exposure` skill has audited. Status m
 | rationale | text | One-paragraph explanation written by the skill, rendered in the finding page's per-dependent table. |
 | scan_id | integer FK | Exposure scan that wrote this row. |
 | scan_commit | text | HEAD of the dependent's clone when the verdict was made; lets the operator tell whether a later rescan would still apply. |
+| campaign_status | text | Operator-managed migration outreach state: `notified`, `acked`, `migrated`, `declined`, or `silent`. Empty means outreach has not started. |
+| campaign_note | text | Operator note about the downstream migration conversation or outcome. |
+| campaign_updated_at | datetime | When campaign status or note last changed. Null until outreach is recorded. |
 | created_at | datetime | |
 | updated_at | datetime | |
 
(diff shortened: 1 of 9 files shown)


andrew (changes requested):
Most of this is well done: `UpsertFindingDependent` already uses an explicit `AssignmentColumns` list so exposure re-scans don't clobber the new campaign columns (tested); `UpdateColumns` keeps `UpdatedAt` tied to the exposure verdict with `CampaignUpdatedAt` tracked separately (tested); the handler 404s on a dependent that belongs to a different finding; note-without-status and unknown-status are both rejected; the no-op path leaves `CampaignUpdatedAt` alone.

One behaviour bug:

**`internal/web/finding_migration.go` — tracked-but-resolved rows get truncated away.** `loadMigrationGuideDependents` now keeps a `Fixed`/`KnownNotAffected` row when it has a `CampaignStatus`, sorts it after all actionable rows, then still applies `rows = rows[:migrationGuideRowLimit]` (10). On any finding with 10+ actionable dependents, every tracked-and-completed campaign row is dropped and the operator's notes disappear from the only place they render. That contradicts the "keep completed tracked campaigns visible" bullet in the description. Apply the cap to actionable rows only and append tracked non-actionable rows after the truncation (or count them outside the cap), and add a test with >10 actionable rows plus a tracked `Fixed` row.

Nit: `internal/db/finding_helpers.go` — `fmt.Errorf("dependent campaign status is required when a note is set")` has no formatting verbs; `errors.New` would do.

abhinavgautam01:
Thanks, fixed.

The migration-guide limit now applies only to actionable dependents. Campaign-tracked resolved rows (`fixed` or `known_not_affected`) are sorted separately and appended after the capped actionable list, so completed campaign notes remain visible regardless of the number of active consumers.

I added a regression test with 11 actionable dependents plus a tracked fixed dependent. It verifies that only the top 10 actionable rows remain while the tracked resolved campaign is still included.

I also replaced the non-formatting `fmt.Errorf` with `errors.New` and rebased the branch onto current `main`.

andrew (approved):
Truncation fix and `errors.New` both addressed; regression test covers the >10-actionable case. Merging.

PR #809, opened 2026-08-05
Add opt-in audit-pii skill

abhinavgautam01:
Part of #106

## Summary

Adds an opt-in `audit-pii` skill for focused static analysis of personal and customer-data exposure.

The skill identifies concrete personal identifiers or customer-confidential data committed to source or exposed through durable and lower-trust sinks, while excluding synthetic examples, reserved addresses, public author metadata and legitimate feature-local processing.

## Changes

- Add the bundled, read-only `audit-pii` skill covering:
  - Real personal or customer data committed to code, documentation, tests, fixtures, snapshots, cassettes or configuration
  - PII in logs, exceptions, telemetry, metrics, monitoring context, cache keys and URLs
  - Over-broad API, GraphQL, serializer, export and response fields
  - Account and email enumeration
  - Customer-specific billing, usage, support, and account data exposure
- Require concrete identifier and exposure evidence before reporting.
- Add explicit false-positive controls for:
  - RFC example domains and reserved IP ranges
  - Synthetic fixtures and placeholder identities
  - Git authors, translators, maintainers, and public role accounts
  - Legitimate storage and processing inside the feature trust boundary
  - Properly anonymized or HMAC-derived identifiers
- Reuse the shared audit findings schema.
- Keep the skill opt-in and out of the default triage pipeline.
- Document the skill in `README.md` and `docs/skills.md`.
- Add metadata, path-filter, schema-validation and findings-ingestion coverage.

## Tests

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `golangci-lint v2.12.2`
- `golangci-lint v2.12.2 --build-tags evals`
- `go mod tidy -diff`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ Adding a repo enqueues the `triage` skill, whose SKILL.md lists the further skil
 | `audit-injection` | Opt-in static audit for command/code execution, unsafe deserialization, and server-side template injection with ecosystem-specific references |
 | `audit-exfil` | Opt-in static audit for SSRF, path traversal, XXE, and response or diagnostic leakage with ecosystem-specific references |
 | `audit-authz` | Opt-in static audit for IDOR, tenant isolation, fail-open guards, privilege escalation, and authorization decisions based on unverified claims |
+| `audit-pii` | Opt-in static audit for real personal or customer-identifying data committed to source or exposed through logs, URLs, telemetry, exports, and responses |
 
 Edit `skills/triage/SKILL.md` to change what gets run by default. Drop new skill directories in `skills/` to add scan types; no code changes needed. See [docs/skills.md](docs/skills.md) for the frontmatter reference, the `scrutineer.*` metadata keys, the `context.json` shape, output kinds, schema validation, and the skill-facing HTTP API.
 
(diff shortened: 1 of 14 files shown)


andrew (changes requested):
The prompt body is solid — the identifier+exposure two-part test, the "don't reproduce the full value in the report" rule, and the authorship-metadata / role-mailbox exclusions are all good PII-specific additions. But it's missing the piece that #106 says is the reason to split these out at all.

**No `references/` directory.** `audit-injection`, `audit-exfil`, and `audit-authz` each ship a per-ecosystem reference pack (authz has eight: python, node, ruby, java-jvm, go, php, graphql, jwt). #106 is explicit that the reference pack is the point: "narrow scope plus deep references beats one prompt that tries to hold every framework's quirks in its head at once." Without one, `audit-pii` is a well-written prompt with no more framework knowledge than `security-deep-dive` already carries. There's plenty of ecosystem-specific material: Rails parameter logging and `filter_parameters`, Django `DEFAULT_EXCEPTION_REPORTER_FILTER` / `DEBUG` behaviour, Sentry `send_default_pii` and `before_send`, structured-logger context binding, DRF/Marshmallow/ActiveModel serializer field defaults, OpenTelemetry span attribute conventions. The `compatibility` line and Workspace section should also pick up the "Reads bundled reference notes in ./references" wording the siblings use, and `TestBundledAuditPIIMetadata` should walk the reference files the way `TestBundledAuditAuthzMetadata` does (the `dir` variable is currently only used once because that block is missing).

**Reserved-name allowlist.** In the false-positive controls: RFC 2606 also reserves `.test`, `.example`, and `.localhost` — `.test` in particular is common in fixtures and should be listed. `example.edu` isn't RFC-reserved (only .com/.net/.org are); harmless to keep but inaccurate to group with the others.

abhinavgautam01:
Thanks, addressed both points and rebased onto the latest `main`.

- Added a seven-file `references/` pack covering Python, Node.js, Ruby, JVM, Go, PHP, and cross-cutting observability behavior.
- The references cover Django exception filtering and serializer behavior, Rails `filter_parameters` and ActiveModel serialization, structured logger context, Sentry PII/event hooks and OpenTelemetry attributes and exporters.
- Updated the compatibility and Workspace sections to tell the skill to read `./references`.
- Extended `TestBundledAuditPIIMetadata` to walk every reference file, verify headings and pin the key ecosystem guidance.
- Corrected the allowlist to include the RFC-reserved `.test`, `.example`, `.invalid`, and `.localhost` namespaces and removed `example.edu` from the reserved-name grouping.

andrew (approved):
Both review points addressed; references pack is substantive and accurate. Merging.

Issue #812, opened 2026-08-05
Support Filtering the Scans Dashboard by Multiple Skills


abhinavgautam01:
### Summary

Allow operators to filter `/scans` by several skill names in one view, making it possible to inspect related model-backed audit jobs without including every scanner and workflow job.

### Affected Code

- `internal/web/scans.go`
  - `jobs`
  - `scansRetryFailed`
- `internal/web/templates/jobs.html`

### Current Behavior

The scans dashboard accepts one exact `skill` query parameter:

```go
skillName := r.URL.Query().Get("skill")
if skillName != "" {
    q = q.Where("skill_name = ?", skillName)
}
```

For example, this works:

```text
/scans?skill=security-deep-dive
```

There is no way to show several related skills together. A value such as:

```text
/scans?skill=security-deep-dive,vuln-scan,advisory-deep-dive
```

is treated as one literal skill name and returns no matching rows.

### Motivation

Operators commonly need a combined view of primary model-backed security audits while excluding auxiliary or deterministic scanner jobs such as `semgrep`, `zizmor`, metadata collection and workflow orchestration. Opening separate dashboard views makes comparison and bulk failure management harder.

### Proposed Behavior

Support a comma-separated set of skill names:

```text
/scans?skill=security-deep-dive,vuln-scan,advisory-deep-dive
```

The handler should:

1. Split the value on commas.
2. Trim whitespace.
3. Remove empty values and duplicates.
4. Use exact matching through `WHERE skill_name IN ?`.
5. Preserve the normalized filter through status, sort and pagination links.

The same parsing and query behavior must be shared with `scansRetryFailed`. Otherwise a combined failed-scans view would display the
correct rows while the existing "Retry all failed" action sends the combined string to an exact-match query and silently retries nothing.

The skill dropdown should also represent a combined selection clearly. The first iteration may keep individual dropdown choices while allowing combined filters through a URL, but it should not mark one individual skill as the current selection when multiple values are active.

### Acceptance Criteria

- A single skill filter preserves current behavior.
- Multiple comma-separated skills return the union of exact skill matches.
- Empty and repeated entries are normalized safely.
- Status, sorting, and pagination preserve the complete skill filter.
- "Retry all failed" applies to the same skill set shown by the dashboard.
- Tests cover single, multiple, malformed/empty and combined retry filters.

PR #813, opened 2026-08-05
feat(web): keep the scan lists up to date without a reload

alexandre-daubois:
Adds live capabilities to the scans page:

https://github.com/user-attachments/assets/e6492aba-160b-46e6-bda2-99f9aab5e109


andrew (changes requested):
Well-executed. The broker gains an event-name filter so list subscribers aren't sent every log line; `publishScanRow` / `publishScanList` split cleanly between "swap this row" and "re-fetch your table"; the repo Scans tab gets its own fragment route so a refresh doesn't re-run `loadRepoShowView`'s findings/deps/inventory loads; `scanRowColumns` narrows the projection off `log`/`report` with a parity test (`TestRepoScansFragment_rowMatchesTheFullPage`) guarding it; the flash cookie is left alone on htmx renders so a POST→redirect→fragment burst doesn't swallow the toast; toasts are suppressed for non-terminal statuses; `since` now emits `<time datetime data-elapsed>` and `app.js` recounts it once a second, with the JS `humanDuration` matching the Go one. `delay:300ms` correctly debounces a burst rather than throttling. All `{{since}}` callsites are element content, none in attributes, so the `template.HTML` return is safe. `docs/development.md` explains the design accurately. Build, vet, and `./internal/web ./internal/worker` tests pass locally.

One fix:

- `CHANGELOG.md:14` — the entry links `[#812](.../pull/812)`; this is PR #813.

Not blocking, just noting:

- `/` and `/scans` subscribe unscoped (`/events?events=scan-status`), so on a busy batch every open list tab re-fetches its table up to ~3×/sec. Fine for a single-operator install; if it ever matters, the fragment queries are already narrowed and indexed.

alexandre-daubois:
Changelog updated 👍

andrew (changes requested):
CHANGELOG link fixed — thanks.

This now conflicts with main in `internal/web/templates/jobs.html`: #829 (merged today) changed the skill dropdown button to use `{{.SkillLabel}}` with `aria-label="Skills: {{.Skill}}"` / `title="{{.Skill}}"`, and reworded the retry `hx-confirm`. This PR extracts most of that file into `job_list.html`, so the two collide.

Please rebase onto main and carry the #829 changes (`.SkillLabel`, `aria-label`/`title`, the reworded `hx-confirm`) into whichever of `jobs.html` / `job_list.html` holds the dropdown after the split.

andrew (approved):
Rebase carries #829's `.SkillLabel`, `aria-label`/`title`, and reworded `hx-confirm` into `job_list.html` verbatim; `SkillLabel` is set before the `isHX` branch so the fragment path gets it. `TestJobs_htmxFragmentKeepsTheSkillFilterLabel` covers the multi-skill label surviving a fragment render. Build/vet/tests pass.

PR #814, opened 2026-08-05
Prevent skills repository tokens from leaking through Git URLs

### LINKED ISSUE #759: skills_repo token-in-URL puts the credential in git argv
`-skills-repo` accepts `https://<token>@host/...` for private skill repos (`docs/skills.md:246`, `internal/skills/git.go` `ParseRepoSpec`). The token lands in git's argv on the host, so it is visible to `ps` and to anything that captures process listings. Repository URLs (#718) and schedule upstream URLs (#737) already reject embedded credentials for the same reason.

Replace with a `skills_repo_token` config key fed to git via `GIT_ASKPASS` or `http.extraHeader`, drop the `token@` mention from `docs/skills.md`, and have `ParseRepoSpec` reject userinfo to match #718/#737.

abhinavgautam01:
Fixes #759

## Summary

Replaces credential-bearing skills repository URLs with config-only token authentication through `GIT_ASKPASS`.

This prevents private repository credentials from appearing in Git command arguments, process listings, logs, or repository URLs.

## Changes

- Bump `github.com/alpha-omega-security/harness` to `v0.1.3`.
- Reject `skills_repo` HTTPS URLs containing userinfo.
- Add the config-only `skills_repo_token` setting.
- Authenticate remote skills repository operations through a temporary `GIT_ASKPASS` helper.
- Keep the token out of Git arguments and the askpass script itself.
- Restrict the authentication environment to remote Git operations.
- Disable ambient credential helpers when the explicit token is used.
- Reject multiline token values.
- Remove the temporary askpass helper after repository synchronization.
- Avoid echoing credential-bearing repository URLs in parser and startup errors.
- Document private skills repository configuration in the README, skills documentation and sample configuration.
- Add focused coverage for configuration loading, URL rejection, credential redaction and askpass behavior.

## Security Notes

`skills_repo_token` intentionally has no command-line flag because command-line values can be exposed through process listings.

Operators should keep `scrutineer.yaml` out of source control and restrict it to the owning user:

```bash
chmod 600 scrutineer.yaml
```
## Tests

- go mod tidy -diff
- go vet ./...
- go test -race ./...
- go vet -tags evals ./internal/evals/...
- go test -race -tags evals ./internal/evals/...
- go vet -tags podman ./...
- golangci-lint v2.12.2
- govulncheck ./...
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ The `docker build` commands shown for the runner image and profiles can be run a
 | `-data` | `./data` | Data directory for the database and workspaces |
 | `-effort` | `high` | Claude effort level (claude backend only) |
 | `-skills` | - | Additional local directory to load SKILL.md files from; same-named skills override the bundled copies (repeatable) |
-| `-skills-repo` | - | `owner/repo[@ref]` or git HTTPS URL `https://host/path[@ref]` to clone skills from on startup; `@ref` pins a branch, tag or commit and the resolved SHA is recorded on every scan |
+| `-skills-repo` | - | `owner/repo[@ref]` or credential-free git HTTPS URL `https://host/path[@ref]` to clone skills from on startup; `@ref` pins a branch, tag or commit and the resolved SHA is recorded on every scan; private repositories use config-file-only `skills_repo_token` |
 | `-backend` | `claude` | Agent CLI the container runner execs: `claude`, `codex`, or `opencode`. Non-claude backends require the containerised runner |
 | `--runtime` | `docker` | Container runtime: `docker`, `podman` (rootless podman supported), or `apple` (Apple, experimental) |
 | `--selinux` | `auto` | Bind-mount SELinux relabeling: `auto` (relabel when SELinux is detected), `on`, or `off` |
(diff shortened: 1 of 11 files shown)


andrew (changes requested):
Does what #759's checklist asks: bumps harness to v0.1.3 for the userinfo rejection, adds config-only `skills_repo_token`, wires it through `GIT_ASKPASS` with the token in env rather than the script body, updates docs. Verified against `git-pkgs/clone@v0.2.0`: `remoteEnv()` is the only source of `GIT_TERMINAL_PROMPT=0` and it's applied exactly to clone/fetch/ls-remote, so the sniff correctly scopes auth to remote commands. `-c credential.helper=` prepend is safe ahead of the existing `-C dst` in fetch args. Tests and lint pass locally.

Changes wanted:

- `cmd/scrutineer/main.go:384` — drop `mergeSkillsConfig` and inline the three assignments back into `merge()`. Every other field in that function (20+) is copied inline; extracting just the skills trio is inconsistent and adds a hop for no gain. The config-only comment can sit above the inline `SkillsRepoToken` block.
- `internal/skills/git.go:78` — the `GIT_TERMINAL_PROMPT=0` sniff works but couples to an undocumented `git-pkgs/clone` internal. Either add a one-line comment pointing at `clone.remoteEnv()` so a future clone bump that changes the env shape has a chance of being caught, or drop the filter entirely and pass the auth env on every call — a token in `git rev-parse`'s env is no more exposed than in `git clone`'s, and it removes the coupling.
- `internal/skills/git.go:63` — `os.CreateTemp("", ...)` writes to `$TMPDIR`; on hosts with `noexec` `/tmp` the askpass exec will fail. Not blocking if you'd rather not thread a dir through, but worth a note.

Nit:

- `internal/config/config_test.go:135` — `TestLoad_skillsRepoToken` only asserts YAML unmarshalling of a plain string field, which the stdlib already guarantees; `TestFlagsMerge_configFillsUnset` covers the meaningful path. Can drop it.

abhinavgautam01:
Thanks, addressed all of these and rebased onto the latest `main`.

- Removed `mergeSkillsConfig` and restored the three skills-related assignments inline in `flags.merge`, with the config-only token comment beside the assignment.
- Documented that the `GIT_TERMINAL_PROMPT=0` marker comes from `git-pkgs/clone.remoteEnv()` and is used to limit credentials to remote-touching commands.
- Moved the temporary askpass executable from `$TMPDIR` into the skills-cache parent directory so hosts with a `noexec` system temp directory still work. The test now pins that location as well as confirming the token stays out of the script and argv.
- Removed the redundant `TestLoad_skillsRepoToken` test; the meaningful config merge path remains covered.

Inlining the fields pushed the deliberately flat merge function over `maintidx`, so I extended its existing focused lint suppression to include that linter.

PR #824, opened 2026-08-06
evals: record security deep-dive prompt A/B results

abhinavgautam01:
Refs #118

## Summary

Records the live `claude-sonnet-5` A/B comparison between the production `security-deep-dive` prompt and the eval-only `reference-driven` candidate introduced by #793.

The candidate reduced model usage and cost, but regressed required-finding performance. Based on this run, the production prompt remains unchanged.

## Results

| Variant | Passed | Required misses | Unexpected | Cost | Turns |
| --- | ---: | ---: | ---: | ---: | ---: |
| production | 1/3 | 2 | 0 | $2.3552 | 64 |
| reference-driven | 0/3 | 3 | 0 | $1.4480 | 61 |

Compared with production, the candidate:

- Cost $0.9072 less, a 38.5% reduction
- Used 3 fewer turns
- Used 28,139 fewer output tokens
- Added one required miss
- Passed none of the three paired scenarios

The efficiency improvement does not justify the quality regression, so this run recommends against promoting the candidate.

## Changes

- Add a dated experiment report containing the tested revision, model, toolchain, per-scenario results, aggregate metrics, limitations and non-promotion decision.
- Link committed experiment decisions from `evals/README.md`.
- Document a two-hour Go test timeout for live model-backed eval commands.
- Leave the production `security-deep-dive` prompt unchanged.

## Live Evaluation

```sh
SCRUTINEER_RUN_EVALS=1 \
SCRUTINEER_EVAL_MODEL=claude-sonnet-5 \
go test -timeout 2h -tags evals ./internal/evals/... -run TestRunFixtures -v
```

All six paired deep-dive scenarios completed without runner errors. The overall command exited non-zero because assertion misses are expected to fail the eval and the unrelated standalone Semgrep fixture reached its 30-turn cap.

## Verification

- `go mod tidy -diff`
- `go vet ./...`
- `go vet -tags podman ./...`
- `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false go test -race ./...`
- `golangci-lint run`
- `git diff --check`

### DIFF
--- a/evals/README.md
+++ b/evals/README.md
@@ -48,6 +48,10 @@ baseline and candidate are directly comparable. For example, the
 `reference-driven`; the latter keeps intent and phase order in `SKILL.md` and
 loads sink taxonomy and report policy from `references/` when needed.
 
+Committed experiment decisions live under [`results/`](results/). Each report
+records the tested revision, model, environment, aggregate metrics, and whether
+the candidate was promoted.
+
 Each `should_find` or `should_not_find` assertion may include
 `evidence_contains`:
 
--- a/evals/README.md
+++ b/evals/README.md
@@ -12,7 +12,7 @@ The actual model-backed skills remain off by default. To run them against every
 fixture, opt in explicitly:
 
 ```sh
-SCRUTINEER_RUN_EVALS=1 SCRUTINEER_EVAL_MODEL=claude-sonnet-5 go test -tags evals ./internal/evals/... -run TestRunFixtures -v
+SCRUTINEER_RUN_EVALS=1 SCRUTINEER_EVAL_MODEL=claude-sonnet-5 go test -timeout 2h -tags evals ./internal/evals/... -run TestRunFixtures -v
 ```
 
 Each scenario YAML names:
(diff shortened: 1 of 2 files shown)


andrew (changes requested):
Report checks out — revision `233ae7cb` is real, the six paired fixtures exist, `SummarizeExperiments` is where the report says, and the Limitations section is honest about n=3 on one stochastic run. Decision follows from 1/3 vs 0/3 required-assertion passes.

Two things before merge:

- `Closes #118` should be a reference, not a close. The report itself says a future candidate should recover the missed assertions before another promotion decision, and the Limitations section says this run can't conclude a reference-driven prompt won't work after refinement. One negative candidate is a data point on #118, not a resolution.
- Fix the documented command rather than only recording the workaround. `evals/README.md:15` (and the block at :77-82) is the command #118 cites, and this run shows it doesn't survive a live model invocation without `-timeout`. Add `-timeout 2h` to the README commands; the Setup paragraph in the results file can then drop the two sentences about the aborted first attempt.

abhinavgautam01:
Thanks, fixed.
The PR now references #118 without closing it, both documented live-eval commands include `-timeout 2h` and the results setup only records the successful invocation.

andrew (approved):

PR #825, opened 2026-08-06
feat: show Disclosure title

jamietanna:
The title of the Finding may be different to the resulting Disclosure,
so we should make sure it's visible under 'Disclosure fields', otherwise
the user needs to find the scan and/or check the Notes.

Co-authored-by: Claude Sonnet 5 <jamie.tanna+claude-code@mend.io>

### DIFF
--- a/internal/web/templates/finding_show.html
+++ b/internal/web/templates/finding_show.html
@@ -674,6 +674,10 @@ git commit -m "fix: {{$f.Title}}"</pre>
     </summary>
     <section class="pb-4">
       <form method="post" action="/findings/{{$f.ID}}/fields" class="form grid grid-cols-2 gap-3">
+        <div class="grid gap-1 col-span-2">
+          <label class="text-xs text-muted-foreground">Title</label>
+          <input class="input" name="title" value="{{$f.Title}}">
+        </div>
         <div class="grid gap-1">
           <label class="text-xs text-muted-foreground">Severity</label>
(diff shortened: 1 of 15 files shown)


internal/web/templates/finding_show.html
```diff
@@ -545,6 +545,10 @@ <h3 class="flex flex-1 items-center justify-between gap-4 py-4 text-sm font-medi
     </summary>
     <section class="pb-4">
       <form method="post" action="/findings/{{$f.ID}}/fields" class="form grid grid-cols-2 gap-3">
+        <div class="grid gap-1 col-span-2">
+          <label class="text-xs text-muted-foreground">Title</label>
+          <input class="input" name="title" value="{{$f.Title}}">
```
jamietanna: This is the finding Title, not the Disclosure

jamietanna:
(This doesn't currently work)

andrew (changes requested):
The new `disclosure_title` field is documented in `ImportMinimal`, but the bundle exporter and `internal/ingest` minimal shape omit it. An `include=all` export/import therefore silently loses the disclosure title. Wire the field through both paths and add a round-trip test. Also refactor `findingFieldAccessor`: the added branch raises its cyclomatic complexity to 31 and the lint job fails at the configured limit of 30. Add coverage that `parseDiscloseOutput` stores the GHSA summary and that an analyst form edit persists it.

Issue #774, opened 2026-07-29
Per-scan and per-group cost cap with cooperative stop


andrew:
`Scan.CostUSD` and token counts are recorded (`internal/db/db.go:325-332`) and surfaced in the UI, and #37 covers measuring and documenting expected ranges. Nothing enforces a limit, so a runaway deep-dive on a huge repo or a saturation loop from #20 spends until the plan cap or turn limit, and the operator finds out from the bill.

Add an optional `max_cost_usd` on the scan request and on `ScanGroup`, checked by the worker between harness turns:

- running estimate summed across every scan in the group (a focus-area fan-out or saturation loop shares one budget)
- when the estimate crosses the cap, the worker signals the harness to stop after the current turn rather than killing mid-turn, so the partial report is written
- the scan finishes with `Scan.Status = done`, #771 `completeness = partial`, `reason = cost_cap`, and the report ingested as-is
- scan page and `/scans` list show configured limit, actual cost, and stop reason
- unset means no cap (current behaviour)

This is deliberately kept out of #37, which says hard budgets are out of scope there. #309's output-filtering experiments reduce spend per turn; a cap bounds the worst case regardless.

Source: [openai/codex-security cost tracker](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/cost.ts) counts parent and delegated worker sessions together, estimates while running, stops after a configured limit, and preserves partial results.

Related: #37 (measurement), #309 (per-turn reduction), #20 (saturation is the main consumer of a group budget), #771 (records the partial-coverage reason).

andrew:
Once #20's work manifest exists, `reason = cost_cap` at the scan level can be more specific: on stop, the worker walks the manifest and marks every unstarted unit `cost_capped` and the in-flight unit `partial` in #771's `receipts[]`. A later resume (or a fresh scan with `parent_scan_id` set per #773) then has an exact list of unfinished units to pick up rather than re-deriving scope from `completeness = partial` alone.

[VVAH's per-backend caps](https://github.com/visa/visa-vulnerability-agentic-harness/blob/d91b28d3af7ee46f9cc1827a9377cbdedd1f6b7e/vvaharness/cli.py#L59-L94) are a negative example here — its turn and cost limits don't apply consistently across the backends it supports, which is what putting the check in the worker (not the harness) avoids.

Issue #772, opened 2026-07-29
Skill capability preflight before spending model turns


andrew:
`verify` reports `inconclusive` with an `env-blocked:` note when the container profile lacks a runtime the reproduction needs (`skills/verify/SKILL.md:42`, `skills/verify/schema.json:39`). That costs a full model run to discover something the worker could check in one `exec` before starting.

Skill frontmatter already has `scrutineer.requires` (prerequisite skills), `scrutineer.requires_profile`, and `scrutineer.requires_remote` (`internal/skills/parse.go:35-39`). Extend it with:

- `scrutineer.requires_commands`: list of executables that must be on `PATH` in the runner (`cargo`, `bundle`, `node`, `z3`)
- `scrutineer.requires_features`: named runtime capabilities the profile must provide (`docker-in-docker`, `fuse`, `network-egress`)
- `scrutineer.degraded_mode`: optional; if set, the skill can run without the above and will report reduced coverage instead of failing

The worker runs the check inside the runner container after staging but before the first harness turn, and records the result in the #771 coverage contract as `preflight: {status, missing[], degraded}` where status is `ready` / `degraded` / `blocked`. A `blocked` result fails the scan immediately with the missing list rather than spending turns; `degraded` runs with `completeness` capped at `partial`.

Source: [openai/codex-security config preflight](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/references/config-preflight.md), which returns `ready` / `warn` / `block` / `incomplete` after checking tools, plugin skills, runtime features, and worker capacity. Scrutineer needs a smaller version since profile selection already handles most of this.

Related: #771 (records the preflight result), #499 (touches the same staging path).

andrew:
The static checks above catch a missing `cargo` on `PATH` but not a backend that rejects the tool set the harness will send it, or a credentials/base-URL misconfiguration that only surfaces on a real request. [visa/visa-vulnerability-agentic-harness](https://github.com/visa/visa-vulnerability-agentic-harness/blob/d91b28d3af7ee46f9cc1827a9377cbdedd1f6b7e/vvaharness/orchestrator/preflight.py#L226-L315) covers that with a live 1-turn probe per unique `(model, backend, allowed-tools)` tuple, exercising the same `agentic()` path the scan will use so a `NotImplementedError` on tool rejection or an argv regression in the CLI backend fails at preflight rather than mid-scan.

[...]

Issue #771, opened 2026-07-29
Typed coverage contract for `Scan.Coverage`


andrew:
`Scan.Coverage` is a text column holding an ad-hoc JSON blob (`internal/db/db.go:284`). The only producer is `diffCoverage` in `internal/worker/diff_rescan.go:24` with three fields: `requested_mode`, `actual_mode`, `fallback_reason`. Full and focus-area scans write nothing, so the UI can't distinguish "no findings after a complete scan" from "no findings in the part that finished", and #594's per-file receipts have nowhere structured to land.

Replace the blob with a typed contract written by every scan mode:

- `requested_mode`, `actual_mode` (kept as-is)
- `completeness`: `complete` / `partial` / `unknown`, with `reason` required whenever it isn't `complete`
- `included_paths`, `excluded_paths`, `deferred_paths` (each with a reason on excluded/deferred)
- `receipts[]`: `{path, disposition, reason}` per file or work item, dispositions `reviewed` / `supporting` / `deferred` / `excluded`; #594's diff-rescan reconciliation is one producer, focus-area scans another
- `surfaces[]`: `{name, disposition, evidence_ref}` for reviewed attack surfaces (deep-dive's phase-1 sink classes)
- `open_questions[]`: things the skill couldn't establish and a later scan should resolve

`completeness` gets its own indexed column on `Scan` so the scans list can filter on it. `internal/worker` derives `completeness` from receipt reconciliation on ingest rather than trusting the skill's own claim; a scan with unreceipted work items in scope is `partial`.

Source: [openai/codex-security `coverage.json`](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/references/scan-contract.md) and the [CLI's incomplete-coverage error path](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/cli.ts), which refuses to treat incomplete coverage as a clean policy result even when no finding crosses the severity threshold.

Related: #594 (diff receipts feed this), #20 (saturation needs it to prove what each pass reached), #14 (a versioned local coverage record is the source for federated clean-certificate claims).

andrew:
Two additions from [visa/visa-vulnerability-agentic-harness](https://github.com/visa/visa-vulnerability-agentic-harness/tree/d91b28d3af7ee46f9cc1827a9377cbdedd1f6b7e):

[...]

PR #828, opened 2026-08-07
Typed coverage contract for Scan.Coverage

charlie-morrison:
Closes part of #771 — the self-contained slice I described in https://github.com/alpha-omega-security/scrutineer/issues/771#issuecomment-5217502449. Happy to reshape it if you'd rather the split ran differently.

`Scan.Coverage` was an ad-hoc JSON blob with **two** writers putting different shapes into it — `diffCoverage` in `internal/worker/diff_rescan.go` and `markThreatModelUpdate` in `internal/web/diff_threat_model.go`, which merged three keys through `map[string]any`. A consumer could not tell "no findings after a complete scan" from "no findings in the part that finished".

**What this does**

- Adds `internal/coverage` with the typed record and an indexed `Scan.Completeness` column, so the scans list can filter without decoding the blob. `setCoverage` writes both together so the column and the record cannot drift apart.
- Moves both existing writers onto the record. The `threat_model_update` / `threat_model_material` / `threat_model_update_reason` keys become a `threat_model` sub-record rather than being dropped: "the model wasn't updated because the diff was too small" is a completeness statement, so it belongs inside the contract. (Those keys had no reader outside their own test, so dropping them was available — I don't think it's right.)
- `Parse` reads **both** legacy shapes back, so rows written before this keep their meaning, and a row that made no completeness claim reads `unknown` instead of being guessed at.
- `Reconcile` derives completeness in the worker from receipts against the scope the worker staged itself. **With no scope of its own it returns `unknown`, never `complete`, however many receipts a skill supplied** — that refusal is the point, and `TestReconcileWithoutScopeNeverClaimsComplete` pins it. A receipt is also not the same as finished work, so `deferred` / `failed` / `cost_capped` keep a scan `partial` while `excluded` settles it.

**What is deliberately not here, and why**

- `surfaces[]`, `dropped_findings[]`, `open_questions[]` and `receipts[]` are **declared but written by nobody**. They can only come from the skill, and per finding (2) in the issue comment, `coverage_metadata_key` is advertised at `skill.go:1351` and never ingested anywhere — making that channel real is slice 2. I'd rather ship the shape fixed and say so than populate these from guesses.
- `preflight` (#772), `reason: cost_cap` (#774) and manifest-derived per-unit receipts (#20) belong to their own issues.
- Diff staging starts at `partial` with `included_paths` set and no receipts, which is the truthful state at staging time. The fallback path records `unknown`, since a full scan has no enumerable scope yet.
- I left `internal/web/api.go` emitting `coverage` as a string. Turning it into an object is right (finding 4) but it's a breaking API change, so I'd rather you called it than have it ride along here.

`go build`, `go vet` and `gofmt` are clean. `internal/coverage` passes, as do the `Diff`/`Rescan`/`ThreatModel`/`Scan` tests in `internal/worker` and `internal/web`. One caveat I'd rather state than paper over: `internal/db`'s `TestWriteFindingField_concurrentUpdatesKeepHistoryChain` failed for me with `SQLITE_BUSY` — but it also fails intermittently on unmodified `main` on this machine (1 failure in 4 runs there, 3 in 4 on the branch), so I can't tell from here whether that's my change or an underactuated box. Nothing in the diff touches findings or that write path; CI is the authority. If it's red there I'll fix it in this PR.

### DIFF
--- a/internal/coverage/coverage.go
+++ b/internal/coverage/coverage.go
@@ -1,0 +1,259 @@
+// Package coverage defines the typed contract stored in Scan.Coverage: what
+// a scan did and did not reach, and how confident the worker is in that.
+//
+// The column was an ad-hoc JSON blob with two owners writing different shapes
+// into it (the diff-rescan staging path in internal/worker and the
+// threat-model update path in internal/web), so a consumer could not tell
+// "no findings after a complete scan" from "no findings in the part that
+// finished". A Record is that distinction, written by every scan mode.
+//
+// Completeness is derived by the worker from receipt reconciliation against
+// the scope it staged itself, never from the skill's own claim about how much
+// it covered. Where the worker has no independently known scope, the honest
+// answer is CompletenessUnknown, not CompletenessComplete.
+package coverage
+
+import (
+	"encoding/json"
+	"sort"
+	"strings"
+)
+
+// Version is the contract version stamped on every Record written. Consumers
+// that need to reason about older rows compare against it; rows written
+// before the typed contract carry 0.
+const Version = 1
+
+// Completeness states how much of the intended scope this scan actually
+// reached. Anything other than Complete requires a Reason.
+const (
+	// CompletenessComplete means every work item in the staged scope has a
+	// receipt. Only the worker may assign it, and only for a mode where it
+	// enumerated the scope itself.
+	CompletenessComplete = "complete"
+	// CompletenessPartial means the scope was known and some of it was not
+	// reached.
+	CompletenessPartial = "partial"
+	// CompletenessUnknown means no scope was enumerable, so no claim can be
+	// made. This is the correct value for a full-repository scan until a
+	// work manifest exists, and it is deliberately not a synonym for
+	// "probably fine".
+	CompletenessUnknown = "unknown"
+)
+
+// Dispositions record what happened to one work item. reviewed_clean and
+// reviewed_findings are both completed work; the rest are all reasons a unit
+// did not produce a verdict, kept apart so a partial scan can say which kind
+// of gap it means.
+const (
+	DispositionReviewedClean    = "reviewed_clean"
+	DispositionReviewedFindings = "reviewed_findings"
+	DispositionSupporting       = "supporting"
+	DispositionDeferred         = "deferred"
+	DispositionExcluded         = "excluded"
+	DispositionFailed           = "failed"
+	DispositionCostCapped       = "cost_capped"
+)
+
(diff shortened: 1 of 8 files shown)


andrew (changes requested):
Matches the #771 spec and reads both legacy blob shapes back correctly. `Reconcile` refusing `complete` without a worker-owned scope is the right property to pin. CI green on both matrices; also builds, vets, lints and passes on a local merge into current main after #829/#830/#832.

One fix before merge:

**`internal/web/diff_threat_model.go:115-117`** — `coverage.Marshal` normalizes `Completeness` on a value-copy, so when `Parse` returned `ok=false` (empty prior coverage — a non-diff threat-model scan hits this) the caller's `rec.Completeness` is still `""`. `scan.Completeness = rec.Completeness` then writes `""` to the column while the stored blob carries `"completeness":"unknown"`. `setCoverage` at `internal/worker/diff_rescan.go:315` guards this; `markThreatModelUpdate` doesn't. Add the same default before the DB write so the column and record can't disagree.

Two non-blocking notes:

- `scanDiffView.Completeness`/`CompletenessReason` at `internal/web/scans.go:157-158` aren't rendered by any template — unrendered fields on a view struct.
- Your design comment's step 1 said "API emits an object"; the PR body backs that out as a breaking change. Fine to leave for a separate PR, just noting the delta.

charlie-morrison:
Fixed in bfa2e7c — defaulted on the record before the write, the way `setCoverage` does, so the column and the blob come from one value. Regression test covers the empty-prior-coverage path.

The two unrendered view fields are for the scan-page follow-up; happy to drop them from this PR instead if you'd rather they land with their template. API-shape delta noted, leaving it for its own PR.

charlie-morrison:
`build (amd64)` went red on infrastructure, not on the change: `sum.golang.org` returned `INTERNAL_ERROR` while verifying `git-pkgs` checksums in the `go-tools` Docker stage. arm64 built and every test job passed on the same commit. Happy to rebase to retrigger it if that's easier than a re-run.

andrew (approved):
bfa2e7c does what was asked — `rec.Completeness` is defaulted before `Marshal` so the column and the blob come from one value, and `TestMarkThreatModelUpdateKeepsCompletenessColumnInStep` pins the drift condition directly.

`build (amd64)` was `sum.golang.org` returning `INTERNAL_ERROR` in the go-tools Docker stage, not this change; re-triggered.

alexandre-daubois (approved):

alexandre-daubois:
Thank you @charlie-morrison!

PR #831, opened 2026-08-09
Add deterministic verification grading rubric

### LINKED ISSUE #111: Tighten `verify` into a real grading rubric
`verify` currently re-runs the reproduction and reports reproduced / fixed / could-not-reproduce. That's one bit of information and it trusts the original scan's workspace.

Replace the freeform check with an explicit rubric, each criterion recorded pass/fail with evidence:

1. PoC is well-formed (the script/input parses, required files exist)
2. Reproduces on 3 of 3 attempts (flaky findings get a lower score, not a pass)
3. The failure is the claimed class, not an OOM, timeout, or assertion the finding didn't mention
4. The crash/trace lands in first-party code, not in a dependency or the test driver
5. Behaviour is deterministic (same input, same crash site)

Score 0.0 to 1.0 across the five. Store the score on the finding alongside the existing status enum.

Run the grade in a fresh checkout, not the find scan's leftover workspace. The find agent may have edited files, installed things, or left state that makes a non-bug reproduce. The only thing that should cross from the original scan to the verify run is the PoC bytes and the reproduction command. Relates to #3.

Optionally accept `--voters N` to run N independent verify agents and take the majority verdict, for findings where an analyst wants higher confidence before spending disclosure effort.

abhinavgautam01:
Fixes #111

## Summary

Adds a deterministic grading rubric to finding-scoped verification runs so reproduction results are backed by repeatable evidence rather than a single model verdict.

Each verification now runs the supplied reproduction three times, grades five fixed criteria, derives a score and stores the complete result as an append-only record associated with the finding and scan.

## Changes

- Update the bundled `verify` skill to require three isolated reproduction attempts.
- Grade every verification against five fixed criteria:
  - PoC is well formed.
  - Finding reproduces in all three attempts.
  - Observed behavior matches the claimed failure class.
  - A public interface reaches a first-party sink.
  - Reproduction is deterministic.
- Record the method, evidence, counterevidence, proof gap and confidence for every criterion.
- Add explicit `not_attempted` handling for setup, build, runtime and harness failures.
- Enforce status, attempt, and criterion consistency beyond the JSON Schema constraints.
- Add append-only `finding_verifications` storage keyed by finding and scan.
- Derive and persist a normalized score from the five criterion verdicts.
- Preserve compatibility with legacy pre-rubric verification reports.
- Make verification parsing transactional and idempotent for queue redelivery.
- Display the latest verification result and complete verification history on finding pages.
- Show criterion grading and all three attempt records with their evidence.
- Expose the latest verification record through the finding API.
- Update the bundled schema, OpenAPI specification, database documentation and skill documentation.
- Add coverage for validation, persistence, legacy reports, idempotency, schema rejection, API responses and UI rendering.

## Tests

- `go test -race ./...`
- `go test ./internal/verification ./internal/db ./internal/worker ./internal/web`
- `go vet ./...`
- `go vet -tags podman ./...`
- `golangci-lint run`
- `git diff --check`


charlie-morrison:
Read through this against the verify path — the fixed-shape rubric and `counterevidence`/`proof_gap` are a real improvement over a single model verdict. One robustness question about the failure path, which I don't think the existing escape hatch covers.

`ErrMissingRubric` deliberately tolerates the pre-rubric report so queued scans survive the upgrade (`internal/verification/report.go:15-17`), and `decodeVerifyOutput` returns `(result, nil, nil, nil)` for it — the run still records its status and prose, just without a score. But a report that *has* a rubric and fails `Validate()` returns an error, and `parseVerifyOutput` propagates it before reaching `recordVerifyOutput` — so the verification row, the finding note and the lifecycle transition are all dropped, and the `status`/`evidence`/`reproducer` the pre-rubric implementation would have stored go with them.

The part I'd weigh: the schema-repair loop in `repairAndParseSkillOutput` will catch most malformed output first, since this PR also grows `skills/verify/schema.json`. What it can't catch is precisely the class `Validate()` was added for — as the doc comment says, "the cross-field invariants JSON Schema cannot express". A `confirmed` where four of five rows pass, or a `fixed` whose `reproduces_three_of_three` isn't `fail`, is schema-valid, survives repair, and then discards the whole run. That's model output judging its own consistency across ~40 fields, so it seems like the steady-state failure rather than an edge case — and the more evidence the run gathered, the more is lost.

Would it fit your design to treat a consistency failure the same way as a missing rubric — store the report and status, leave `score` nil, and put the validation error in the note — so the rubric degrades to "ungraded" instead of the run failing? That keeps the invariants strict (an invalid rubric is never scored or displayed as graded) without making a three-attempt reproduction unrecoverable because one enum disagrees.

Not a blocker, and possibly deliberate — if the intent is that an inconsistent rubric *should* fail loudly so it gets fixed rather than silently landing ungraded, that's a coherent choice, just worth stating somewhere since it's a behaviour change from the current parser.

abhinavgautam01:
Thanks, this was a valid robustness gap.

The schema already rejects the two concrete examples in the comment through its conditional rules, but schema-valid semantic failures were still possible, including duplicate attempt numbers and whitespace-only rubric evidence.

I updated the flow so verify-specific semantic validation now runs through `ValidateSkillReport`, which gives the existing repair loop a chance to correct those cross-field inconsistencies.

If the report remains inconsistent after repair, Scrutineer now preserves the complete report, raw status, reproduction and evidence as an append-only verification row with a null score. The finding note records `grading: ungraded` and the validation error, the UI labels the result as ungraded and the report is prevented from changing the finding lifecycle.

Legacy pre-rubric reports retain their existing compatibility behavior. I also added tests covering semantic repair and the unrepaired fallback using duplicate attempt numbers.

andrew:
Thanks both. Could we tone down the volume and formality of the AI-style review/comments here?

Short, direct comments are much easier for me to review. You don’t need to restate the architecture, explain every trade-off, or provide a full summary of changes after each iteration. If there’s a bug or design concern, just point to it and explain it briefly.

More generally, I’d much rather have a few high-signal comments than a long agent-to-agent review thread that I then have to review myself.

abhinavgautam01:
Yep, Sorry for that @andrew
I will keep that in mind.

andrew (changes requested):
This matches the #111 spec closely (five fixed criteria, per-criterion `{method, evidence, counterevidence, proof_gap, confidence}`, `not_attempted` split, public-interface tier rule, append-only `finding_verifications` keyed on `{finding_id, scan_id}`, derived score) and correctly leaves out `control_bypass` and voters.

One bug blocks this:

**`internal/web/finding_verification.go:40-41`** — `loadFindingVerificationViews` returns an error for any row whose `Report` fails `verification.Parse` with something other than `ErrMissingRubric`. An ungraded row (the case commit 06a4ed2 exists to preserve) has `criteria` present, so it isn't `ErrMissingRubric`; it fails `Validate()` instead. `findingShow` swallows that into a warning log and renders `verifications` as nil, so one ungraded row hides every graded row in the history section. Repro:

```go
func TestLoadFindingVerificationViews_toleratesUngraded(t *testing.T) {
    // seed one FindingVerification with duplicate attempt numbers (schema-valid, semantics-invalid)
    // loadFindingVerificationViews returns:
    //   parse verification 1: attempts[1].number 1 is not unique in 1..3
}
```

Any per-row parse failure should render like a legacy row (`ScoreLabel: "ungraded"`, `HasRubric: false`) and continue the loop.

Two smaller points, non-blocking:

- `reportValidationForParsing` in `internal/worker/skill.go`: the schema's `allOf/if/then` blocks already enforce nearly everything `validateStatusConsistency` re-checks. A report violating those fails schema, which is `recoverable=false`, so under `SchemaStrict` it still fails hard. The only semantics-only rule left is attempt-number uniqueness, so the recoverable/ungraded path fires in one narrow case. Consider dropping the redundant Go checks and keeping only what the schema can't express.
- `skills/verify/schema.json` drops most `description` fields. The model reads the schema as well as SKILL.md; the old `preflight.justification` description ("the exact lines from the reproduction that decided the classification, quoted verbatim") is worth keeping in both.

abhinavgautam01:
Thanks, fixed
Invalid stored rubrics now render as ungraded per row instead of hiding valid verification history, with regression coverage for the duplicate-attempt case.

I also reduced the Go semantic checks to attempt-number uniqueness and restored the useful schema field guidance. Full tests, race tests, vet and lint pass.

andrew (approved):

PR #839, opened 2026-08-09
feat(web): consolidate the repository tab menu

### LINKED ISSUE #838: Redesign the Repository tabs
This is getting quite unwieldy 😅 

Image

connorshea:
AI Disclosure: Generated with Claude Code, Opus 5. Reviewed and tested by me. Needs a bit more review before I mark it undraft.

The repository page had grown to 15 tabs in a single flat row. This reshapes that into six tabs plus a grouped **More** overflow menu. Fixes #838.

| | |
|---|---|
| Tab bar | Summary · Findings ③ · Scanners ⑥ · Supply chain · Scans · Chat · More ⌄ |
| More › Analysis | Threat model · Workbench *override* · Benchmark |
| More › Disclosure | Maintainers · Advisories |
| More › Config | Scan config *set* · Data |

Screenshot 2026-08-09 at 3 22 03 PM

**Supply chain** is one tab holding a nested bar for Packages / Dependencies / Dependents / Alternatives. The panels move verbatim — the only edit is the `hidden` attribute, since any of the four can be absent and the first available one has to open with the parent. They keep their original indentation so the move reads as a move rather than a rewrite.

**Selection logic.** basecoat only tracks the tabs of a single tablist, so a panel opened from the More menu would have lingered on screen next to the next tab clicked. `selectTab()` in `app.js` now settles selection for tabs, menu entries, and nested bars alike, running after basecoat and reconciling what it missed — including the arrow-key path, hooked via the `focusin` that follows basecoat's own move. Deep links resolve through nesting, so `#rt15` opens Supply chain *and* its Alternatives sub-tab.

**Styling.** The bar moves from basecoat's segmented pills to underlined tabs, scoped to this page. The nested supply-chain bar keeps the pill look so the two levels stay distinguishable. Selection reads from colour and the rule rather than a heavier font weight, so moving between tabs cannot shift the row.

**Drive-by fix.** `rt15`/`rp15` was serving as the id for both Alternatives and Chat — a duplicate that made the `/repositories/N/alternatives` redirect ambiguous. Chat moves to `rt16`/`rp16`; every existing `#rt*` redirect in the Go handlers still lands correctly.

### DIFF
--- a/internal/web/static/app.js
+++ b/internal/web/static/app.js
@@ -17,19 +17,55 @@
     }
   }
 
+  // A panel can be opened from a tab in the tablist or from an entry in an
+  // overflow ("More") menu beside it, and a panel may hold a nested tab bar of
+  // its own. Basecoat only knows about the tabs of a single tablist, so
+  // selection is settled here; it runs after basecoat's own handler and
+  // reconciles whatever that missed.
+  function tabBar(tabs) {
+    return tabs.querySelector(':scope > .tab-bar') ||
+           tabs.querySelector(':scope > [role="tablist"]');
+  }
+
+  function selectTab(el) {
+    var tabs = el.closest('.tabs');
+    var bar = tabs && tabBar(tabs);
+    if (!bar) return;
+    var id = el.getAttribute('aria-controls');
+    tabs.querySelectorAll(':scope > [role="tabpanel"]').forEach(function (p) {
+      p.hidden = p.id !== id;
+    });
+    bar.querySelectorAll('[role="tab"]').forEach(function (t) {
+      var sel = t === el;
+      t.setAttribute('aria-selected', sel);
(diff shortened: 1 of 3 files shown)


connorshea:
@andrew don't bother looking at the code here yet as I haven't looked at it closely enough to mark it ready for review. But before I do that: does this look like a direction we're willing to go in, UI/UX-wise?

andrew (changes requested):
Selecting an item from More makes every primary repository tab `tabindex="-1"`. Basecoat then returns focus to the More trigger, and Tab or Shift+Tab skips the entire primary tablist, so keyboard users cannot return to Summary, Findings, Scans, or Chat. Keep a primary tab keyboard-reachable after an overflow selection and add a regression test for this focus path.

Issue #779, opened 2026-07-29
Add structured `controls` to the threat-model contract


andrew:
The threat-model schema records `trust_boundaries`, `entry_points` (with per-parameter `attacker_controllable` / `caller_must_enforce`), `adversaries`, `properties_provided`, and `known_non_findings` (`skills/threat-model/schema.json:14-58`). Design-level mitigations — an authenticated router in front of a package, a WASM sandbox around a parser, CSRF middleware on all mutating routes — appear only as prose on trust boundaries, so `verify` can't be told which controls apply to a finding's path and #725's severity caps have to trust the model's own precondition claim.

Add an optional top-level `controls[]`:

```json
{
  "id": "admin-auth",
  "kind": "authorization",
  "protects": {
    "paths": ["internal/admin/**"],
    "entry_points": ["admin-api"]
  },
  "assumptions": ["requests reach this package only through the authenticated router"],
  "provenance": "documented",
  "source": "docs/SECURITY.md:12"
}
```

`kind` is an open enum (`authorization`, `sandbox`, `input-validation`, `csrf`, `rate-limit`, `other`). `protects.paths` are globs matched against a finding's file; `protects.entry_points` reference `entry_points[].entry_point` names. `provenance` follows the existing `documented` / `inferred` convention.

Consumers:

- the worker glob-matches each finding's file against `controls[].protects.paths` and passes the matched set into `verify`
- #111's rubric gains a `control_bypass` row: for each matched control, either bypass evidence or a reason it doesn't apply; a matched control with neither caps the verdict at `uncertain`
- #725 reads matched-but-not-bypassed controls as attacker-prerequisite inputs (`kind: authorization` with no bypass → authenticated-user cap)
- #249 can replay a proposed control edit against #728's FP corpus and check whether it changes verifier verdicts

Source: [visa/visa-vulnerability-agentic-harness `Control` model](https://github.com/visa/visa-vulnerability-agentic-harness/blob/d91b28d3af7ee46f9cc1827a9377cbdedd1f6b7e/vvaharness/models.py#L151-L161) and its [verifier rule](https://github.com/visa/visa-vulnerability-agentic-harness/blob/d91b28d3af7ee46f9cc1827a9377cbdedd1f6b7e/vvaharness/pipeline/stages/s6_verify.py#L235-L276) that requires a demonstrated bypass of matched controls before returning `TRUE_POSITIVE`. Apache-2.0; the schema shape here is a reimplementation, not a copy.

[...]

PR #841, opened 2026-08-10
Stage matched threat-model controls for verify

charlie-morrison:
Second slice of #779, on top of #830 (contract, merged). #779 names `verify` as a consumer, but it had no channel to receive a control — this adds it.

- `internal/worker/controls.go` resolves the controls whose globs match the finding's file and stages them into `context.json`, gated on `verify` + finding-scoped runs.
- `skills/verify/SKILL.md` documents what a matched control does and does not license.

Two choices worth flagging:

- Matching happens host-side, mirroring `skillContextNovelty` — it needs the finding-path rebase (`noveltyFindingPath`, reused here) that a container-side skill can't do reliably. Only matched controls are staged, not the whole model, so a finding-scoped run can't re-derive the match in prose.
- A malformed controls block reports `unavailable_reason` instead of failing the scan; SKILL.md tells the skill to read that as *no information*, not as "nothing protects this".

No persistence and no new report field, so no `openapi.yaml`/`docs/database.md` change. The #111 rubric row I'd sketched as slice 3 landed in #831, so this stays clear of it.

`gofmt`/`build`/`vet` clean; `internal/worker`, `internal/skills`, `internal/threatmodel`, `internal/repoconfig` pass.

### DIFF
--- a/internal/findingnorm/normalize.go
+++ b/internal/findingnorm/normalize.go
@@ -52,3 +52,28 @@ func allDigits(s string) bool {
 	}
 	return true
 }
+
+// FindingPath rebases a finding's reported location into the repository-root
+// namespace that repository-wide documents — control globs, git pathspecs —
+// are authored in.
+//
+// A finding from a subpath-scoped scan reports its location relative to the
+// sub-folder, because the skill treats that folder as the project root. The
+// sub-folder is denormalised onto the finding row, so the two are joined
+// here rather than at each call site.
+//
+// An absolute path or a parent-segment escape yields "" rather than being
+// cleaned: a malformed location must produce no match instead of a match
+// against the wrong file.
+func FindingPath(subPath, location string) string {
+	file := LocationFile(location)
+	subPath = RepoPath(subPath)
+	if !validFindingPath(file) || (subPath != "" && !validFindingPath(subPath)) {
+		return ""
+	}
+	return path.Join(subPath, file)
+}
+
+func validFindingPath(p string) bool {
+	return p != "" && p != "." && !strings.HasPrefix(p, "/") && !HasParentPathSegment(p)
+}
--- a/internal/findingnorm/normalize_test.go
+++ b/internal/findingnorm/normalize_test.go
@@ -1,0 +1,34 @@
+package findingnorm
+
+import "testing"
+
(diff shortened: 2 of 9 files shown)


andrew (changes requested):
Two things before this lands:

**`scan.SubPath` is the wrong subpath.** `controlsContext` rebases with `scan.SubPath` (`controls.go:76`), but verify scans are enqueued with `SubPath = ""` — neither `enqueueFindingScopedSkillIfIdle` (`internal/web/server.go:3176`) nor `enqueueValidateFixVerify` (`internal/web/validate_fix_enqueue.go:113`) sets it. The finding's `Location` is relative to the *originating* audit scan's subpath, which is denormalised onto `finding.SubPath` (`internal/db/db.go:746`). Novelty reads it from the finding row (`novelty.go:96`); this needs to do the same: `Select("location", "sub_path")` and pass `finding.SubPath`. As written, a finding from a subpath-scoped audit silently matches nothing. `TestControlsContextRebasesSubpathLocation` passes because the fixture sets `scan.SubPath = "internal"` on the verify scan, which production never does — the test should set `finding.SubPath` on the fixture row instead.

**`controlsFindingPath` duplicates `noveltyFindingPath` byte for byte.** The comment even says so. Two callers of the same rebase-and-validate step should share one function; otherwise the next change to `validNoveltyPath` drifts. Rename `noveltyFindingPath` to something neutral (or push it into `findingnorm`) and call it from both places.

Everything else — the degrade-not-fail on a bad model, staging only the matched set, the SKILL.md guidance that a matched control is a claim rather than a verdict — is spot on.

charlie-morrison:
Both fixed in aa67ab6.

Subpath now comes from the finding row (`Select("location", "sub_path")`), like novelty does — the verify scan has none. The fixture sets it on the finding instead, and `resolve()` no longer takes a subpath at all, so the test can't re-create a state production never produces; it fails on the old code with `finding_file = "web/server.go"`.

Shared the rebase as `findingnorm.FindingPath` and dropped both copies. It is only a composition of that package's own primitives, so the reject-rather-than-clean contract is table-tested there now.

alexandre-daubois (approved):

andrew (approved):

PR #845, opened 2026-08-10
Use harness directive stripping

andrew:
Replaces scrutineer's local agent directive list and walker with `harness.StripDirectives` (harness v0.1.5).

Keeps the worker integration coverage, adds the shared `.opencode` path case, and pins `.git` preservation at this layer.


internal/worker/skill.go
```diff
@@ -1054,7 +1055,7 @@ func stripWorkspaceAgentDirectives(workRoot string, emit func(Event)) error {
 	// instructions into the auditing agent. Runs before the paths filter
 	// because scrutineer.paths bypasses BuiltinSkipPaths and must not be
 	// able to opt these back in. See threatmodel.md T5.
-	stripped, err := stripAgentDirectives(src)
+	stripped, err := harness.StripDirectives(src)
```
alexandre-daubois, line 1058: `harness.StripDirectives` has a bare `skills` in its `directiveDirs`, so every directory named `skills` in `./src` gets deleted from the audited tree. Scanning scrutineer itself, or any repo with a `skills/` folder, silently drops that code from the scan scope.

internal/worker/strip_test.go
```diff
@@ -1,171 +1,27 @@
 package worker
```
alexandre-daubois, line 1: Deleting all five strip tests leaves nothing pinning the T5 pattern set, which is how the `skills` entry slipped through. A "benign names survive" test over `harness.DirectivePaths()` would catch it.

PR #846, opened 2026-08-10
Use clone.Cache for exposure and repo-cache checkouts

andrew:
Replaces the hand-rolled dependent-cache and repo-cache paths with `git-pkgs/clone.Cache`, which already carries the sha256-per-URL layout, `Prepare` (fetch/reset, `RemoveAll(dst)`, `CopyTree`), `EnsureCommit`, and `DiskBytes`.

`prepareDependentSrc` now builds a `clone.Cache` and calls `Prepare` directly; `dependentCacheRoot` and its test are deleted. `RepoCacheRoot`, `RepoDiskUsage`, and `EnsureCommit` delegate to the same helpers under the existing per-URL worker lock. `EnsureCommit` lowercases the commit before delegating since `clone.ValidCommit` is stricter than `validGitOID`.

`prepareRepoSrcWithOptions` is left as-is: its `<sha>/with-submodules/src` layout does not match `Cache.Dir`, so folding it in would move existing submodule caches on disk.

The retry-notify closure duplicated in `cloneOrFetch` and `gitRetry.do` is extracted to `gitRetry.toCloneWithNotify`.

### DIFF
--- a/internal/worker/exposure.go
+++ b/internal/worker/exposure.go
@@ -2,8 +2,6 @@ package worker
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -12,18 +10,10 @@ import (
 	"path/filepath"
 	"sync"
 
+	"github.com/git-pkgs/clone"
+
 	"scrutineer/internal/db"
 )
-
-// dependentCacheRoot returns the shared on-disk path scrutineer reuses
-// across exposure scans of the same dependent URL. Keyed by sha256 of the
-// URL so different URLs cannot collide and the path is filesystem-safe.
-// The directory survives wrap()'s per-scan workspace cleanup, so the
-// second exposure scan on the same dependent only fetches the delta.
-func dependentCacheRoot(dataDir, url string) string {
-	sum := sha256.Sum256([]byte(url))
-	return filepath.Join(dataDir, "dependent-cache", hex.EncodeToString(sum[:]))
-}
 
 // cacheMutex returns the per-URL mutex used to serialise fetch+copy on
 // the dependent cache. Lazily created on first use.
@@ -42,23 +32,16 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
 	mu.Lock()
 	defer mu.Unlock()
 
-	cacheRoot := dependentCacheRoot(w.DataDir, url)
-	if err := os.MkdirAll(cacheRoot, dirPerm); err != nil {
-		return "", err
+	cache := clone.Cache{
+		Root:  filepath.Join(w.DataDir, "dependent-cache"),
+		Retry: gitRetry{}.toCloneWithNotify(emit),
 	}
-	cacheSrc, err := ensureClone(ctx, db.Repository{URL: url}, cacheRoot, false, ref, emit)
-	if err != nil {
-		return "", err
+	if _, err := os.Stat(filepath.Join(cache.Dir(url), "src", ".git")); err == nil {
+		emit(Event{Kind: KindText, Text: "$ git fetch origin " + fetchTarget(ref) + " && reset"})
+	} else {
+		emit(Event{Kind: KindText, Text: "$ git clone " + url + " (shallow)"})
 	}
-	commit := gitHead(cacheSrc)
-	dst := filepath.Join(workRoot, "src")
-	if err := os.RemoveAll(dst); err != nil {
-		return "", err
-	}
-	if err := CopyTree(cacheSrc, dst); err != nil {
-		return "", fmt.Errorf("copy dependent cache: %w", err)
-	}
-	return commit, nil
+	return cache.Prepare(ctx, url, ref, filepath.Join(workRoot, "src"))
 }
 
 // CopyTree recursively copies src to dst, preserving permissions but not
--- a/internal/worker/exposure.go
+++ b/internal/worker/exposure.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"sync"
@@ -28,6 +27,14 @@ func (w *Worker) cacheMutex(url string) *sync.Mutex {
 // is whatever the skill leaves behind. Returns the HEAD commit of the
 // freshly-synced cache.
 func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot string, emit func(Event)) (string, error) {
+	// Validate before emitting so a bad URL or ref does not log a
+	// "$ git clone" line for a command that will never run.
+	if err := clone.ValidateURL(url); err != nil {
+		return "", &RepoUnreachableError{URL: url, Err: err}
+	}
+	if err := clone.ValidateRef(ref); err != nil {
+		return "", err
+	}
 	mu := w.cacheMutex(url)
 	mu.Lock()
 	defer mu.Unlock()
@@ -47,47 +54,7 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
 // CopyTree recursively copies src to dst, preserving permissions but not
 // ownership or timestamps. Symlinks are recreated; everything else is
 // copied byte-for-byte. Fast enough for git trees up to a few hundred MB.
-func CopyTree(src, dst string) error {
-	return filepath.Walk(src, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
-		}
-		rel, err := filepath.Rel(src, path)
-		if err != nil {
-			return err
-		}
-		target := filepath.Join(dst, rel)
-		switch {
-		case info.IsDir():
-			return os.MkdirAll(target, info.Mode().Perm())
-		case info.Mode()&os.ModeSymlink != 0:
-			link, err := os.Readlink(path)
-			if err != nil {
-				return err
-			}
-			return os.Symlink(link, target)
-		default:
-			return copyFile(path, target, info.Mode().Perm())
-		}
-	})
-}
-
-func copyFile(src, dst string, perm os.FileMode) error {
-	in, err := os.Open(src)
-	if err != nil {
-		return err
-	}
-	defer func() { _ = in.Close() }()
-	out, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm)
-	if err != nil {
-		return err
-	}
-	if _, err := io.Copy(out, in); err != nil {
-		_ = out.Close()
-		return err
-	}
-	return out.Close()
-}
+func CopyTree(src, dst string) error { return clone.CopyTree(src, dst) }
 
 // doExposure runs the exposure skill for one (finding, dependent) pair.
 // The scan's Repository stays the library being audited; ./src in the
--- a/internal/worker/novelty_test.go
+++ b/internal/worker/novelty_test.go
@@ -150,6 +150,10 @@ func TestPrepareNoveltyHistoryDeepensSharedCache(t *testing.T) {
 	scan := fixture.scan(head)
 	scan.Repository.URL = repoURL
 	w := &Worker{DB: fixture.gdb, DataDir: dataDir}
+	// clone.Cache.EnsureCommit hardens fetch to https-only unless the
+	// caller's env already sets GIT_ALLOW_PROTOCOL; this test's cache
+	// origin is a file:// path.
+	t.Setenv("GIT_ALLOW_PROTOCOL", "file")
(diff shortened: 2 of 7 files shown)


internal/worker/exposure.go
```diff
@@ -42,23 +45,17 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
...
-	cacheSrc, err := ensureClone(ctx, db.Repository{URL: url}, cacheRoot, false, ref, emit)
-	if err != nil {
-		return "", err
-	}
-	commit := gitHead(cacheSrc)
-	dst := filepath.Join(workRoot, "src")
-	if err := os.RemoveAll(dst); err != nil {
-		return "", err
+	cache := clone.Cache{
+		Root:  filepath.Join(w.DataDir, "dependent-cache"),
```
andrew, line 43: This inlines what `dependentCacheRoot` computed, and nothing else on the branch calls that function now — only `TestDependentCacheRoot_keyedByURL` references it. Delete the function and its test; the surviving `a == cache.Dir(url)` assertion is testing `clone.Cache`'s contract, not scrutineer's. Copilot's stale-comment note on line 25 goes with it.

internal/worker/exposure.go
```diff
@@ -42,23 +45,17 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
...
+		Retry: gitRetry{}.toCloneWithNotify(emit),
```
andrew, line 44: `gitRetry{}` uses `resolved()` defaults. `cloneOrFetch` takes a caller-supplied `gitRetry`; if exposure previously threaded a worker-configured retry through `ensureClone`, this quietly reverts it to defaults.

internal/worker/exposure.go
```diff
@@ -42,23 +45,17 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
...
-	if err := CopyTree(cacheSrc, dst); err != nil {
-		return "", fmt.Errorf("copy dependent cache: %w", err)
+	if _, err := os.Stat(filepath.Join(cache.Dir(url), "src", ".git")); err == nil {
+		emit(Event{Kind: KindText, Text: "$ git fetch origin " + fetchTarget(ref) + " && reset"})
+	} else {
+		emit(Event{Kind: KindText, Text: "$ git clone " + url + " (shallow)"})
 	}
-	return commit, nil
+	checkout := dependents.CacheCheckout{Cache: &cache, Ref: ref}
+	return checkout.Prepare(ctx, url, filepath.Join(workRoot, "src"))
```
andrew: Two things dropped on the way here:

- `os.RemoveAll(dst)` — the old path cleared `workRoot/src` before `CopyTree`. `scanWorkRoot` survives resume lineages, so unless `Prepare` wipes `dst` itself a resumed exposure run inherits the previous checkout under a new HEAD.
- URL/ref validation — `ensureClone` took a `db.Repository{URL: url}` and applied the https-only / conservative-ref checks before shelling out; `Prepare` gets the raw ecosyste.ms string. If `clone.Ensure` enforces the same rules inside `CacheCheckout` this is fine; if not, the guard needs reinstating here or pushing into `git-pkgs/clone`.

andrew (commented):
Self-review notes before undrafting.

`dependents` v0.1.0 pulls in `git-pkgs/brief` and `BurntSushi/toml` as indirects. `toml` is stock; check whether `brief` is heavier than the checkout helper warrants.

internal/worker/repo_cache.go
```diff
@@ -2,26 +2,28 @@ package worker
...
 	"strings"
 
+	"github.com/git-pkgs/clone"
 	"gorm.io/gorm"
 
 	"scrutineer/internal/db"
 )
 
+func repoCache(dataDir string) *clone.Cache {
+	return &clone.Cache{Root: filepath.Join(dataDir, "repo-cache")}
```
alexandre-daubois: The `clone.Cache` built here carries no `Retry`, so the unshallow that `EnsureCommit` used to run as a single `git fetch` now runs under `clone`'s default 3-attempt policy with backoff. A remote answering `connection timed out` makes `repoBlob` re-download the full history twice more inside the HTTP handler while `w.cacheMutex(url)` stays held, and `cmd/scrutineer/main.go:723` sets no `WriteTimeout`, so a scan of the same repository queues behind that page load. You can pass `Retry: gitRetry{attempts: 1}.resolved().toClone()` to keep the previous budget and put these git calls back on `gitWithEnv`.

internal/worker/repo_cache.go
```diff
@@ -2,26 +2,28 @@ package worker
...
+}
+
 // RepoCacheRoot returns the persistent per-URL clone directory under
 // dataDir. The cache survives scan cleanup so subsequent scans only
 // fetch the delta; EnsureCommit deepens it on demand when the code
 // browser asks for a commit that the shallow clone doesn't have.
 func RepoCacheRoot(dataDir, url string) string {
-	sum := sha256.Sum256([]byte(url))
-	return filepath.Join(dataDir, "repo-cache", hex.EncodeToString(sum[:]))
+	return repoCache(dataDir).Dir(url)
```
alexandre-daubois, line 33: The on-disk layout now comes from `clone.Cache.Dir`, and `TestRepoCacheRoot` only checks determinism and the `repo-cache/` prefix, so nothing pins the `sha256(url)` mapping that `docs/skills.md:177` still promises. A `clone` release that changes `Dir` re-keys every cached checkout, so each repository re-clones from scratch and the orphaned directories survive the `os.RemoveAll(worker.RepoCacheRoot(...))` in `internal/web/server.go:3053`. Asserting the full expected path in `TestRepoCacheRoot` would surface that at bump time.

internal/worker/exposure.go
```diff
@@ -42,23 +32,16 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
...
-	dst := filepath.Join(workRoot, "src")
-	if err := os.RemoveAll(dst); err != nil {
-		return "", err
+	cache := clone.Cache{
+		Root:  filepath.Join(w.DataDir, "dependent-cache"),
+		Retry: gitRetry{}.toCloneWithNotify(emit),
 	}
-	if err := CopyTree(cacheSrc, dst); err != nil {
-		return "", fmt.Errorf("copy dependent cache: %w", err)
+	if _, err := os.Stat(filepath.Join(cache.Dir(url), "src", ".git")); err == nil {
```
alexandre-daubois, line 46: Rebuilding `<cache.Dir(url)>/src/.git` hardcodes `clone.Cache`'s internal layout for the sake of one log line, so a rename upstream flips every dependent scan to the `git clone` message with nothing to catch it. An accessor on `clone.Cache` for the checkout path, or letting the library emit the line itself, would keep that knowledge on one side of the boundary.

internal/worker/exposure.go
```diff
@@ -42,23 +32,16 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
...
-	}
-	cacheSrc, err := ensureClone(ctx, db.Repository{URL: url}, cacheRoot, false, ref, emit)
-	if err != nil {
-		return "", err
-	}
-	commit := gitHead(cacheSrc)
-	dst := filepath.Join(workRoot, "src")
-	if err := os.RemoveAll(dst); err != nil {
-		return "", err
+	cache := clone.Cache{
```
alexandre-daubois, line 42: A fresh `clone.Cache` per call allocates and discards `Prepare`'s per-URL mutex every time, so serialisation rests on `w.cacheMutex(url)` alone, and `repoCache` repeats the shape on `RepoCacheRoot`, which the code browser calls per request and `BackfillRepoDiskUsage` calls per repository row. Holding one `clone.Cache` per data dir on the `Worker` would drop the allocation and let the library's own lock carry the contract its doc comment describes.

internal/worker/exposure.go
```diff
@@ -42,23 +32,16 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
...
+	if _, err := os.Stat(filepath.Join(cache.Dir(url), "src", ".git")); err == nil {
+		emit(Event{Kind: KindText, Text: "$ git fetch origin " + fetchTarget(ref) + " && reset"})
+	} else {
+		emit(Event{Kind: KindText, Text: "$ git clone " + url + " (shallow)"})
 	}
-	return commit, nil
+	return cache.Prepare(ctx, url, ref, filepath.Join(workRoot, "src"))
 }
 
 // CopyTree recursively copies src to dst, preserving permissions but not
```
alexandre-daubois, line 54: This helper and the `copyFile` under it are byte-identical to `clone.CopyTree` and its own `copyFile`, and the file now imports `clone`, which `Cache.Prepare` already calls for the same copy. Collapsing the body to `return clone.CopyTree(src, dst)` would drop around 45 lines and leave `internal/evals/runner.go:70` and the four in-package callers untouched.

internal/worker/novelty_test.go
```diff
@@ -150,6 +150,10 @@ func TestPrepareNoveltyHistoryDeepensSharedCache(t *testing.T) {
 	scan := fixture.scan(head)
 	scan.Repository.URL = repoURL
 	w := &Worker{DB: fixture.gdb, DataDir: dataDir}
+	// clone.Cache.EnsureCommit hardens fetch to https-only unless the
+	// caller's env already sets GIT_ALLOW_PROTOCOL; this test's cache
+	// origin is a file:// path.
+	t.Setenv("GIT_ALLOW_PROTOCOL", "file")
```
alexandre-daubois, line 156: `GIT_ALLOW_PROTOCOL=file` replaces the whitelist instead of widening it, so every git call in this test runs with https refused, and this is the only test covering the unshallow path that the delegation changed. `file:https` as the value keeps the production protocol allowed while still letting the `file://` cache origin fetch.

internal/worker/exposure.go
```diff
@@ -42,23 +32,16 @@ func (w *Worker) prepareDependentSrc(ctx context.Context, url, ref, workRoot str
...
+		Root:  filepath.Join(w.DataDir, "dependent-cache"),
+		Retry: gitRetry{}.toCloneWithNotify(emit),
 	}
-	if err := CopyTree(cacheSrc, dst); err != nil {
-		return "", fmt.Errorf("copy dependent cache: %w", err)
+	if _, err := os.Stat(filepath.Join(cache.Dir(url), "src", ".git")); err == nil {
+		emit(Event{Kind: KindText, Text: "$ git fetch origin " + fetchTarget(ref) + " && reset"})
+	} else {
+		emit(Event{Kind: KindText, Text: "$ git clone " + url + " (shallow)"})
 	}
```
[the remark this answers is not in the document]
andrew: Done in b2e0452: `clone.ValidateURL`/`clone.ValidateRef` before the emit block, matching the ordering in `cloneOrFetch`. A bad URL wraps in `RepoUnreachableError` so `doExposure`'s existing check still upserts `under_investigation`.

internal/worker/repo_cache.go
andrew: Done in b2e0452: `repoCache()` now sets `Retry: gitRetry{attempts: 1}.toCloneWithNotify(nil)`, so the unshallow is one attempt via `gitWithEnv` again.

internal/worker/repo_cache.go
andrew: Done in b2e0452: `TestRepoCacheRoot` now asserts the exact `repo-cache/<sha256(url)>` path.

internal/worker/exposure.go
andrew: Done in b2e0452: `CopyTree` delegates to `clone.CopyTree`, `copyFile` and the `io` import gone.

internal/worker/exposure.go
andrew: Agreed on the coupling; the fix wants a `Cache.SrcDir(url)` (or similar) accessor on `git-pkgs/clone`, which I'll add there and pick up on the next bump. The failure mode is only that this cosmetic log line says "clone" instead of "fetch" if the library changes its layout, and `TestRepoCacheRoot` now pins the layout so a change there would fail CI first.

internal/worker/exposure.go
andrew: Not done in this PR. `RepoCacheRoot` and `RepoDiskUsage` are package-level (called from `internal/web` and `BackfillRepoDiskUsage` with a bare `dataDir`), so hanging the cache off `Worker` means either turning those into methods or a package-level singleton keyed by dir. The behaviour regression was the retry budget, which the `Retry` field on `repoCache()` fixes; I'd rather do the struct-on-Worker refactor as its own change.

internal/worker/novelty_test.go
andrew: Left as-is. The test's origin is a `file://` path and nothing in `prepareNoveltyHistory` reaches for https, so `file` alone is sufficient here; `file:https` would be harmless but doesn't change what the test exercises.

alexandre-daubois (approved):

PR #849, opened 2026-08-11
fix(web): make the disclosure draft reviewable before a finding is ready

p-linnane:
The disclose skill writes a GHSA-shaped draft to the finding's disclosure_draft column, but the finding page had nowhere to put it. The workflow kept offering "Draft disclosure" after a draft already existed, the only editor was a six-row textarea buried in the analyst fields form, and the Markdown exports rebuilt generic scan metadata instead of reading what was saved.

The finding page now carries a disclosure section with a preview/edit toggle, save and cancel, writing through the usual finding-field history. "Review disclosure" becomes the primary action once a draft is saved and "Mark ready" is gated on one in both the UI and the handler, regeneration drops to a secondary action that confirms before overwriting analyst edits and no longer enqueues a duplicate scan, and the finding and disclose-scan exports serve the saved draft, returning 404 when there is none.

Opus 5 was used to implement this.

### DIFF
--- a/internal/web/scan_report.go
+++ b/internal/web/scan_report.go
@@ -34,10 +34,8 @@ func (s *Server) scanReport(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Skill is looked up separately rather than via Preload because Scan.SkillID
-	// is nullable and Preload on a nullable FK is fiddly across GORM versions;
-	// a missing skill row (e.g. skill deleted after the scan ran) is non-fatal
-	// here, we just lose the OutputKind dispatch and fall back to raw.
+	// SkillID is nullable, so load the skill separately rather than with Preload.
+	// Missing rows are non-fatal; disclose scans can still dispatch by name.
 	var skill *db.Skill
 	if scan.SkillID != nil {
 		var sk db.Skill
@@ -46,7 +44,16 @@ func (s *Server) scanReport(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	body := renderScanReport(s.DB, &scan, skill)
+	var body string
+	if isDiscloseScanReport(&scan, skill) {
+		body = renderSavedDisclosureScanReport(s.DB, &scan)
+		if body == "" {
+			http.Error(w, "no saved disclosure draft to export", http.StatusNotFound)
+			return
+		}
+	} else {
+		body = renderScanReport(s.DB, &scan, skill)
+	}
 
 	w.Header().Set("Content-Type", "text/markdown; charset=utf-8")
 	w.Header().Set("Content-Disposition", `attachment; filename="`+scanReportFilename(&scan)+`"`)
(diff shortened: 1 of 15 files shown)


internal/web/static/app.js
```diff
@@ -93,7 +93,59 @@
...
+    if (disclosureCancel) {
+      var disclosureForm = disclosureCancel.closest('form');
+      var disclosureRoot = disclosureCancel.closest('[data-disclosure-editor]');
+      if (disclosureForm) disclosureForm.reset();
+      if (disclosureRoot) {
+        var previewButton = disclosureRoot.querySelector('[data-disclosure-mode="preview"]');
+        if (previewButton) previewButton.click();
+      }
+      return;
+    }
```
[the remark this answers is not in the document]
p-linnane: Fixed in c534d01.

internal/web/server.go
```diff
@@ -1497,6 +1502,10 @@ func (s *Server) findingStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	status := db.FindingLifecycle(r.FormValue(statusKey))
+	if status == db.FindingReady && strings.TrimSpace(f.DisclosureDraft) == "" {
+		http.Error(w, "a saved disclosure draft is required before marking ready", http.StatusUnprocessableEntity)
```
alexandre-daubois, line 1506: The gate is unconditional, but a repo with no dependents does an operator handoff, not a disclosure. "Mark ready" is its only exit and is now disabled + 422, so those findings are stuck in `triaged`. You should scope the gate to `.HasDependents`.

internal/web/scan_report.go
```diff
@@ -46,7 +44,16 @@ func (s *Server) scanReport(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	body := renderScanReport(s.DB, &scan, skill)
+	var body string
+	if isDiscloseScanReport(&scan, skill) {
+		body = renderSavedDisclosureScanReport(s.DB, &scan)
+		if body == "" {
+			http.Error(w, "no saved disclosure draft to export", http.StatusNotFound)
```
alexandre-daubois, line 51: A disclose scan with an empty draft now 404s on report.md even though the scan row holds a report, so a refused run can't be exported at all. You should fall back to `renderScanReport` instead of 404.

alexandre-daubois (commented):
A CHANGELOG.md entry should be added as well with the PR number

andrew (approved):

internal/web/server.go
p-linnane: This is intentional. “Mark ready” is gated on a saved draft for all findings, including operator-handoff cases with no recorded dependents. Those findings still expose the editor and instruct the analyst to save the handoff text, so they have a path forward. Scoping this to `HasDependents` would weaken the requested workflow contract.

internal/web/scan_report.go
p-linnane: Agreed that displaying an export action which returns 404 is a UX issue. Falling back to `renderScanReport` would restore the generic scan-metadata export that this change intentionally removed. I’ll handle this separately by hiding the export action for disclose scans without a saved draft, while retaining the explicit 404 for direct requests.

p-linnane: https://github.com/alpha-omega-security/scrutineer/pull/852

Issue #191, opened 2026-05-08
Group structurally identical findings


andrewmusselman:
Loving the tool overall — one more piece of feedback, on report formatting:

## Summary

Reports list every occurrence of the same rule against the same pattern as a separate finding, even when they're structurally identical. This inflates finding counts and makes it harder to assess actual scope.

## Example

A recent report on [`apache/tooling-trusted-releases`](https://github.com/apache/tooling-agents/tree/main/scrutineer/tooling-trusted-releases) listed 102 individual findings for `generic.html-templates.security.var-in-href`. 37 of them were in a single file (`includes/topnav.html`) — all instances of the same pattern (`<a href="{{ url_for(...) }}">`) being flagged repeatedly. The report partially acknowledged this by collapsing F56–F92 into a range, but the rest of the file-level repetition was left expanded, and the same pattern in other templates wasn't grouped at all.

A reviewer triaging this has to either scroll through 102 nearly identical entries or rebuild the grouping themselves.

## Suggested behavior

Group findings by `(rule_id, pattern_signature)` and present them as:

- One header per group with the rule and a representative snippet
- A collapsed list of file:line locations underneath
- A single triage note that applies to the whole group

Findings that share a rule but differ meaningfully in context (e.g. one uses a user-controlled variable, another uses a server-built URL) should stay separate. The grouping heuristic could be as simple as normalizing the matched AST and hashing it.

This would also make the Markdown table in the summary section more useful — counts per group rather than counts per rule.

**Suggested labels:** `enhancement`, `skill`

andrew:
Thanks for reporting, looking into this now

andrew:
@andrewmusselman https://github.com/alpha-omega-security/scrutineer/pull/193

PR #816, opened 2026-08-05
feat: update default model examples to the Claude 5 generation

alexandre-daubois:

PR #850, opened 2026-08-12
skills: add a poutine scanner skill for CI/CD pipeline findings

### LINKED ISSUE #593: Suggestion: integrate poutine Build Pipeline scanner
The project currently has a skill for `zizmor`, which is a really great scanner, but it's more like a pendantic linter than good at quickly spotting high risk issues. I am the maintainer of poutine (https://github.com/boostsecurityio/poutine). It comes with built-in rules, but it's very easy to customize custom rules, or at least focus on the more critical findings with the built in rules.

sujeito-operator:
Closes #593.

The `zizmor` skill covers GitHub Actions workflows and is, as the issue puts it, closer to a pedantic linter than a way to spot high-risk issues quickly. poutine is aimed at the other end — its `error`-level rules are things like arbitrary code execution from an untrusted checkout — and it also reads GitLab CI, Azure Pipelines and Tekton, which nothing here covers today.

I'm not the poutine maintainer; @fproulx-boostsecurity was asked in the thread and the issue has sat unassigned for a month, so this is offered as a starting point rather than as the canonical version. Happy to hand it over or rework it.

### Shape

It mirrors the zizmor skill rather than inventing a second shape: same frontmatter, same `schema.json`, same `scripts/scan.py` layout, same grouping behaviour. `scan.py` runs `poutine analyze_local . --format json` and groups by rule, so one rule firing across several jobs becomes a single finding carrying every `file:line` in `locations` — the behaviour added for zizmor in #191.

### Two decisions worth a reviewer's attention

**Rule metadata lives outside the finding.** poutine emits rule metadata once in a top-level `rules` map and each finding carries only `rule_id`. Title, severity and references are read from that map. A finding naming a rule the map doesn't describe still produces a finding (titled with the raw rule id) rather than being dropped — a location silently discarded is a vulnerability not reported.

**Nothing maps to `Critical`.** poutine's levels are SARIF's, so `error`/`warning`/`note` → `High`/`Medium`/`Low`. poutine never emits `critical`, and promoting its `error` into scrutineer's `Critical` would let it outrank a finding another scanner genuinely rated critical. Easy to change if you'd rather it did.

### Verification

Against poutine's own `scanner/testdata` fixtures, the adapter turns 50 raw findings into 11 grouped ones:

| severity | rule | locations |
|---|---|---|
| High | Arbitrary Code Execution from Untrusted Code Checkout | 13 |
| High | If condition always evaluates to true | 1 |
| Medium | Injection with Arbitrary External Contributor | 8 |
| Medium | Workflow job exposes all secrets | 2 |
| Medium | Default permissions used on risky events | 3 |
| Medium | Build Component with a Known Vulnerability used | 3 |
| Medium | Pull Request Runs on Self-Hosted GitHub Actions Runner | 3 |
| Low | CI Runner Debug Enabled | 5 |
| Low | Github Action from Unverified Creator used | 5 |
| Low | Unpinnable CI component used | 2 |
| Low | Unverified Script Execution | 3 |

That output validates against `skills/poutine/schema.json`. Run against this repository, poutine reports one finding (`github_action_from_unverified_creator_used`, `note`, on the `golangci-lint-action` pin in `tests.yml`).

`internal/worker/poutine_script_test.go` follows `zizmor_script_test.go`: a fake `poutine` on `PATH`, asserting the rule grouping, the title and severity lookup through the `rules` map, and that a finding poutine reports with no line keeps a bare path instead of a fabricated `:0`. The empty paths — no `./src`, no pipeline files, poutine missing from `PATH`, malformed output, non-zero exit — were each exercised by hand and all return an empty findings set with an `error` string rather than failing the scan.

### Wiring

- Installed in the `go-tools` stage of both Dockerfiles; the existing `COPY --from=go-tools /out/*` picks it up, so no new `COPY` line.
- poutine stamps its version through `-ldflags`, so a plain `go install` would leave `poutine version` reporting `development` and the settings row blank. The install line sets it, and the renovate rule matches **both** halves (module pin and ldflags stamp) with `minimumGroupSize: 4` so they can't drift apart.
- `poutine version` rather than `--version` — there is no such flag. In a subshell the three-line block flattens and `versionRe` picks the version out of the first token, which is why it slots into `queryToolsScript` unchanged.
- Findings are excluded from exposure analysis alongside zizmor's: exposure asks whether a finding is reachable from outside, and a pipeline-level finding has no answer to that.

### CI, run locally

When I opened this the caveat here was that I had no Go toolchain and could not run the test suite. That is no longer true, so replacing it with results. Everything below was run against `b7edb8f` on go1.26.5, the version `go.mod` and both Dockerfiles pin.

| CI step | result |
|---|---|
| `scripts/release-resolver_test.sh` | 13/13 |
| `scripts/wait-for-runner-image_test.sh` | 42/42 |
| `go mod tidy -diff` | clean |
| `go vet ./...` | clean |
| `go test ./...` | 16 packages ok, 0 fail |
| `go test -race ./...` | 16 packages ok, 0 data races |
| `go vet -tags evals ./internal/evals/...` | clean |
| `go test -race -tags evals ./internal/evals/...` | ok |
| `go vet -tags podman ./...` | clean |
| `golangci-lint run` | 0 issues |
| `govulncheck ./...` | 0 called; 1 in a required module this code doesn't reach |
| `zizmor --persona auditor .github/` | no findings |

`zizmor` is the one worth calling out, since this PR edits `runner-image.yml` and that job would be the first to object.

**Still not verified by me: the two image builds.** No container runtime here. The part of that diff that can actually fail is the one `go install` line, so I ran it directly rather than leaving it as an assertion — `GOBIN=/out go install -ldflags "-X main.version=1.1.6" github.com/boostsecurityio/poutine@v1.1.6` succeeds on go1.26.5 and the resulting binary reports `Version: 1.1.6`, so the `main.version` ldflags path is right and the settings row won't read `development`.

### One thing I had wrong

Running it turned up an error in my own comment. I wrote that the subshell flattens `poutine version`'s three-line output. It doesn't — command substitution strips only the *trailing* newline, so `Commit:` and `Built At:` arrive at `parseToolVersions` as their own lines. The parsed result is still correct, because neither line carries an `=` so both are skipped and the version rides the first line, but that is a different mechanism than the one I documented and nothing tested it.

`b7edb8f` corrects the comment and adds `TestParseToolVersions_poutineVersionBlock`, built from verbatim poutine v1.1.6 output. It asserts the version parses *and* that poutine's extra lines don't swallow the `semgrep` and `harness` keys echoed after it — which is the failure that would actually have mattered.

### DIFF
--- a/.github/workflows/runner-image.yml
+++ b/.github/workflows/runner-image.yml
@@ -78,6 +78,7 @@ jobs:
           docker run --rm scrutineer-runner:ci curl --version
           docker run --rm scrutineer-runner:ci semgrep --version
           docker run --rm scrutineer-runner:ci zizmor --version
+          docker run --rm scrutineer-runner:ci poutine version
           docker run --rm scrutineer-runner:ci git-pkgs --version
           docker run --rm scrutineer-runner:ci brief --version
           docker run --rm scrutineer-runner:ci git --version
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Entries are grouped by release, newest first. Each entry is a summary written fo
 ## Unreleased
 
 - A `default_model` pinned in the config file but missing from the model pick list is now reported at startup instead of being silently ignored, and the documentation and sample configuration now use current-generation model ids. ([#816](https://github.com/alpha-omega-security/scrutineer/pull/816), [@alexandre-daubois](https://github.com/alexandre-daubois))
+- Added a `poutine` scanner skill, so pipeline supply-chain weaknesses — untrusted-checkout code execution, injection, secrets exposed to a whole job — are reported alongside the existing zizmor workflow lint, and GitLab CI, Azure Pipelines and Tekton repositories get pipeline coverage for the first time. ([#593](https://github.com/alpha-omega-security/scrutineer/issues/593))
 
 ## 2026-08-11
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,10 @@ FROM golang:1.26.5-alpine@sha256:0178a641fbb4858c5f1b48e34bdaabe0350a330a1b1149a
 RUN apk add --no-cache git
 RUN GOBIN=/out go install github.com/git-pkgs/git-pkgs@v0.15.3 && \
     GOBIN=/out go install github.com/git-pkgs/brief/cmd/brief@v0.9.3
+# poutine stamps its version through -ldflags, so a plain `go install` would
+# leave `poutine version` reporting "development" and the settings page blank.
+RUN GOBIN=/out go install -ldflags "-X main.version=1.1.6" \
+    github.com/boostsecurityio/poutine@v1.1.6
 
 # vid links tree-sitter grammars (C), so unlike the main binary it needs
 # cgo; build-base provides gcc and musl headers, matching the musl-based
--- a/Dockerfile.runner
+++ b/Dockerfile.runner
@@ -16,6 +16,10 @@ ARG SEMGREP_VERSION=1.167.0
 FROM golang:1.26.5-trixie@sha256:4ee9ffa999b4583ce281939cdff828763083610292f252279a0cee77473bd9a7 AS go-tools
 RUN GOBIN=/out go install github.com/git-pkgs/git-pkgs@v0.15.3 && \
     GOBIN=/out go install github.com/git-pkgs/brief/cmd/brief@v0.9.3
+# poutine stamps its version through -ldflags, so a plain `go install` would
+# leave `poutine version` reporting "development" and the settings page blank.
+RUN GOBIN=/out go install -ldflags "-X main.version=1.1.6" \
+    github.com/boostsecurityio/poutine@v1.1.6
 
 FROM rust:1.97-trixie@sha256:1bcff4befb740599103a2c7cb51058e14479b2e35e3a34a3f0dc4ede09927488 AS zizmor-build
 ENV CARGO_HTTP_MULTIPLEXING=false
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ To onboard a whole GitHub org at once, open **Add multiple** → **Import a whol
 
 You can also scan a directory on disk, useful before pushing, or for code not hosted on a git forge. Paste an absolute path (`/path/to/project`) in the same **Add repository** field. Scrutineer copies the directory into a per-scan workspace and runs the default skill set; skills that need a forge URL or ecosyste.ms enrichment (`advisories`, `exposure`, `fork`, `maintainers`, `metadata`, `packages`, `public-issue`, `report-upstream`) are skipped automatically. Symlinks are recreated as-is rather than dereferenced during the copy; in container mode their targets then resolve inside the container, so host files reached only through such a link are not visible to skills. Under `--no-container` the kernel dereferences them normally, so only point scrutineer at trees you trust.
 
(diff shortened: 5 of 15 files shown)


sujeito-operator:
The checks on this PR are sitting at `action_required` — first-time contributor, so the workflows need a maintainer to click approve before any of them run. That leaves you with no CI to review against, which is a bad trade for whoever picks this up.

So I ran your CI locally instead and folded the results into the description. Short version, against `b7edb8f` on go1.26.5: `go test ./...` and `go test -race ./...` both green across all 16 packages with no data races, `golangci-lint` 0 issues, `govulncheck` clean, `go mod tidy -diff` clean, both `scripts/*_test.sh` green, and `zizmor --persona auditor .github/` reports no findings — that last one matters most here since this PR edits `runner-image.yml`.

The image builds I still can't verify; there's no container runtime on my side. The one line in that diff that can genuinely fail is the `go install`, so I ran it on its own: it builds under the same go1.26.5 the Dockerfiles pin, and the binary reports `Version: 1.1.6`, so the `-X main.version` path is correct.

Doing that also caught something I'd got wrong. My comment on `queryToolsScript` claimed the subshell flattens `poutine version`'s three-line output — it doesn't. Command substitution strips only the trailing newline, so `Commit:` and `Built At:` reach `parseToolVersions` as separate lines. The value still parses correctly, for a different reason than the one I wrote down (neither extra line has an `=`, so they're skipped), and nothing covered it. `b7edb8f` fixes the comment and adds a test from real poutine output that pins both the version and the thing that would actually hurt — that poutine's extra lines don't swallow the `semgrep` and `harness` keys echoed after it.

No rush on my end, and no need to approve the workflows on my account if you'd rather not — I'd just rather you evaluate this against real results than against a pending check.

sujeito-operator:
Rebased onto `main` (now `8a1f8a4`). #816 landing added a `## Unreleased` entry, which put this branch into a conflicted state and made it unmergeable; the only conflict was that CHANGELOG section and both entries are kept, mine second.

The matrix from the comment above was re-run on the rebased tree rather than assumed:

| check | result on `8a1f8a4` |
| --- | --- |
| `go build ./...`, `go vet ./...` | clean |
| `go test ./...` | 16 packages ok, 0 fail |
| `go vet`/`go test -tags evals ./internal/evals/...` | clean, ok |
| `go vet -tags podman ./...` | clean |
| `go mod tidy -diff` | clean |
| `zizmor --persona auditor .github/` | no findings |
| `scripts/release-resolver_test.sh` | 13/13 |
| `scripts/wait-for-runner-image_test.sh` | 42/42 |

Not re-run after the rebase, in the interest of being exact about it: `go test -race ./...` and `golangci-lint run`, both of which were clean on `b7edb8f` and touch no code that the rebase changed. The container builds are still the thing I cannot run here at all.

sujeito-operator:
No problem either way — flagging only in case this was swept rather than decided.

#593 was closed as inactive about forty minutes after this branch was rebased onto `8a1f8a4` and the CI matrix went up, and #805 — same ticket, different contributor — went the same way. So the ticket has now had two independent implementations and neither was read.

If poutine isn't something you want beside zizmor any more, that's a completely fine answer and I'll stop here. If the issue just aged out of triage, the branch is still up at `sujeito-operator:poutine-skill` at `8a1f8a4` and still reads mergeable; I'm happy to reopen and keep it current against `main`.

One thing worth knowing whichever way you go: the workflows on this PR never ran. They sat at `action_required` the whole time because I'm a first-time contributor, which is why I reproduced your matrix locally and posted it — but it does mean nothing here was ever evaluated by your CI.

PR #853, opened 2026-08-13
Process scheduled repositories with bounded concurrency

### LINKED ISSUE #815: Process Due Scheduled Repositories with Bounded Concurrency
### Summary

Use bounded concurrency when processing repositories due for scheduled scans so a slow or unavailable Git remote does not delay every subsequent repository in the scheduler tick.

### Affected Code

- `internal/web/schedule.go`
- `scheduleTick`
- `runScheduledScan`

### Current Behavior

`scheduleTick` loads all repositories whose next scheduled time is unset or due, then processes them sequentially:

```go
for _, repo := range repos {
  // Resolve schedule, optionally run the scan, and advance its due time.
}
```

For a due repository, `runScheduledScan` may perform several remote Git operations:

- resolve the staging repository's remote HEAD;
- resolve the upstream repository's remote HEAD;
- clone or reuse a persistent bare mirror;
- fetch the upstream;
- force-push the synchronized branch;
- resolve the final remote HEAD before deciding whether to enqueue scans.

Remote Git retries are bounded by attempt count and backoff, but the commands inherit the scheduler's long-lived server context and currently have no per-repository deadline. A slow remote, DNS failure, stalled transport or forge outage can therefore delay every due repository behind it.

### Impact

When many repositories become due at the same time, especially under a global daily or weekly schedule, scheduler latency grows with the sum of every repository's remote-operation latency. Healthy repositories may not be processed until long after their intended schedule because an unrelated remote is slow.

### Proposed Improvement

Process due repositories through a bounded worker pool or an `errgroup` with a small concurrency limit, for example five repositories at a time.

Each repository's unit of work should include:

1. Resolving its effective schedule.
2. Running the scheduled remote synchronization and HEAD comparison when due.
3. Advancing `next_scheduled_scan_at` exactly once.
4. Logging repository-specific errors without cancelling unrelated work.

Add a per-repository timeout around remote scheduler work as a complementary safeguard. Bounded concurrency prevents one slow repository from blocking the entire list, while the timeout prevents all worker slots from remaining occupied indefinitely.

### Correctness Requirements

- Keep concurrency bounded to avoid overwhelming Git hosts or increasing SQLite write contention.
- Preserve `lockRepoFederation` behavior and repository-specific cache locks.
- Do not run the same repository more than once in a tick.
- Advance each schedule once regardless of whether the run enqueues, skips or fails in the same cases as today.
- Wait for all work from the current tick before returning, so scheduler ticks do not overlap unexpectedly.
- Keep individual repository failures isolated and visible in logs or skipped scan records.

### Suggested Tests

- Prove that no more than the configured number of repository jobs run at once.
- Block one fake remote and verify another due repository still progresses.
- Verify schedule advancement remains correct under concurrent completion.
- Verify cancellation and per-repository timeout release worker slots.
- Run with the race detector to cover scheduler state and test stubs.

abhinavgautam01:
Fixes #815

## Summary

Processes repositories due for scheduled scans through a bounded worker pool so a slow or unavailable Git remote does not delay every subsequent repository in the scheduler tick.

The scheduler derives its worker limit from the operator-configured queue concurrency. Remote HEAD probes receive a scoped timeout, while potentially large mirror clone, fetch and push operations retain the scheduler's parent context.

## Changes

- Process scheduled repositories concurrently using the queue's configured concurrency limit.
- Fall back to one scheduler worker when no queue is configured.
- Apply a 10-minute timeout only to remote HEAD probes.
- Apply the probe timeout to staging/upstream HEAD comparisons and the final repository HEAD lookup.
- Keep mirror clone, fetch and push operations outside the probe timeout so large initial clones are not cancelled and restarted on every tick.
- Preserve existing schedule resolution, firing and advancement behavior.
- Advance each repository's `next_scheduled_scan_at` exactly once after a due run, including skipped and failed runs.
- Preserve repository federation locking and Git cache locking.
- Keep failures and probe timeouts isolated so one repository does not cancel unrelated work.
- Wait for all dispatched repository work before completing the scheduler tick.
- Add deterministic tests covering:
  - Enforcement of the configured queue concurrency.
  - A timed-out remote HEAD probe releasing its worker slot.
  - Remote probe timeout isolation.
  - Mirror operations not inheriting the remote HEAD timeout.
  - Exactly-once schedule advancement.

## Tests

- `go mod tidy -diff`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `golangci-lint run`
- `git diff --check`

### DIFF
--- a/internal/web/schedule.go
+++ b/internal/web/schedule.go
@@ -29,10 +29,9 @@ const (
 	// tick interval only bounds the firing latency, not the cadence.
 	schedulerTick = time.Minute
 
-	// Keep remote Git work parallel enough that one slow host does not block
-	// every due repository, but bounded to limit forge and SQLite pressure.
-	schedulerMaxConcurrentRepositories = 5
-	schedulerRepositoryTimeout         = 10 * time.Minute
+	// Bound remote HEAD probes without also placing large first-time mirror
+	// clones under a deadline that would make them restart on every tick.
+	schedulerRemoteHeadTimeout = 10 * time.Minute
 )
 
 // ScheduleNext validates a scan-schedule value, the "daily"/"weekly"
@@ -91,7 +90,17 @@ func (s *Server) scheduleTick(ctx context.Context, now time.Time) {
 		s.Log.Error("scheduler: list repositories", "err", err)
 		return
 	}
-	s.runScheduledRepositories(ctx, now, global, repos, schedulerMaxConcurrentRepositories, schedulerRepositoryTimeout)
+	s.runScheduledRepositories(ctx, now, global, repos, s.schedulerConcurrency(), schedulerRemoteHeadTimeout)
+}
+
+// schedulerConcurrency keeps scheduled Git work under the same operator-set
+// concurrency budget as scan jobs. Tests may construct a Server without a
+// queue, in which case one worker preserves forward progress.
+func (s *Server) schedulerConcurrency() int {
+	if s.Queue == nil {
+		return 1
+	}
+	return max(1, s.Queue.Concurrency())
 }
 
 func (s *Server) runScheduledRepositories(
@@ -100,7 +109,7 @@ func (s *Server) runScheduledRepositories(
 	global string,
 	repos []db.Repository,
 	maxConcurrent int,
-	timeout time.Duration,
+	remoteHeadTimeout time.Duration,
 ) {
 	if len(repos) == 0 {
 		return
@@ -119,16 +128,20 @@ func (s *Server) runScheduledRepositories(
 	for range min(maxConcurrent, len(repos)) {
 		workers.Go(func() {
 			for repo := range jobs {
-				repoCtx, cancel := context.WithTimeout(ctx, timeout)
-				s.processScheduledRepository(repoCtx, now, global, repo)
-				cancel()
+				s.processScheduledRepository(ctx, now, global, repo, remoteHeadTimeout)
 			}
 		})
 	}
 	workers.Wait()
 }
 
-func (s *Server) processScheduledRepository(ctx context.Context, now time.Time, global string, repo db.Repository) {
+func (s *Server) processScheduledRepository(
+	ctx context.Context,
+	now time.Time,
+	global string,
+	repo db.Repository,
+	remoteHeadTimeout time.Duration,
+) {
 	expr := repo.ScanSchedule
 	if expr == "" {
 		expr = global
@@ -150,7 +163,7 @@ func (s *Server) processScheduledRepository(ctx context.Context, now time.Time, 
 		return
 	}
 	if repo.NextScheduledScanAt != nil {
-		s.runScheduledScan(ctx, repo)
+		s.runScheduledScanWithRemoteHeadTimeout(ctx, repo, remoteHeadTimeout)
 	}
 	if err := s.DB.Model(&db.Repository{}).Where("id = ?", repo.ID).
 		UpdateColumn("next_scheduled_scan_at", next).Error; err != nil {
@@ -164,6 +177,14 @@ func (s *Server) processScheduledRepository(ctx context.Context, now time.Time, 
 // diff-rescan group (which falls back to full coverage when no baseline
 // exists, e.g. on a never-scanned repository).
 func (s *Server) runScheduledScan(ctx context.Context, repo db.Repository) {
+	s.runScheduledScanWithRemoteHeadTimeout(ctx, repo, schedulerRemoteHeadTimeout)
+}
+
+func (s *Server) runScheduledScanWithRemoteHeadTimeout(
+	ctx context.Context,
+	repo db.Repository,
+	remoteHeadTimeout time.Duration,
+) {
 	// Held for the whole firing, so the opt-out cannot commit between the check
 	// below and the network calls it gates: recording one waits here, then sweeps
 	// whatever this firing enqueued. See lockRepoFederation.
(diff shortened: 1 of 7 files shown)


internal/web/schedule.go
```diff
@@ -130,9 +174,8 @@ func (s *Server) runScheduledScan(ctx context.Context, repo db.Repository) {
 	// and the remote HEAD lookup have already contacted their host, which is
 	// the other half of what the opt-out asks us not to do. Read live rather
 	// than off the tick's snapshot: the tick loads every due repository up
-	// front and then fires them one at a time, each doing its own network I/O,
-	// so an opt-out recorded mid-tick has to be seen by the repositories still
-	// waiting their turn.
+	// front, so an opt-out recorded while its bounded worker is waiting has to
+	// be seen before that repository performs network I/O.
```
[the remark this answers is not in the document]
abhinavgautam01: I reworded the comment to describe the repository as waiting in the shared bounded work queue, rather than implying that a per-repository worker has already been assigned.

internal/web/schedule.go
```diff
@@ -22,10 +23,17 @@ const ScheduleOff = "off"
...
+const (
+	// schedulerTick is how often the scheduler re-evaluates every repository's
+	// schedule. Due-ness is driven by the persisted NextScheduledScanAt, so the
+	// tick interval only bounds the firing latency, not the cadence.
+	schedulerTick = time.Minute
+
+	// Keep remote Git work parallel enough that one slow host does not block
+	// every due repository, but bounded to limit forge and SQLite pressure.
+	schedulerMaxConcurrentRepositories = 5
+	schedulerRepositoryTimeout         = 10 * time.Minute
```
alexandre-daubois: `schedulerRepositoryTimeout` also covers `syncUpstream`'s first bare mirror clone, whose duration tracks repository size rather than remote health, and `Retry.Do` runs `cloneDestReset` on the context error so the partial clone is removed. A repository whose first sync doesn't fit the budget restarts from zero every tick, never completes, and accrues one `skipped` row per firing with no way to raise the limit. The deadline is better scoped to the `ls-remote` calls and the HEAD comparison, or exposed as configuration.

internal/web/schedule.go
```diff
@@ -22,10 +23,17 @@ const ScheduleOff = "off"
...
 const scheduleKind = "schedule"
 
-// schedulerTick is how often the scheduler re-evaluates every repository's
-// schedule. Due-ness is driven by the persisted NextScheduledScanAt, so the
-// tick interval only bounds the firing latency, not the cadence.
-const schedulerTick = time.Minute
+const (
+	// schedulerTick is how often the scheduler re-evaluates every repository's
+	// schedule. Due-ness is driven by the persisted NextScheduledScanAt, so the
+	// tick interval only bounds the firing latency, not the cadence.
```
alexandre-daubois, line 29: `schedulerMaxConcurrentRepositories` is fixed at 5 independently of the operator's `-concurrency` budget, so those five slots run `git clone`/`fetch` against the same disk and network as the scan pool: a host configured with `-concurrency 1` ends up with six concurrent git operations. Deriving it from `Queue.Concurrency()` the way `chatSlots` does would keep one knob in charge, and failing that the ceiling belongs on the `-concurrency` row in `README.md`.

internal/web/schedule.go
abhinavgautam01: Thanks, fixed. The deadline no longer wraps the complete scheduled repository operation. It now bounds only the remote HEAD probes, including the staging/upstream comparisons performed before synchronization and the final repository HEAD lookup.

Mirror clone, fetch and push operations retain the scheduler's parent context, so a large initial clone is not cancelled, deleted and restarted on every tick. I added regression coverage confirming that HEAD probes receive the deadline while mirror operations do not.

internal/web/schedule.go
abhinavgautam01: I removed the fixed five-worker limit. Each scheduler tick now derives its worker count from `Queue.Concurrency()`, with a one-worker fallback for tests or servers constructed without a queue.

The concurrency test now reconfigures the queue and verifies that the scheduler's observed peak matches that configured budget.

andrew (changes requested):
The pool and probe-scoped timeout are correct now that Alexandre's two points are applied. What's left:

1. `deadcode ./...` reports three unreachable functions after this change: `Server.runScheduledScan` (schedule.go:179), `Worker.SyncUpstream` (upstream.go:66), and `Worker.syncUpstream` (upstream.go:83). The `...WithHeadTimeout` variants were layered alongside the originals instead of replacing them. Add the `headTimeout time.Duration` parameter to the existing three functions and drop the wrappers; the four test call sites in `upstream_retry_test.go` pass `0`. That takes upstream.go back to two functions instead of four and removes the shim in schedule.go.
2. `var cancel context.CancelFunc = func() {}` is duplicated in schedule.go and upstream.go with manual `cancel()` on each return path. In schedule.go nothing after `resolveRemoteHead` uses `remoteCtx`, so `defer cancel()` right after the conditional assignment is simpler and drops the no-op closure. Same in upstream.go once the error-path `cancelHead()` is a defer.
3. `TestRunScheduledRepositories_slowRepoDoesNotBlockHealthyRepo` runs two repos with `maxConcurrent=2`; both starting concurrently is a given. It exercises the same thing the ceiling test already proves and doesn't compile on main, so it's not a regression test. Drop it.
4. `schedulerRemoteHeadTimeout` is a hard-coded 10m with no config knob. If it's staying fixed, the constant's comment should say why 10m rather than what it doesn't cover.

abhinavgautam01:
Thanks, addressed all four points.

- Removed the `runScheduledScan`, `SyncUpstream` and `syncUpstream` compatibility wrappers by adding the timeout parameter directly to the canonical functions. The affected test call sites now pass either the scheduler timeout or `0` explicitly.
- Replaced the duplicated no-op cancellation functions and manual cleanup paths with conditional `defer cancel()` calls.
- Removed `TestRunScheduledRepositories_slowRepoDoesNotBlockHealthyRepo`; the configured concurrency ceiling test already covers the relevant worker-pool behavior.
- Reworded the timeout comment to explain why the fixed ten-minute budget exists: it accommodates transient retry backoff and slow forge responses while preventing an unreachable remote from retaining a scheduler worker indefinitely.

The three newly introduced dead-code findings are gone.

andrew (approved):

PR #854, opened 2026-08-13
fix(worker): don't count sibling focus-area scans as missed observations

### LINKED ISSUE #835: Remove the "not seen x6" marker on Findings
Now that findings are discovered in parallel batched scans w/ focus areas, this signal is basically useless and should be removed from the UI.

charlie-morrison:
Closes #835 by fixing the count rather than removing the marker, per your comment.

`markNotObserved` (`internal/worker/skill.go:813`) scoped by `(repository_id, skill_name, sub_path)` with no `focus_area` term, so each of N parallel focus-area deep-dives bumped `MissedCount` on every open finding from the other N-1. With `AutoRejectMissedCount` set, one batch auto-rejects valid findings — `TestParseFindingsOutput_focusAreaScanDoesNotAutoReject` fails that way on main.

Took the second option (skip when `scan.FocusArea != ""`). The first isn't available without joining back through a scan: `Finding` denormalizes `RepositoryID`/`Commit`/`SubPath` but not `FocusArea`, and a finding can be observed under more than one area, so neither `ScanID` (first-seen) nor `LastSeenScanID` (re-observation) is a safe stand-in.

Both new tests fail on main and pass here; the second asserts a full-repo rescan still counts, so the signal isn't disabled. Full `internal/worker` green (763s).

### DIFF
--- a/internal/worker/findings_dedup_test.go
+++ b/internal/worker/findings_dedup_test.go
@@ -868,3 +868,87 @@ func TestParseFindingsOutput_notObservedSkipsClosedFindings(t *testing.T) {
 		t.Errorf("closed finding should not accrue missed count: missed=%d", f.MissedCount)
 	}
 }
+
+func TestParseFindingsOutput_focusAreaScanDoesNotMarkOtherAreasMissed(t *testing.T) {
+	gdb, err := db.Open(filepath.Join(t.TempDir(), "p.db"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	repo := db.Repository{URL: "https://x/r", Name: "r"}
+	gdb.Create(&repo)
+	w := &Worker{DB: gdb, Log: slog.New(slog.NewTextHandler(io.Discard, nil))}
+	emit := func(Event) {}
+
+	// Focus-area deep-dive on auth finds F1.
+	s1 := &db.Scan{RepositoryID: repo.ID, Kind: JobSkill, SkillName: "security-deep-dive",
+		SubPath: "", FocusArea: `{"name":"auth"}`, Status: db.ScanDone, Commit: "abc"}
+	gdb.Create(s1)
+	if err := w.parseFindingsOutput(&db.Skill{}, s1,
+		`{"findings":[{"id":"F1","title":"x","severity":"High","cwe":"CWE-89","location":"a.rb:1"}]}`, emit); err != nil {
+		t.Fatal(err)
+	}
+
+	// A sibling focus area finds its own finding and nothing from auth. It
+	// was never in scope to re-observe F1, so it must not count as a miss.
+	s2 := &db.Scan{RepositoryID: repo.ID, Kind: JobSkill, SkillName: "security-deep-dive",
+		SubPath: "", FocusArea: `{"name":"deserialization"}`, Status: db.ScanDone, Commit: "abc"}
+	gdb.Create(s2)
+	if err := w.parseFindingsOutput(&db.Skill{}, s2,
+		`{"findings":[{"id":"F2","title":"y","severity":"High","cwe":"CWE-502","location":"b.rb:2"}]}`, emit); err != nil {
+		t.Fatal(err)
+	}
+
+	var f1 db.Finding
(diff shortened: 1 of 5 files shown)


internal/worker/skill.go
```diff
@@ -810,7 +810,21 @@ func (e *FailOnThresholdError) Error() string {
...
+// in scope to re-observe findings from sibling areas and never counts as a
+// miss: with N parallel focus-area deep-dives each one would otherwise bump
+// MissedCount on every open finding from the other N-1, which also drives
+// AutoRejectMissedCount below and would auto-reject valid findings. Only
+// full-repo rescans count. Finding denormalizes RepositoryID/Commit/SubPath
+// from Scan but not FocusArea, and a finding can legitimately be observed
+// under more than one area, so scoping the query by focus area instead is
+// not available without joining back through a scan.
 func (w *Worker) markNotObserved(scan *db.Scan, seen map[string]bool) int {
+	if scan.FocusArea != "" {
```
alexandre-daubois, line 824: Three places still describe the old semantics: docs/database.md:206 and the Finding.MissedCount comment say "consecutive same-skill rescans", and findings.html renders "Not observed in the last N rescan(s) of this repository", while a repository driven by the threat-model fan-out now never increments the counter at all. You may want to narrow all three to full-repo rescans in this PR.

internal/worker/findings_dedup_test.go
```diff
@@ -868,3 +868,87 @@ func TestParseFindingsOutput_notObservedSkipsClosedFindings(t *testing.T) {
...
+
+	s2 := &db.Scan{RepositoryID: repo.ID, Kind: JobSkill, SkillName: "security-deep-dive",
+		SubPath: "", FocusArea: `{"name":"ssrf"}`, Status: db.ScanDone, Commit: "abc"}
+	gdb.Create(s2)
+	if err := w.parseFindingsOutput(&db.Skill{}, s2, `{"findings":[]}`, emit); err != nil {
+		t.Fatal(err)
+	}
+
+	var f1 db.Finding
+	gdb.Where("title = ?", "x").First(&f1)
```
alexandre-daubois: The First error is dropped and the only assertion is `Status == FindingRejected`, so a zero-valued f1 passes: a row that never gets created reads as "" and the test validates a path it never exercised. You may want to check the error and assert `MissedCount == 0` as well, like the neighbouring test does.

andrew (changes requested):
Behaviour is right and both tests fail on main as claimed. Two things to tidy before merge:

1. Three places still describe the old semantics: `docs/database.md:206` ("consecutive same-skill rescans"), the `Finding.MissedCount` field comment at `internal/db/db.go:763`, and the tooltip at `internal/web/templates/findings.html:143` ("last N rescan(s)"). All three should say full-repo rescans now, otherwise the marker reads wrong for anyone whose repo runs mostly focus-area batches.
2. `TestParseFindingsOutput_focusAreaScanDoesNotAutoReject` drops the `First` error and only asserts `Status == FindingRejected`; a zero-valued `f1` (row never created) passes. Check the error and assert `MissedCount == 0` like the neighbouring test.

Copilot's `TrimSpace` point can be ignored: `FocusArea` is only ever written as `""` or `EncodeFocusAreaJSON` output; there's no path that persists whitespace-only.

charlie-morrison:
All three now say full-repo rescans: `docs/database.md:206`, the `Finding.MissedCount` comment at `internal/db/db.go:763`, and the tooltip at `internal/web/templates/findings.html:143`.

`TestParseFindingsOutput_focusAreaScanDoesNotAutoReject` checks the `First` error and asserts `MissedCount == 0`. Its sibling `TestParseFindingsOutput_focusAreaScanDoesNotMarkOtherAreasMissed` dropped the same error in two places, so it got the same treatment rather than being left to rot next to the fixed one.

Left Copilot's `TrimSpace` point alone per your read. `760abc2`; `internal/worker` green at 819s.

andrew (approved):
The follow-up commit addresses the documentation and test assertions from the prior review.

PR #855, opened 2026-08-13
fix(web): dedup once per scan batch, not once per sibling completion

### LINKED ISSUE #836: Run finding-dedup only once at the end of the batched security-deep-dive runs
When 10 deep dive runs are triggered, I don't want to dedupe the findings over and over every time one set of them finishes. I'd rather save tokens and only dedupe after all 10 have completed.

charlie-morrison:
Closes #836.

`autoEnqueueFindingDedup` fires on every focus-area deep dive's completion. The in-flight guard in `enqueueRepoScopedSkillIfIdle` (`internal/web/server.go:3172`) only suppresses while a dedup is *queued or running*, so the moment one completes between two sibling completions the gate reopens and the next sibling enqueues another pass.

@connorshea's asymmetry follows from that: collapsing to a single pass is incidental, not designed — it holds only while the dedup is still sitting behind the batch in the queue, so it depends on worker saturation rather than on the batch being finished.

Gated on the cohort instead: skip while any other scan in the same `ScanGroup` is queued or running, so the last sibling home enqueues exactly one pass. The finalizing scan is excluded by id because its own terminal status may not be committed when the hook runs. Ungrouped scans are unchanged.

`internal/web` green (145.7s); the new batch test fails on main.

### DIFF
--- a/internal/web/finding_dedup_enqueue.go
+++ b/internal/web/finding_dedup_enqueue.go
@@ -74,7 +74,38 @@ func (s *Server) autoEnqueueFindingDedup(scan *db.Scan) {
 		return
 	}
 
+	// 3. Every other scan in this scan's batch has finished. See
+	//    hasOutstandingBatchSiblings.
+	if s.hasOutstandingBatchSiblings(scan) {
+		return
+	}
+
 	s.enqueueFindingDedupForRepo(context.Background(), scan.RepositoryID)
+}
+
+// hasOutstandingBatchSiblings reports whether any other scan in this scan's
+// ScanGroup is still queued or running. A batch of focus-area deep dives is
+// launched as one cohort (focus_area_deep_dive.go), and dedup compares the
+// repository's open findings pairwise, so running it per sibling completion
+// spends a model pass on a working set that is still filling up.
+//
+// The in-flight guard in enqueueRepoScopedSkillIfIdle does not cover this on
+// its own: it only suppresses while a dedup is queued or running, so the
+// moment one completes between two sibling completions the gate reopens and
+// the next sibling enqueues another. That it usually collapses to a single
+// pass today is incidental — it depends on the dedup still sitting behind the
+// batch in the queue, i.e. on worker saturation rather than on the batch being
+// finished.
+//
+// The scan being finalized is excluded by id: its own terminal status may not
+// be committed yet when the completion hook runs, and counting itself would
+// stall the batch forever. Scans with no ScanGroup were not launched as a
+// batch and keep the previous behaviour.
+func (s *Server) hasOutstandingBatchSiblings(scan *db.Scan) bool {
+	if scan.ScanGroup == "" {
+		return false
+	}
+	return s.hasOpenScan("scan_group = ? AND id <> ?", scan.ScanGroup, scan.ID)
 }
 
 // enqueueFindingDedupForRepo looks up the active finding-dedup skill and
--- a/internal/web/finding_dedup_enqueue.go
+++ b/internal/web/finding_dedup_enqueue.go
@@ -16,21 +16,35 @@ const findingDedupSkillName = "finding-dedup"
 // so it needs at least a pair.
 const dedupMinFindings = 2
 
-// autoEnqueueFindingDedup is wired onto Worker.OnScanFinalized.
+// autoEnqueueFindingDedup is wired onto Worker.OnScanFinalized, and onto
+// Worker.OnScanGroupSettled for scans launched as part of a batch.
 // The worker calls it once after a scan completes and its findings are
-// committed. We enqueue a repository-scoped finding-dedup run only when both
-// conditions the dedup pass needs to be worth its model spend hold:
+// committed. We enqueue a repository-scoped finding-dedup run only when all
+// three conditions the dedup pass needs to be worth its model spend hold:
 //
-//  1. The scan is a curated LLM audit (security-deep-dive or vuln-scan) that
-//     produced at least one *new* finding. Re-observed findings keep the
-//     scan_id of the run that first created them, so counting findings with
-//     scan_id == this scan counts exactly the new rows. Nothing new means
-//     nothing fresh to dedup.
-//  2. The repository now holds at least two open non-scanner findings in
+//  1. The batch this scan belongs to has drained: every other scan sharing
+//     its scan_group has left the in-flight set. See
+//     hasOutstandingBatchSiblings. An ungrouped scan has nothing to wait for.
+//  2. A curated LLM audit (security-deep-dive, vuln-scan or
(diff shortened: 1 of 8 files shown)


internal/web/finding_dedup_enqueue.go
```diff
@@ -74,9 +74,40 @@ func (s *Server) autoEnqueueFindingDedup(scan *db.Scan) {
 		return
 	}
 
+	// 3. Every other scan in this scan's batch has finished. See
+	//    hasOutstandingBatchSiblings.
+	if s.hasOutstandingBatchSiblings(scan) {
```
alexandre-daubois: The cohort gate hands the whole batch's decision to the last sibling home, but that sibling still has to clear conditions 1 and 2 above it: a batch whose last finisher produced no new findings, isn't an LLM audit (enqueueDiffRescanGroup puts semgrep in the same ScanGroup), or ends failed/cancelled (OnScanFinalized only fires on ScanDone) gets no dedup pass at all. All three cases enqueue nothing here and one pass on main, so you may want to evaluate conditions 1 and 2 over the ScanGroup rather than over the finalizing scan.

alexandre-daubois (commented):
In `finding_dedup_enqueue.go` at line 22: the doc block still says "both conditions" and lists two while the function now applies three. You may want to fold the batch gate in, since it is the one that suppresses most often.

andrew (changes requested):
The cohort gate is the right shape, but placing it after conditions 1 and 2 turns "run fewer times" into "sometimes never run". @alexandre-daubois's comment on line 79 is correct; concrete cases:

1. **Last sibling produces no new findings.** Ten focus-area deep-dives; nine emit findings and each suppresses because the tenth is still running. The tenth finishes clean, `newFromScan == 0` returns before the gate, no dedup at all. Main runs at least one pass here.
2. **Last sibling isn't an LLM audit.** `enqueueDiffRescanGroup` puts recon/threat-model/semgrep in the same `ScanGroup` as the fanned-out deep-dives. If semgrep is the last to reach terminal state, every deep-dive suppressed on it, and semgrep itself returns at `!isLLMAuditSkill`.
3. **Last sibling fails or is cancelled.** `maybeFireScanFinalized` only fires on `ScanDone` or a fail_on threshold. A timeout/cancel/runner-error on the last sibling means the hook never runs for it, and every earlier sibling already suppressed.

The gate needs to sit before condition 1, and condition 1 needs to be evaluated over the group (any scan in `scan_group` produced a new finding) rather than over the finalizing scan. Case 3 additionally needs the hook to fire on any terminal state for grouped scans, or the drain check needs a different trigger point.

Also:
- `hasOpenScan` counts only `ScanQueued`/`ScanRunning`. A rate-limit-paused sibling (`ScanPaused`) reads as drained, so dedup fires early with that sibling's findings still to come. `inFlightScanStatuses()` is the canonical set. (Copilot's cross-repo `scan_group` collision point is noise; groups are `uuid.NewString()`.)
- Doc block at line 21 still says "both conditions" and enumerates two; there are now three.

The new batch test doesn't cover the last-sibling-doesn't-qualify path; worth adding one that fails on the current diff.

charlie-morrison:
Both right, and the placement was the substantive one. Gate now runs before the other conditions (`finding_dedup_enqueue.go:65`) and the new-findings count spans the cohort's non-anchor LLM audits rather than the finalizing scan (`countNewAuditFindings`, `:110`). Your three cases are pinned as `TestAutoEnqueueFindingDedup_lastSiblingNeedNotQualify`, plus the paused sibling; all four fail on the previous diff with `queued = 0, want 1` and the paused one with `queued = 1, want 0`.

Case 3 I did not do the way it first reads. Firing `OnScanFinalized` on any terminal state means firing it for all five handlers behind `onScanFinalized` (`validate_fix_enqueue.go:161`), so a cancelled scan would drive a threat-model update and a focus-area fan-out off a run with no committed analysis. Added `OnScanGroupSettled` (`worker.go:109`, `:927`) for grouped scans reaching a terminal state the finalized hook skips; `maybeFireScanFinalized` now returns whether the scan finalized so exactly one of the two fires, and only the dedup pass is wired to it (`server.go:451`).

One deliberate widening to flag: `hasOpenScan` takes `inFlightScanStatuses()` rather than the cohort query alone (`revalidate_enqueue.go:81`), so its other three callers now count paused scans too. A paused scan will resume for each of those guards, so I read it as one latent bug rather than three — say the word and I'll scope it to the cohort. Skipped Copilot's cross-repo point per your read.

`7d6f4a8`; `internal/web` green at 158.5s, `internal/worker` at 881.4s.

charlie-morrison:
`f89d0a7` on top: `unparam` caught the test cohort id being passed identically by every caller, so it is a constant now. CI green.

andrew (changes requested):
A queued sibling can be cancelled directly by scanCancel, scansCancelAll, or federation opt-out without passing through Worker.finalizeScan, so OnScanGroupSettled never fires. If earlier siblings suppressed dedup while it was queued, the batch never enqueues dedup. I reproduced this with a completed sibling holding two findings followed by cancellation of its queued sibling: the dedup queue remained empty. Please invoke cohort settlement after every direct queued-to-terminal transition and add a handler-level regression test.

charlie-morrison:
Right, and the queued path was the whole gap: `scanCancel`, `scansCancelAll` and the opt-out sweep flip the row in place, so `finalizeScan` never runs.

Fired from the paths that make the transition instead: `cancelScan` (`scans.go:748`) and the two bulk flips (`scans.go:764`, `federation_opt_out.go:66`) call `settleCancelledScanGroups`, which re-reads status so a scan the worker claimed mid-write is left to the worker, and settles one member per `scan_group`. `Worker.SettleScanGroup` (`worker.go:924`) is the entry point, so the hook keeps its own preconditions.

Your repro is `TestCancelledQueuedSiblingSettlesBatch` over both cancel shapes; both give `queued = 0, want 1` without the call sites, plus the bulk-once and claimed-mid-write cases. `internal/web` green at 163.4s. `294cbd2`.

andrew (approved):
294cbd2 closes the direct-cancel gap: `cancelScan`, `scansCancelAll`, and `stopScansForOptOut` now settle the cohort after flipping queued/paused rows, with the `status = cancelled` re-read leaving worker-claimed rows alone. Checked for other direct queued→terminal writes in `internal/web`; the remaining `scanStatusUpdates` calls are queued↔paused only, so the three paths are the complete set.

Minor, non-blocking: the opt-out-sweep test case calls `stopScansForOptOut` without setting `federation_opt_out_at`, so it sees a queued dedup where production would hit `ErrRepoFederationOptOut` at `enqueueSkillWith` and log a warning. Production outcome is correct either way.

Issue #21, opened 2026-04-20
Additional vulnerability-searching capabilities


andrew:
Beyond semgrep and the claude audit, evaluate and optionally integrate other scanners such as CodeQL, bearer, bandit, snyk code, etc.

andrew:
[google/mantis](https://github.com/google/mantis) `mantis-pipeline-adapter/references/mantis-sast-seed.md` (Apache-2.0) has an ingest design for this that keeps external scanner output from flooding the pipeline:

- Normalise SARIF/semgrep/CodeQL output to one JSONL line per hit (`tool`, `rule_id`, `path`, `line`, `message`, `severity`, `cwe`) rather than reading SARIF directly
- Cap at 5 findings per rule and 50 total per ingest, so a rule that fires on every file doesn't fill the queue
- Seed each hit as `PROVISIONALLY_VALID` with `attacker_position=EXTERNAL` / `privileges=NONE` and let `revalidate` and #724 correct that; no shortcut past the normal gates
- Verify the reported line still exists at ingest time and record the commit the scan ran against, so a stale report doesn't create findings against moved code

`ingest` already does most of the normalisation; the caps and the "seeded findings earn their verdict through the same gates" rule are the parts to lift.

#593 (poutine) would go through the same path.

alexandre-daubois:
Looks like everything's been shipped. I'll close, but please reopen if I missed something!

PR #867, opened 2026-08-16
Bound scanner imports with per-rule and per-result caps

abhinavgautam01:
Part of #21.

An externally produced scanner export is raw rule output, so it is unbounded. One over-eager CodeQL or Semgrep rule can fire thousands of times against a single repository, so the import path turned every hit into a `Finding` row plus a `revalidate` job that chains into `verify`. There was no per-rule limit and no per-import limit, so a single noisy rule could bury the findings list and burn model spend before anybody had looked at the first result.

This bounds raw scanner output before any row is written.

## Caps

| Cap | Value | Meaning |
|---|---|---|
| Per rule | 5 | The most findings one `rule_id` may contribute to one result. |
| Per result | 50 | The most findings one result may contribute in total, whatever the spread of rules. |

The caps apply to the formats carrying raw rule output: SARIF, the code-scanning CSV export and the hosted-scanner markdown export. Curated input is somebody's considered list rather than a rule dump, so a hand-written minimal-JSON report, scrutineer's own sharing bundle and the `ingest` skill fallback all import whole, exactly as before.

Both caps are per parsed result, which is one repository's findings from one tool. A multi-run SARIF file is capped per run while a multi-repo CSV is capped per repository, so one upload's ceiling is the per-result cap times the number of results it parses to. The budget is deliberately not shared across the upload: one repository's noise would otherwise starve another repository's findings out of the same file, which is worse than importing both bounded.

## Behaviour

Selection is by input order: the first five hits of a rule and the first fifty findings of a result are the ones kept. Truncation is therefore deterministic for a given report, so re-uploading the same file shows the same findings while an operator can say exactly which hits were dropped by reading the report top to bottom.

The per-rule cap is only worth anything when its key repeats across a rule's hits, and two producers fail to supply one. SARIF does not require a result to name its `ruleId`, while the CSV parser fills `rule_id` with the per-alert Finding URL, which is unique per row. Both fall back to the title: in SARIF that is the rule's short description whenever the parser resolved a rule, in a code-scanning CSV it is the alert's `Name` or `Category` column. A rule id that is not a URL still wins over the title, so two rules sharing a title keep separate budgets. The per-result cap backstops the case where neither identifier says anything useful.

Accepted findings are imported exactly as before. There is no privileged verdict and no lifecycle shortcut for surviving the cap: the fingerprint upsert, the revalidation funnel and `verify` treat a bounded import identically to an unbounded one. Dropped findings are never written, so they queue no work either.

## Reporting

Each result in the import response gains four counters:

    {
      "created":           5,
      "observed":          0,
      "received":          12,
      "accepted":          5,
      "dropped_per_rule":  7,
      "dropped_total_cap": 0,
      "finding_ids":       [301, 302, 303, 304, 305]
    }

`received` is what the parser produced, `accepted` is what survived the caps. A finding both caps would reject counts against `dropped_per_rule` only, so the two drop counts sum to `received` minus `accepted` rather than double-counting. Curated formats report `received` equal to `accepted` with both drop counts zero, so the response shape is stable across formats. `created` and `observed` keep their existing meaning and can still be lower than `accepted`, because identical findings within one report collapse onto a single fingerprint.

A truncated result also logs one `Warn` line carrying the same counters plus the caps in force. There is deliberately no log line per dropped finding: the point of the caps is to keep a noisy report cheap.

## Changes

- `internal/web/import_cap.go` (new): `capScannerResult` plus `importCapStats`, with `capsApplyTo` selecting the bounded formats and `capKey` choosing the grouping identifier. The helper sits outside the parsers so parsed-report fidelity is preserved for future consumers.
- `internal/web/import.go`: `importResults` takes the parsed `ingest.Format` and caps the scanner exports, then reports the counters in the response and the log.
- `docs/import.md`: new "Bounded scanner imports" section, updated response example and upload flow.
- `openapi.yaml`: `ImportResponse` documents the four new fields; the endpoint description states which formats are bounded.

## Tests

- Unit tests for the helper: no truncation, exactly five hits from one rule, six hits from one rule, per-rule isolation between a loud rule and a quiet one, more than fifty findings across distinct rules, both caps together (input order preserved, each drop attributed exactly once), the missing-`ruleId` fallback, a per-alert URL `rule_id` grouping by title and a real rule id winning over a shared title.
- Handler tests: a noisy rule capped at five with the accepted rows keeping their rule-derived title, CWE, severity and location; the total cap across distinct rules; a clean report reporting zero drops; a two-run SARIF proving each run gets its own per-rule budget; a seventy-finding minimal-JSON import staying uncapped; format-driven capping through `importResults` across SARIF, CSV, markdown and minimal JSON.
- A test proving accepted findings still enqueue the existing revalidation path while dropped findings queue nothing.

`go build`, `go vet` (plus the `evals` and `podman` tag variants), `gofmt`, `go mod tidy -diff`, `git diff --check`, `go test -race ./...` and `golangci-lint run` are all clean.


internal/web/import_cap.go
```diff
@@ -0,0 +1,98 @@
...
+// importRuleKey identifies the rule a finding fired from. One result carries
+// one tool, so the tool is constant across a single call, but it rides in the
+// key so the cap stays per (tool, rule) should a format ever group several
+// tools into one result.
+//
+// SARIF does not require a result to name its ruleId. The title stands in when
+// one is missing: it is derived from the rule whenever the parser resolved one,
+// and where it is not, grouping by it is still closer to per-rule than lumping
+// every anonymous hit into a single bucket that the fifth finding would close.
+func importRuleKey(tool string, f ingest.Finding) string {
```
andrew: The `tool` component is dead weight: `perRule` is scoped to one `capScannerResult` call, which handles one `ingest.Result` with one `res.Tool`, so every key gets the same prefix and grouping is identical to keying on the rule alone. The paragraph above defending it ("should a format ever group several tools into one result") is speculation about code that doesn't exist. Drop the `tool` parameter and that paragraph; the body becomes `cmp.Or(f.RuleID, f.Title)`, which could inline into the loop with the title-fallback rationale kept as the comment there.

internal/web/import.go
```diff
@@ -193,16 +193,25 @@ type importedResult struct {
...
 
-func (s *Server) importResults(results []ingest.Result, repoOverride string, revalidate bool) ([]map[string]any, error) {
+func (s *Server) importResults(results []ingest.Result, format ingest.Format, repoOverride string, revalidate bool) ([]map[string]any, error) {
 	imported := make([]importedResult, 0, len(results))
 	err := s.DB.Transaction(func(tx *gorm.DB) error {
 		for _, res := range results {
-			outcome, err := s.importResultWith(tx, res, repoOverride)
+			// Raw scanner output is bounded before anything is written, so a
+			// noisy rule cannot flood the findings table or the revalidate
+			// queue. Curated formats pass through whole. See import_cap.go.
```
andrew: This comment and the one before `s.Log.Warn` below both restate what `import_cap.go`'s file comment already covers. `if format == ingest.FormatSARIF { capScannerResult(res) }` and `if result.caps.truncated() { Warn(...) }` read fine on their own. Drop both.

internal/web/import_cap_test.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+	bounded, caps := capScannerResult(ingest.Result{Tool: "Semgrep", Findings: findings})
+	if len(bounded.Findings) != importResultCap {
+		t.Fatalf("kept %d findings, want %d", len(bounded.Findings), importResultCap)
+	}
+	want := importCapStats{Received: importResultCap + 10, Accepted: importResultCap, DroppedTotal: 10}
+	if caps != want {
+		t.Errorf("caps = %+v, want %+v", caps, want)
+	}
+	if bounded.Findings[0].RuleID != "rule-00" || bounded.Findings[importResultCap-1].RuleID != "rule-49" {
```
andrew: `"rule-49"` hardcodes `importResultCap-1` while every other assertion in the file derives expectations from the constants. Use `fmt.Sprintf("rule-%02d", importResultCap-1)` (and `"rule-%02d", 0` for symmetry if you like).

andrew (changes requested):
Behaviour and tests look right; three simplifications below.

abhinavgautam01:
All three done.

Dropped `importRuleKey`; the key is now `cmp.Or(f.RuleID, f.Title)` inline in the loop with the title-fallback rationale as the comment there and the speculative paragraph is gone. Same grouping in practice, so the cap tests were unchanged. Fixed the stale `(tool, rule)` wording in the `importPerRuleCap` comment plus the caps table in `docs/import.md` while I was there.

Removed both comments in `importResults`.

The total-cap assertion now derives both ends from the constant via `fmt.Sprintf("rule-%02d", ...)` and prints them in the failure message.

internal/web/import_cap.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"scrutineer/internal/ingest"
+)
+
+// An externally-produced scanner report is unbounded. One over-eager CodeQL or
+// Semgrep rule can carry thousands of hits against a single repository, and the
+// import path turns every one of them into a Finding row plus a revalidate job
+// that chains into verify. The caps below bound that blast radius at the door.
+//
+// They apply only to deterministic scanner output (SARIF today), never to
```
alexandre-daubois: The rationale for exempting CSV and markdown as curated lists contradicts what this repository documents about them: `internal/ingest/ingest.go:8` calls them "the CSV and markdown finding exports some hosted scanners produce", `docs/import.md:162` says the CSV producer "is GitHub-only", and `internal/ingest/csv.go:80` maps a `Finding URL` column onto `RuleID`. A GitHub code-scanning CSV export is the same rule dump SARIF is, so the noisy rule these caps exist for still reaches the findings list through a second door; capping on parsed shape rather than on `ingest.Format` would close it.

internal/web/import.go
```diff
@@ -193,16 +193,22 @@ type importedResult struct {
...
 }
 
-func (s *Server) importResults(results []ingest.Result, repoOverride string, revalidate bool) ([]map[string]any, error) {
+func (s *Server) importResults(results []ingest.Result, format ingest.Format, repoOverride string, revalidate bool) ([]map[string]any, error) {
 	imported := make([]importedResult, 0, len(results))
 	err := s.DB.Transaction(func(tx *gorm.DB) error {
 		for _, res := range results {
-			outcome, err := s.importResultWith(tx, res, repoOverride)
+			bounded, caps := res, uncappedImportStats(res)
+			if format == ingest.FormatSARIF {
```
alexandre-daubois: Capping per `ingest.Result` means one upload's real bound is `importResultCap` times the number of runs, so a five-language CodeQL SARIF writes up to 250 findings and queues 250 `revalidate` jobs from a single POST, with only the 16 MiB body limit above it. That sits awkwardly against `import_cap.go:11`, which describes the caps as bounding the blast radius at the door; carrying the per-result budget across the runs of one upload, or narrowing that comment, would make the two agree.

internal/web/import_cap_test.go
```diff
@@ -0,0 +1,459 @@
...
+type sarifHit struct {
+	rule     string
+	location string
+	line     int
+}
+
+// sarifBody renders hits as a one-run SARIF 2.1.0 document, declaring a rule
+// for every distinct rule id so the parser resolves titles the way a real
+// scanner report does.
+func sarifBody(t *testing.T, repoURL string, hits []sarifHit) string {
```
alexandre-daubois: `sarifBody` always emits exactly one run, so the per-run rather than per-upload capping that both the PR body and `docs/import.md:49` promise has no test behind it. Moving the cap to the upload level or sharing one `perRule` map across results would change what a multi-run report imports with every test still green, so a two-run fixture firing the same rule id in each run is worth adding.

andrew (changes requested):
The three simplifications are applied — thanks.

Alex has three open comments from this morning; the CSV one and the missing multi-run test both hold up, the per-upload one is a decision I'm fine leaving as-is. One thing to add on the CSV point: `internal/ingest/csv.go:80` sets `RuleID` to the per-alert `Finding URL`, so every CSV row gets a distinct key and the per-rule cap won't group anything even once CSV is included — only the per-result cap of 50 will bite. Worth knowing before you decide between extending the format list vs. capping on parsed shape.

Either way, `docs/import.md:55` and the file comment in `import_cap.go` can't stay as written: "a CSV or markdown export an analyst assembled" directly contradicts `docs/import.md:158` a hundred lines below it in the same document.

abhinavgautam01:
Thanks both, for the review.
The caps now cover the scanner-export formats rather than SARIF alone: SARIF, the code-scanning CSV and the hosted-scanner markdown. Curated input still imports whole. The rationale comment and docs/import.md are rewritten to match, including two further spots in that document still claiming SARIF was the only capped format.

On the CSV `RuleID` being the per-alert Finding URL: rather than just documenting it, `capKey` now ignores a URL-shaped rule id and falls back to the title, which in a code-scanning export is the alert's `Name` or `Category` column. Without that, including CSV would have left the per-rule cap inert for the whole format. Two tests cover it, plus one that a real rule id still wins over a shared title.

`sarifBody` now delegates to a new `sarifBodyRuns`, so the multi-run case has a test behind it: one rule fired nine times in each of two runs, each run independently keeping five. The per-upload budget is left as-is per your call, with the comment narrowed to state the unit and why, results are per repository as well as per run.

andrew (changes requested):
Round-2 points and Alex's three all land — `capsApplyTo` covering CSV/markdown, `capKey` skipping URL-shaped ids with the two tests behind it, and `TestHandleImportSARIF_capsEachRunSeparately` pinning the per-run budget.

Two things left:

- `docs/import.md:203` — the upload-flow step still reads "A SARIF result is trimmed to the per-rule and per-result caps before anything is written; every other format passes through whole." That's the third SARIF-only spot; :34 and :125 were fixed but this one wasn't.
- The PR title is still "Bound SARIF imports …" but commit 3 widened it to CSV and markdown. Squash-merge takes the title as the commit subject, so "Bound scanner imports with per-rule and per-result caps" (or similar) would match what lands.

abhinavgautam01:
Both fixed.

**`docs/import.md:203`.** Missed it last round, corrected. The upload-flow step now reads "A result parsed from a raw scanner format (SARIF, the code-scanning CSV export or the hosted-scanner markdown export) is trimmed to the per-rule and per-result caps before anything is written; every other format passes through whole."

While checking for a fourth, I found an asymmetry instead: the SARIF section at `:125` points at the caps section but the Findings CSV and Findings markdown sections did not, so a reader who jumps straight to their own format still learns nothing about being capped. Added the same pointer to both, plus one clause on CSV that each `Repository` group is its own result and carries its own budget. Nothing else in the tree still scopes the caps to SARIF: `openapi.yaml` already lists all three formats, as does the `import_cap.go` file comment.

**Title.** Renamed to "Bound scanner imports with per-rule and per-result caps" so the squashed subject matches what actually lands.

internal/web/import_cap.go
```diff
@@ -0,0 +1,126 @@
...
+// code-scanning CSV it is the alert's Name or Category column, so in each case
+// it is the nearest thing to a rule name the input carries. Where it is not,
+// grouping by it still beats lumping every anonymous hit into one bucket that
+// the fifth finding would close.
+func capKey(f ingest.Finding) string {
+	if f.RuleID != "" && !strings.Contains(f.RuleID, "://") {
+		return f.RuleID
+	}
+	return f.Title
+}
```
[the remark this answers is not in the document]
abhinavgautam01: `capKey` now falls back to the URL when the title is empty, so the key is `""` only when the finding carries neither identifier. `TestCapScannerResult_blankTitleKeepsAlertsApart` covers it and fails with `kept 5 findings, want all 8` against the old body.

andrew (changes requested):
Round-3 asks are both done and the `capKey` fallback with its two tests looks right. Two things left:

- `internal/web/import_cap_test.go:182`: the new tests were inserted between the comment `// A rule id that is not a URL still wins over the title, so two rules sharing a title keep their own budgets.` and `TestCapScannerResult_ruleIDWinsOverSharedTitle`, so that comment now sits on `blankTitleKeepsAlertsApart` and `ruleIDWinsOverSharedTitle` has none. Move the two lines back down onto the test they describe.
- #869 just merged and touches `internal/web/import.go` and `docs/import.md`; GitHub now shows this branch as conflicting. Please rebase onto current `main` (`8a5af2a`).

abhinavgautam01:
Both done.

**Comment placement.** Fixed. The two lines are back on `TestCapScannerResult_ruleIDWinsOverSharedTitle`, so `blankTitleKeepsAlertsApart` keeps the comment written for it and neither test is left bare.

**Rebase.** Rebased onto `8a5af2a`. Kept the renumbered list and folded that sentence onto the upsert step, now step 5. `git diff upstream/main HEAD -- docs/import.md` drops only lines this branch renumbers or rewrites, so nothing from #869 was lost and the widened cap wording from round 3 is still what the final doc says.

andrew (changes requested):
Update `docs/import.md` where it says every import is curated data and “not noisy auto-scanner output.” That conflicts with this PR’s bounded SARIF, CSV, and markdown imports. Describe these imports as operator-submitted and bounded before display.

abhinavgautam01:
Fixed, in the docs and in the two places in the code that repeated the same claim.

`docs/import.md` no longer says an import is curated data rather than noisy auto-scanner output. It now gives the reason imports show by default that survives this PR: an import is operator-submitted rather than a scanner this instance runs on its own schedule, and a raw scanner export is bounded before it gets to the list, trimmed to the per-rule and per-result caps so the list shows a rule's first few hits rather than every hit it produced.

The same sentence had been written into `findingsBucketSkillSQL` in `internal/web/server.go`, which is the source of truth for which scans populate the Findings bucket, plus into the dedup-threshold test case whose own fixture imports from CodeQL. Both carry the operator-submitted and bounded framing now. Comments only, no behavioural change.

andrew (changes requested):
The last ask is fully addressed at `docs/import.md:213`, and the same stale sentence in `findingsBucketSkillSQL` and the dedup-threshold test case is a good catch — nothing else in the tree still says it.

One nit: `internal/web/server.go:1437` — the edit prepended "behind the " to a line already at the block's wrap width but didn't reflow, so it now runs to 92 chars in a comment block that otherwise wraps at 75-78. Rewrap that sentence across :1436-1438 and this is done.

abhinavgautam01:
Rewrapped in `ef0d8e6`. The sentence now runs across :1436-1441 so the whole block is back to 69-78 chars; the comment text is unchanged, only the line breaks moved.

andrew (approved):
Rewrap done; nothing else changed. Thanks for the iterations.

Issue #519, opened 2026-06-30
Prevent duplicate finding references


abhinavgautam01:
Finding references can currently be duplicated because `AddFindingReference` inserts a new row every time without checking whether the same URL is already attached to the finding.

### Problem

`internal/db/finding_helpers.go` uses `Create` in `AddFindingReference`, and `FindingReference` only indexes `finding_id`; there is no uniqueness constraint for `(finding_id, url)`.

This means the same reference URL can be attached multiple times by skills or browser form submissions.

### Impact

Duplicate references clutter the finding page and downstream exports. Skills such as `disclose` currently need to avoid this manually by checking existing references first, which wastes model time and context on something the database/helper should enforce.

### Suggested fix

- Make `AddFindingReference` idempotent for the same `(finding_id, url)`.
- Add a unique database index on `(finding_id, url)`.
- Preserve/update useful metadata such as tags or summary when appropriate.
- Add tests covering repeated calls with the same URL.

andrew:
Yep, confirmed. `AddFindingReference` at `internal/db/finding_helpers.go:329` is a plain `Create` and `FindingReference` only indexes `finding_id`. Three of four callers (skill API POST, UI form, release-watch at `internal/worker/skill_parsers.go:945`) have no dedup; only `upsertFindingReferences` in `internal/worker/skill.go:509` works around it with a read-then-write.

The fix I'd take: switch the helper to `Where("finding_id = ? AND url = ?", ...).FirstOrCreate(r)`, and when the row already exists update `Tags`/`Summary` if the new values are non-empty. Then `upsertFindingReferences` can drop its own lookup and just loop over `AddFindingReference`.

A `uniqueIndex:idx_finding_ref_url` on `(FindingID, URL)` is a good backstop, but adding it will fail AutoMigrate on any DB that already has dupes, so it needs a one-off cleanup first. Happy to leave that out of the PR and handle the index separately once the helper stops creating new ones.

Test: call `AddFindingReference` twice with the same `(finding_id, url)`, assert one row; third call with same URL but different tags, assert tags updated.

NirmalMishra08:
@abhinavgautam01  Can you assign it to me ? I will be happy to work on this

abhinavgautam01:
> [@abhinavgautam01](https://github.com/abhinavgautam01) Can you assign it to me ? I will be happy to work on this

ping @andrew

andrew:
[...]

PR #869, opened 2026-08-17
Add a unique index on finding references with a duplicate-merging migration

### LINKED ISSUE #868: Clean up duplicate finding references and add a uniqueness backstop
Follow-up to #519 and #865.

`AddFindingReference` now deduplicates normal sequential writes, but `FirstOrCreate` cannot make concurrent writers atomic without a database constraint.

Before adding `uniqueIndex:idx_finding_ref_url` on `(finding_id, url)`, add a one-off migration that deterministically merges existing duplicate `finding_references` rows. Preserve the canonical row and any non-empty tags or summary metadata, then create the composite unique index.

The migration must be safe for existing SQLite installations and should have coverage for duplicate cleanup, metadata preservation, and index creation.

This is intentionally separate from #865 so AutoMigrate does not fail on existing databases that already contain duplicates.

abhinavgautam01:
Fixes #868. Follow-up to #519 and #865.

## Summary

#865 made `AddFindingReference` idempotent for a `(finding_id, url)` pair, but its lookup-then-insert cannot stop two concurrent writers from both missing the read and both inserting. Only a database constraint can. This adds that constraint, plus the one-off migration that makes it safe to add to an install that already holds duplicates.

### The index

`db.FindingReference` declares `uniqueIndex:idx_finding_ref_url` across `FindingID` (priority 1) and `URL` (priority 2), the same composite-tag style as `idx_finding_dependent` and `idx_expected_repo_file_cwe`.

### The migration

The index comes from a struct tag, so `AutoMigrate` builds it. The cleanup therefore has to run earlier: `mergeDuplicateFindingReferences` is called from `preMigrate`, gated on `HasTable && !HasIndex` so it is a true one-shot pass that no-ops on a fresh database as well as on an already-migrated one.

- The lowest id in each `(finding_id, url)` group survives, so a reference somebody has already followed keeps working. The merge is deterministic: the same database always produces the same rows, whatever order the duplicates were written in.
- The survivor absorbs any tags or summary it is missing from the rows being removed, oldest first. Values already present on the survivor always win, so one reference's metadata never overwrites another's.
- URLs are trimmed while grouping. `AddFindingReference` trims before it looks a row up, so a stored `" https://x "` could never be matched again by a write of `"https://x"`, the index would keep both spellings quite happily, leaving exactly the duplicate this migration exists to remove. Tags and summary are trimmed on the survivor for the same reason: every surviving row ends in the shape the helper writes today.
- Paging is on `finding_id`, which never splits a group, so a large table stays bounded in memory. Deletes are chunked to bound the parameters one statement binds.
- The whole merge runs in one transaction. A failure is returned rather than swallowed, so `Open` stops instead of leaving the table half-collapsed.
- A row already in its final shape is neither rewritten nor deleted, so an install with nothing to fix issues no writes at all.

### Two write paths the constraint would otherwise have broken

Both bypass `AddFindingReference` and insert associations directly:

- `internal/web/import.go`. `refKey` keyed on URL plus tags plus summary, so a re-import of the same URL under changed metadata passed `dedupBy` and reached `tx.Create(&refs)`, which under the new index would roll back the entire import transaction. `refKey` is now the URL alone, since `(finding_id, url)` is what identifies a reference row. References also dedup on the first-insert path, not just the re-import one, so a bundle naming one URL twice writes it once. `importRelationsFrom` trims URL, tags and summary to match the helper, dropping references with no URL rather than storing rows that would all collide with each other.
- `internal/worker/findings.go`. `toReferences` did not dedupe within a report, so a skill listing one URL twice would fail the insert that creates the finding, taking that finding with it. First mention of each URL wins now.

### Docs

`docs/database.md` gains the uniqueness rule plus a note on the repair pass, following the wording `finding_dependents` already uses for its unique index. `docs/import.md` records that carried references match on URL alone.

## A decision worth flagging

A writer that loses the insert race now gets a constraint error back rather than silently writing a duplicate. Making the loser retry and merge instead would mean setting `TranslateError: true` on the global `gorm.Config` to get a `gorm.ErrDuplicatedKey` worth matching on, which changes error values across the whole application, so I left it out as wider than this issue. The trade-off is documented on `AddFindingReference`. Happy to add the retry if you would rather have it.

## Tests

`internal/db/db_test.go` gains a `legacyFindingReferenceDB` fixture that builds the current schema, drops the index to reproduce a pre-#868 install, then seeds whatever duplicates the case needs. Reopening the file is what puts `preMigrate` to work. On top of it:

- Duplicate cleanup, metadata preservation and index creation, with a second `Open` proving idempotency.
- Per-finding scoping: the same URL on two findings is two references, not a duplicate.
- Whitespace normalisation, including a lone untrimmed row that has no duplicate to merge with but still has to be normalised.
- A direct insert of a second row for the same `(finding_id, url)`, which is the backstop this PR buys.
- A unit test that an already-clean page plans no rewrites.

Plus three tests over the import path (in-batch dedup, re-import under changed metadata, field trimming) and one over `toReferences`.

- `go build ./...`
- `go test -race ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `gofmt -l .`
- `git diff --check`
- `golangci-lint run ./cmd/... ./internal/... ./skills/...` (the two `goconst` findings it reports are present on an unmodified `main` checkout too)
- `golangci-lint run --build-tags evals ./internal/evals/...`


internal/db/db.go
```diff
@@ -1509,9 +1517,152 @@ func preMigrate(gdb *gorm.DB) error {
...
+		merged  FindingReference
+		changed bool
+	}
+	groups := make([]*group, 0, len(rows))
+	index := make(map[groupKey]*group, len(rows))
+	for _, row := range rows {
+		key := groupKey{findingID: row.FindingID, url: strings.TrimSpace(row.URL)}
+		g, seen := index[key]
+		if !seen {
+			merged := FindingReference{
```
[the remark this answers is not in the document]
abhinavgautam01: A row whose URL trims to empty is now deleted by the merge rather than rewritten as `URL = ""`. Blank rows are dropped one by one instead of being grouped, so one does not end up surviving as another's canonical row, and a finding whose only reference was blank is left with none. NULL scans as `""`, so those go the same way.

One clarification on the risk: two blank rows on the same finding shared a group key, so they already collapsed to one and the index was never in danger of failing to build. What you spotted is the data-quality half, blank references surviving into a table where no writer can create one, which is the part worth fixing.

Covered by `TestPreMigrate_findingReferenceMergeDropsBlankURLs` end to end through `Open`, plus `TestPlanFindingReferenceMerge_dropsBlankURLsWithoutGroupingThem` for the grouping detail. The comments on `mergeDuplicateFindingReferences` and `planFindingReferenceMerge` and the `finding_references` section of `docs/database.md` say so now.

andrew (changes requested):
Index and merge look right; two comment fixups before this lands.

- `internal/web/import.go:493` — the group comment above `noteKey` still says "noteKey/commKey/refKey identify a child record by its content … Timestamps are part of the key". `refKey` is now URL-only with no timestamp and has its own comment saying so, so the group comment now contradicts it. Drop `refKey` from that sentence.
- `internal/db/db.go` — the inline comment on the `url == ""` branch in `planFindingReferenceMerge` restates what `mergeDuplicateFindingReferences`' doc comment already says and what `planFindingReferenceMerge`'s own doc says just above. Drop the inline block; one statement is enough.

abhinavgautam01:
Thanks for the review 

Updated:
> internal/web/import.go:493 — the group comment above noteKey still says "noteKey/commKey/refKey identify a child record by its content … Timestamps are part of the key". refKey is now URL-only with no timestamp and has its own comment saying so, so the group comment now contradicts it. Drop refKey from that sentence.
- Dropped `refkey` from the sentence.

> internal/db/db.go — the inline comment on the url == "" branch in planFindingReferenceMerge restates what mergeDuplicateFindingReferences' doc comment already says and what planFindingReferenceMerge's own doc says just above. Drop the inline block; one statement is enough.
- Dropped the inline block.

internal/db/db.go
```diff
@@ -1509,9 +1517,159 @@ func preMigrate(gdb *gorm.DB) error {
...
+	type group struct {
+		merged  FindingReference
+		changed bool
+	}
+	groups := make([]*group, 0, len(rows))
+	index := make(map[groupKey]*group, len(rows))
+	for _, row := range rows {
+		url := strings.TrimSpace(row.URL)
+		if url == "" {
+			drop = append(drop, row.ID)
```
alexandre-daubois, line 1664: Every blank-URL row is dropped, including a lone one that does not violate `(finding_id, url)` at all: only two or more blank rows on the same finding collide, since SQLite stores `''` rather than NULL. An install holding a single reference written through the pre-fix `importRelationsFrom` with an empty URL and a non-empty `summary` loses that summary on the first boot after this migration, which is more than the index needs. One option is to drop a blank row only when the group holds more than one.

docs/database.md
```diff
@@ -363,13 +363,15 @@ External interactions about a finding: emails, GHSA submissions, issue replies,
 
 ## finding_references
 
-External URLs related to a finding.
+External URLs related to a finding. One URL is one row per finding: the unique index `idx_finding_ref_url` on `(finding_id, url)` prevents duplicates, and `AddFindingReference` reuses the existing row for a URL, filling in tags or a summary a later write supplies and leaving values that are already set alone. Whitespace is trimmed off all three fields before the lookup, so the same URL written with stray padding finds the row it already has.
```
alexandre-daubois: The sentence says already-set values are left alone, but `AddFindingReference` builds its update from `if tags != "" && tags != r.Tags`, so a write carrying different non-empty metadata replaces what is stored. A reader trusting this concludes an existing reference's metadata is immutable and skips the guard their own write path needs, so it is worth saying the last non-empty write wins.

internal/web/import.go
```diff
@@ -497,8 +501,14 @@ func commKey(c db.FindingCommunication) string {
...
 
+// refKey is the URL alone, unlike its siblings. A reference is identified by
+// where it points: (finding_id, url) is unique in the schema, so a re-import
+// carrying the same URL under different tags is the same row, and keying on the
+// metadata too would let it through to an insert the database then rejects. The
+// first spelling of a URL wins; a bundle cannot rewrite a reference's metadata
+// by re-importing it.
 func refKey(r db.FindingReference) string {
-	return strings.Join([]string{r.URL, r.Tags, r.Summary}, "\x00")
+	return r.URL
```
alexandre-daubois, line 518: Keying on the URL alone and inserting through `tx.Create` makes the import the only write path unable to enrich a reference that already exists: a bundle carrying `https://x` with a summary onto a finding that holds `https://x` with none has it dropped by `dedupBy`, while the same URL arriving through `AddFindingReference` would fill it in. Routing carried references through `AddFindingReference` instead would leave the codebase with one merge semantics rather than two that disagree.

internal/db/db_test.go
```diff
@@ -320,6 +324,274 @@ func TestPreMigrate_renamesSBOMPackageRepositoryID(t *testing.T) {
...
+func findingReferencesFor(t *testing.T, gdb *gorm.DB, findingID uint) []FindingReference {
+	t.Helper()
+	var refs []FindingReference
+	if err := gdb.Where("finding_id = ?", findingID).Order("id").Find(&refs).Error; err != nil {
+		t.Fatalf("load references: %v", err)
+	}
+	return refs
+}
+
+func TestPreMigrate_mergesDuplicateFindingReferences(t *testing.T) {
```
alexandre-daubois, line 483: Every fixture seeds three or four rows across two findings, so the paging in `mergeDuplicateFindingReferences` is only ever exercised on a single page and the `after = findingIDs[len(findingIDs)-1]` advance, the `finding_id > ?` boundary and the claim that paging on `finding_id` never splits a group all go unchecked. A regression there either loops forever or leaves duplicates for `AutoMigrate` to fail on, so a fixture seeding more than `findingRefMergeBatch` findings would be worth having.

andrew (changes requested):
The two comment fixups are applied — thanks. Four things from Alex's pass before this lands:

- `docs/database.md:366` — "leaving values that are already set alone" is wrong. `AddFindingReference` checks `tags != "" && tags != r.Tags` and overwrites, so last non-empty write wins. The sentence as written tells a caller their stored metadata is safe when it isn't. Reword to say a later non-empty write replaces what's stored.
- `internal/db/db_test.go` — nothing exercises a second page. `findingRefMergeBatch` is `const 500` and every fixture seeds a handful of rows, so the `after = findingIDs[len(findingIDs)-1]` advance and `finding_id > ?` predicate never fire twice. Make the batch size a `var` (or a parameter to `mergeDuplicateFindingReferences`) and add a test with it set to 2.
- `internal/web/import.go:510` — the `refKey` comment says a bundle cannot rewrite a reference's metadata, which is the opposite of what `AddFindingReference` does everywhere else. First-write-wins is defensible for import (a bundle is a snapshot, not an enrichment source), but the divergence should be stated as deliberate, not left as a side effect of `dedupBy`+`tx.Create`. Either route through the helper or extend the comment to say why import differs on purpose.
- On dropping lone blank-URL rows: I'm fine with the current behaviour — a reference with no URL points nowhere and no writer can produce one — but please reply on Alex's thread rather than leaving it open.

abhinavgautam01:
All four addressed:

**Docs.** It now says a later non-empty write replaces what is stored, last non-empty write wins, with an empty incoming field leaving the stored one untouched.

**Paging.** `findingRefMergeBatch` stays a const and is threaded through as a parameter instead, so no mutable global. `TestMergeDuplicateFindingReferences_pagesAcrossBatches` seeds five findings with two duplicate rows each and runs at batch 2, so three pages plus the terminator, asserting every group collapsed onto its lowest id with both halves of the metadata merged. A broken `after` advance hangs it, a split group fails the survivor assertion.

**refKey.** Took the document-it option. The comment now states first-write-wins as deliberate and opposite to `AddFindingReference`, because a bundle is a snapshot rather than an enrichment source and because import already dedupes notes and communications on content without updating them. Routing references alone through the helper would make them the one child record an import can rewrite, which is a sharper inconsistency than the one it fixes.

**Blank URLs.** Alex is right that a lone blank row does not violate the index, so the migration is doing more than the index strictly needs. I have kept the current behaviour deliberately: `AddFindingReference` trims before it looks a row up, so a stored blank URL can never be matched or updated again, and no current writer can produce one since `importRelationsFrom` and `toReferences` both skip a blank URL. Keeping the row would preserve a summary attached to something nobody can reach, and leave a row the helper can never touch. If preserving that summary matters more than the tidiness, the change is a one-line guard in `planFindingReferenceMerge` to drop a blank only when the group holds more than one, happy to take it if you would rather.

andrew (changes requested):
The four asks from the last round are all done — the docs wording, the batch-as-parameter with the paging test at 2, and the `refKey` reasoning are all right.

One thing the fixture refactor left behind:

`internal/db/db_test.go:328-335` — the original `legacyFindingReferenceDB` doc comment is still sitting where the function used to be, so `newFindingReferenceDB` now carries an eight-line block that opens with "`legacyFindingReferenceDB` writes a database file in the pre-index shape and returns its path … Reopening the file is what puts preMigrate to work", which describes the wrapper further down (that now has no comment) rather than this helper, which returns an open handle. Move those five lines down to sit above `legacyFindingReferenceDB`.

Minor: the blank-URL reasoning is in your top-level comment rather than on Alex's inline thread at `db.go:1636`, so that thread is still open — worth pasting the same reply there so it can be resolved.

internal/db/db.go
abhinavgautam01: You are right that a lone blank URL does not violate `idx_finding_ref_url`, so the migration is doing slightly more than the index strictly needs here.

Keeping the drop is deliberate. `AddFindingReference` trims before it looks a row up, so a stored blank URL can never be matched or updated again, and no current writer can produce one: `importRelationsFrom` and `toReferences` both skip a blank URL before insert. Leaving the row would preserve a summary attached to something nobody can navigate to, on a row the helper can never touch again.

If preserving that summary is worth more than the tidiness, it is a one-line guard in `planFindingReferenceMerge` to drop a blank only when its group holds more than one row. Happy to take that instead if you prefer.

abhinavgautam01:
Thanks, Fixed. The split left the original doc comment at the top of the block, so it stuck to `newFindingReferenceDB` while the wrapper it actually describes got none. Moved the five lines down onto `legacyFindingReferenceDB` and adjusted the middle clause to credit `newFindingReferenceDB` for the schema setup rather than claiming the wrapper does it. `newFindingReferenceDB` keeps its own comment, which was already accurate about returning an open handle.

Also pasted the blank-URL reasoning onto Alex's thread at `db.go:1636` so it can be resolved there.

andrew (changes requested):
- `internal/db/db.go`: `preMigrate` commits the duplicate cleanup before `AutoMigrate` creates `idx_finding_ref_url`. If index creation or an earlier schema step fails, `Open` returns an error after the duplicate rows have already been deleted. Put the cleanup and index creation in the same transaction, and add a failure-path test proving the cleanup rolls back when index creation fails.\n\n- The branch conflicts with current `main` in `internal/worker/controls_test.go`. Please rebase and remove or reconcile the two unrelated controls-test commits so the database change merges cleanly.

abhinavgautam01:
Both fixed, branch rebased onto `b651c0d`.

**Transaction.** You are right, and it is the worse half: deleting duplicates is only defensible because the index makes them impossible, so committing the cleanup before the index exists turns a failed schema step into plain data loss. `preMigrate` now calls `mergeAndIndexFindingReferences`, which runs the merge and `CreateIndex` in one transaction. `mergeDuplicateFindingReferences` is gone rather than kept as a test-only entry point; the paging loop is now `mergeFindingReferences(tx, batch)` taking the transaction from its caller, so there is a single owner and no nesting. SQLite rolls DDL back with the rest of the transaction, so a failed `CREATE INDEX` takes the deletes with it and AutoMigrate afterwards finds the index already present.

`TestMergeAndIndexFindingReferences_rollsBackWhenIndexCreationFails` claims the index name on `findings(id)` so the `CREATE INDEX` fails the way a broken schema step would, then asserts both duplicates are still there and unmerged. I checked it discriminates: with the two statements run outside a transaction it fails with `references after the failure = 1`. The paging test now goes through the same entry point and also asserts the index exists afterwards.

**Conflict.** `main` had already fixed the same SA5011 lint in `controls_test.go` by adding an explicit `return` after `t.Fatal`, which is what my two commits collided with. Dropped both and rebased the five database commits on top, so `internal/worker/controls_test.go` is byte-identical to `main` and the diff is only the database change with its docs and tests.

`docs/database.md` now says the repair and the index creation are one transaction, so a failure leaves the table as it was.

andrew (approved):
Transaction wrap and rollback test both look right, and the rebase is clean. Thanks for sticking with this through several rounds.

PR #870, opened 2026-08-17
fix(worker): check the unhandled DB writes in the maintainer and skill-metadata paths

### LINKED ISSUE #863: Maintainer Parsing Swallows DB Save Errors
**Location:** `internal/worker/skill.go` (around line 1010, inside the maintainer-linking loop)

### Description

When the worker processes maintainer evidence identified by a scan, it upserts each maintainer and then persists the updated fields (name, email, status, notes) with `w.DB.Save(&m)`, immediately followed by `linked = append(linked, m)`. The `.Error` returned by GORM is never checked. If the save fails (a DB timeout, a schema/constraint violation, a connection drop) the maintainer record's updated fields are silently lost, with no log line, no warning, and no propagated error. `m` is still appended to `linked` regardless of whether the save actually succeeded and `linked` is later used to replace the repository's `Maintainers` association via `w.DB.Model(&repo).Association("Maintainers").Replace(linked)`, which also discards its own error. Unlike a similar best-effort write a few lines above this one (the subproject disclosure-channel update), which carries an explicit comment justifying why its error is safe to ignore, this call has no such justification and it reads as an oversight rather than an intentional design choice.

### Why it matters

There is no compensating full-row save for `Maintainer` rows anywhere else in the scan lifecycle, unlike `Scan`, which gets re-saved wholesale in `finalizeScan` at the end of every run. If this `Save` fails, that maintainer's data for the current scan run is gone for good, and the failure is completely invisible to operators, it just shows up later as incomplete or stale maintainer data with no error trail anywhere to explain why.

### Fix

Check `.Error` on the `Save` call and at minimum log the failure, and ideally skip appending the record to `linked` on failure so a bad row does not get carried into the association replace.

### LINKED ISSUE #864: Scan Skill Metadata Update Swallows DB Errors
**Location:** `internal/worker/exposure.go` (around line 140, in `doExposure`)

### Description

When setting up an exposure scan, the worker sets the skill name and version on the in-memory scan object and then attempts to persist just those two columns early, via `scan.SkillName = skill.Name`, `scan.SkillVersion = skill.Version`, followed by `w.DB.Model(scan).Updates(map[string]any{"skill_name": skill.Name, "skill_version": skill.Version})`. The result of that `Updates` call is discarded entirely, no `.Error` check, no logging. The identical pattern also exists in `internal/worker/skill.go` around line 155, inside `doSkill`, which is not scoped to a specific finding but carries the exact same bug.

### Why it matters

This does violate the codebase's dominant convention: every other `Save`/`Updates` call site in `internal/worker`, including `worker.go:696,829,1207,1220`, `preflight.go:200`, `skill_parsers.go:97,842,878`, `skill.go:687,859`, `diff_rescan.go:144,217`, `novelty.go:57` and `ecosystems.go:193`, checks `.Error` and handles it. These two call sites, plus the sibling in `doSkill`, are the only exceptions in the package. 

That said, the practical blast radius is narrower than the bug report initially suggested. `scan.SkillName` and `scan.SkillVersion` are already set in memory on the two lines immediately before this call, so nothing later in `doExposure` or `doSkill` ever reads stale data from the database, the in-process struct is correct regardless of whether this particular write lands. On top of that, every scan, on completion, passes through `finalizeScan` in `worker.go:811-829`, which performs a full-row `Save(scan)` that writes the entire in-memory `scan` struct, including the already-correct `SkillName` and `SkillVersion`, back to the database. This self-heals the row before the scan finishes, even if the early `Updates` call silently failed. 

The real, narrower impact is a window while the scan is still running: something reading the `scans` table directly during that window, such as a live dashboard, could see a blank or stale `skill_name`/`skill_version` until the scan reaches its terminal save. It does not cause the pipeline itself to operate on outdated metadata and it does not produce a lasting inconsistency once the scan completes.

### Fix

Check and handle, at minimum by logging, the `.Error` from the `Updates` call in both `exposure.go:140` and the sibling `skill.go:155`, for consistency with the rest of the codebase and to close the mid-run visibility gap.

charlie-morrison:
Closes #863, closes #864.

`parseMaintainersOutput` discards the error from `FirstOrCreate` (`internal/worker/skill.go:959`), which leaves `m` zero-valued when the *lookup* fails rather than the create. The `Save` at `:975` then inserts a second maintainer with an empty `Login`, and `Association("Maintainers").Replace` at `:979` links the repository to that blank row instead of the real one — so the failure mode is a wrong association, not only the lost field update the issue describes. `TestParseMaintainers_lookupFailureSkipsRecord` reproduces it against main: `inserted 1 blank-login maintainer row(s), want 0`, with the association holding `{ID:2 Login:}`.

The failed `Save` is deliberately **not** skipped, which is where this departs from #863's suggested fix. There the row exists and is correctly identified — only its field refresh was lost — so dropping it from `linked` would unlink a real maintainer through the wholesale `Replace`. Applying the suggestion turns `TestParseMaintainers_saveFailureKeepsMaintainerLinked` red with an empty association. The `Replace` error itself is now logged rather than returned, keeping the existing non-fatal behaviour of the explicit `_ =`.

For #864, the two `Updates` calls (`internal/worker/exposure.go:140`, `internal/worker/skill.go:155`) are checked and logged, non-fatal: both fields are already set in memory and `finalizeScan` re-saves the whole row, so a failure only leaves the columns stale for readers watching the table mid-scan. No test there — the failure path needs a full scan harness and its only observable effect is the log line.

`gofmt`, `go build ./...` and `go vet` clean; `internal/worker` green.

### DIFF
--- a/internal/worker/skill.go
+++ b/internal/worker/skill.go
@@ -933,20 +933,44 @@ func (w *Worker) markRetracted(scan *db.Scan, seen map[string]bool) int {
 	return retracted
 }
 
+// reportedMaintainer is one entry of a maintainers skill report.
+type reportedMaintainer struct {
+	Login    string `json:"login"`
+	Name     string `json:"name"`
+	Email    string `json:"email"`
+	Role     string `json:"role"`
+	Status   string `json:"status"`
+	Evidence string `json:"evidence"`
+}
+
+// applyTo copies the fields this report entry refreshes onto the stored row.
+// An absent or unrecognised field leaves the stored value alone, so a partial
+// report never blanks what an earlier scan established.
+func (rm reportedMaintainer) applyTo(m *db.Maintainer) {
+	if rm.Name != "" {
+		m.Name = rm.Name
+	}
+	if validEmail(rm.Email) {
+		m.Email = rm.Email
+	}
+	switch rm.Status {
+	case "active":
+		m.Status = db.MaintainerActive
+	case "inactive":
+		m.Status = db.MaintainerInactive
+	}
+	if rm.Evidence != "" {
+		m.Notes = rm.Role + ": " + rm.Evidence
+	}
+}
+
 // parseMaintainersOutput upserts Maintainer rows and links them to the
 // scanned repo. Mirrors the legacy doMaintainerAnalysis logic so the
 // maintainers skill and the old Go handler stay interchangeable.
 func (w *Worker) parseMaintainersOutput(scan *db.Scan, report string, emit func(Event)) error {
 	var result struct {
-		Maintainers []struct {
-			Login    string `json:"login"`
-			Name     string `json:"name"`
-			Email    string `json:"email"`
-			Role     string `json:"role"`
-			Status   string `json:"status"`
-			Evidence string `json:"evidence"`
-		} `json:"maintainers"`
-		DisclosureChannel string `json:"disclosure_channel"`
+		Maintainers       []reportedMaintainer `json:"maintainers"`
+		DisclosureChannel string               `json:"disclosure_channel"`
 		// Subprojects optionally carries a per-sub-package disclosure channel
 		// for a monorepo, so a report against one gem in rails/rails routes to
 		// that gem's maintainers rather than the repo-wide channel. Additive:
@@ -1007,21 +1031,7 @@ func (w *Worker) parseMaintainersOutput(scan *db.Scan, report string, emit func(
 			w.Log.Warn("upsert maintainer", "scan", scan.ID, "login", rm.Login, "err", err)
 			continue
 		}
-		if rm.Name != "" {
-			m.Name = rm.Name
-		}
-		if validEmail(rm.Email) {
-			m.Email = rm.Email
-		}
-		switch rm.Status {
-		case "active":
-			m.Status = db.MaintainerActive
-		case "inactive":
-			m.Status = db.MaintainerInactive
-		}
-		if rm.Evidence != "" {
-			m.Notes = rm.Role + ": " + rm.Evidence
-		}
+		rm.applyTo(&m)
 		if err := w.DB.Save(&m).Error; err != nil {
 			// Only the field refresh was lost; m still identifies a row that
 			// exists, so it stays in linked. Dropping it would remove a real
--- a/internal/worker/skill.go
+++ b/internal/worker/skill.go
@@ -154,10 +154,15 @@ func (w *Worker) doSkill(ctx context.Context, scan *db.Scan, emit func(Event)) (
 	}
 	scan.SkillName = skill.Name
 	scan.SkillVersion = skill.Version
-	w.DB.Model(scan).Updates(map[string]any{
+	// Non-fatal: the in-memory scan already carries both fields and finalizeScan
+	// saves the whole row at the end, so a failure here only leaves the columns
+	// stale for readers watching the table while the scan runs.
+	if err := w.DB.Model(scan).Updates(map[string]any{
 		"skill_name":    skill.Name,
 		"skill_version": skill.Version,
-	})
+	}).Error; err != nil {
+		w.Log.Warn("update scan skill metadata", "scan", scan.ID, "skill", skill.Name, "err", err)
+	}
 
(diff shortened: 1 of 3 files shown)


charlie-morrison:
Scope note, since #864's body says these were the only exceptions in `internal/worker`: after this PR there are still eight discarded GORM writes in the package, all the same shape — set the field in memory, persist it early, drop the result.

`skill.go:209` (`prompt`), `skill.go:274` (`scope_mode`), `skill.go:309` (`profile`), `skill.go:313` (`backend`), `exposure.go:187` (`prompt`), `worker.go:519` (`session_id`), `worker.go:531` (periodic `log` flush), and `repo_cache.go:74` (`disk_bytes`).

The first seven all write `scans` and are self-healed by the full-row `Save` in `finalizeScan`, exactly like the two fixed here, so they carry the same narrow mid-run visibility cost. **`repo_cache.go:74` is the one that differs**: it writes `repositories.disk_bytes` from the cache sweep, and there is no compensating full-row save on that path, so a failure there is a lasting silent loss rather than a stale window.

Worth noting for the general case that `errcheck` cannot see any of these — GORM returns `*gorm.DB` and hangs the error off `.Error`, so an unchecked write is invisible to the linter that exists for unchecked errors.

Happy to fold the remaining eight into this PR or leave them for a follow-up — say which and I'll do it.

internal/worker/skill.go
```diff
@@ -956,27 +985,28 @@ func (w *Worker) parseMaintainersOutput(scan *db.Scan, report string, emit func(
...
-			m.Status = db.MaintainerActive
-		case "inactive":
-			m.Status = db.MaintainerInactive
+		if err := w.DB.Where(db.Maintainer{Login: rm.Login}).FirstOrCreate(&m).Error; err != nil {
+			// m is a zero value here: nothing was found and nothing was
+			// created. Falling through would Save() a record with no primary
+			// key and an empty Login, inserting a second, blank maintainer row
+			// and linking the repository to that instead of the real one.
+			w.Log.Warn("upsert maintainer", "scan", scan.ID, "login", rm.Login, "err", err)
+			continue
```
alexandre-daubois, line 1036: This skip leaves `linked` holding only the maintainers that resolved, and the unconditional `Association("Maintainers").Replace(linked)` below then drops every skipped one from the repository: a report naming `alice` and `bob` whose first lookup fails ends with `bob` alone linked. The comment on the `Save` branch rejects exactly that outcome, and `TestParseMaintainers_lookupFailureSkipsRecord` cannot catch it because a one-maintainer report leaves `linked` empty and never reaches the `Replace`. It may be worth recording that a record failed to resolve and skipping the `Replace` altogether when the set is known to be partial.

internal/worker/skill.go
alexandre-daubois: The `disclosure_channel` write keeps its discarded error while every other DB write in this function was just given a `Warn`, so a failed update silently leaves a monorepo sub-package routed to a stale or absent channel with nothing in the scan log. Since the PR's scope is the unhandled writes in the maintainer path, you can give this one the same treatment as the two above it.

andrew (changes requested):
Agree with both of Alex's points.

The `continue` at :994 is the one that blocks: a report naming `[alice, bob]` where alice's lookup transiently fails ends with `linked = [bob]`, the `len(linked) > 0` guard passes, and `Replace([bob])` unlinks alice — the exact outcome the comment on the `Save` branch three lines down says is worse than the bug being fixed. `TestParseMaintainers_lookupFailureSkipsRecord` can't see it because a one-maintainer report leaves `linked` empty and never reaches `Replace`. Set a flag on that `continue` and skip the `Replace` (with a warn) when the set is known partial, and make the test a two-maintainer report so the `Replace` path actually runs.

The subproject `disclosure_channel` write at :977 should get the same warn as its neighbours — the comment there justifies non-fatal, not silent.

On the scope note: keep this PR to what it is. The seven `scans` writes are all self-healed by `finalizeScan` the same way as the two here, so a mechanical follow-up is fine. `repo_cache.go:74` is a different shape — no compensating save, lasting loss — so open that as its own PR with a test rather than folding it in.

andrew (approved):

PR #871, opened 2026-08-18
Add supply-chain risk flags to the packages report

abhinavgautam01:
Part of #106: folds the [Trail of Bits supply-chain-risk-auditor](https://github.com/trailofbits/skills/tree/main/plugins/supply-chain-risk-auditor) heuristic checklist into the existing `packages` skill as a `risk_flags` array, rather than adding a separate skill.

Five flags, exactly the ones the issue names, no more:

| Flag | Raised when |
|---|---|
| `single_maintainer` | the registry lists one owner or publisher |
| `no_security_policy` | no `SECURITY.md` anywhere in the checkout |
| `native_extension` | the package ships compiled code, so installing it runs a toolchain |
| `stale_release` | the latest release is more than eighteen months old |
| `maintainer_domain_expired` | a maintainer contact's email domain no longer resolves |

## Report shape

Each entry is `{"id": ..., "evidence": ...}`, where the evidence is one sentence naming what the skill actually saw: the file, the field, the date. The schema requires it to be non-empty. The skill is told to omit any flag it could not check and SKILL.md says so in as many words: an absent flag means "not checked or not found", never "checked and clean". Guessing is worse than silence, because an analyst reads every flag as something the skill actually verified. Under `--hardened` the DNS lookup behind `maintainer_domain_expired` is unavailable, so that flag is simply left off.

## Storage

Only the validated ids reach the new `packages.risk_flags` column, comma-joined and canonically ordered. The evidence sentences stay in `scan.Report`. That is the split `posture` already uses, where the full check list stays in the report while only the tier is promoted to a column, and it keeps the health summary and the package page free of JSON decoding.

`internal/db/package_risk.go` owns the closed set plus `NormalisePackageRiskFlags`, `PackageRiskFlags` and a label map for display, so the parser and the health scorer share one definition instead of repeating flag names. `joinPackageRiskFlags` in the worker is the write side of that pair. An id outside the set is dropped with a warning event naming the package and its ecosystem rather than failing the scan: the rest of the package row is still accurate, while a model inventing a sixth flag name should not cost the repository its whole package set.

New nullable text column, so `AutoMigrate` adds it. No `preMigrate` work, nothing to backfill.

## Health

`RepositoryHealthAssessment` gains a `RiskFlags` union across the repository's packages, deduped and canonically ordered, computed before the unassessed early return so the flags are still carried when there is not enough evidence to classify. `healthSummary` names them in human form, so a classification explains itself.

The flags are advisory and never move the classification. An absent flag means "not checked", not "checked and clean", and `parsePackagesOutput` replaces the whole package set on every run while the health scorer recomputes hourly. Scoring on a flag would therefore let a single scan that skipped the check silently flip a stored verdict back with no upstream change behind it.

Release staleness is instead derived host-side from `Package.LatestReleaseAt`, authoritative upstream data already in a column. The assessment carries `LastReleaseAt`, the newest release across every package the repository publishes, and a repository whose newest release is more than eighteen months old is held at `stale`. Taking the newest release rather than any one package's keeps a monorepo `active` while any of its packages still ships, which matters now that #806 gave sub-packages their own identity. It deliberately does **not** push anything towards `abandoned`, which still requires archival or maintainer evidence. The summary gained a `last release N ago` clause so a stale verdict says why.

## Also

- `risk_flags` on the packages API response plus `openapi.yaml`, so a skill can read back what the last run recorded.
- Badges on the package page, hidden when there are none. Both the badges and the health summary render labels (`ships native code`, `no recent release`) while the column, the API and the skill contract keep the ids.
- `docs/database.md` and `docs/skills.md`.

## Testing

- `NormalisePackageRiskFlags` and `PackageRiskFlags`: canonical reordering, dedup, trimming, empty and unknown ids, nil on empty input. Labels for all five ids plus the fallback for an unrecognised one.
- Parser: flags land comma-joined and canonically ordered whatever order the skill reported them in, a package with no `risk_flags` key gets an empty column, an unknown id is dropped while the known flags on the same package survive and the warning names both the flag and the ecosystem.
- Health: a release older than eighteen months holds a recently-pushed repository at `stale`; a monorepo stays `active` while one package still ships; a `stale_release` flag alongside a recent release date moves nothing; flags union and order across two packages with the summary naming them; the summary names the last release age; flags survive the early-return path while `Health` stays empty.

`gofmt -l`, `go build`, `go vet` (including the `evals` and `podman` tags), `go mod tidy -diff`, `go test -race ./...`, `go test -race -tags evals` and `golangci-lint` are all clean. The one `golangci-lint` hit, `goconst` on `internal/web/audit_test.go:57`, is present on an unmodified `main`.

Remaining on #106 after this: the `security-deep-dive` upgrade (convention-not-bug rule plus the sharp-edges API-misuse probes).

**Note: The `verify` attack-tree line on the tracking comment is already done by #851, just not ticked.**

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -446,7 +446,7 @@ type Package struct {
 	// maintainer_domain_expired. The ids are validated and canonically
 	// ordered on write, so a stored value only ever names known flags. Each
 	// flag's evidence sentence stays in the scan report; promoting only the
-	// ids keeps health scoring and the package page free of JSON decoding.
+	// ids keeps the health summary and the package page free of JSON decoding.
 	RiskFlags string
 
 	CreatedAt time.Time
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -440,6 +440,14 @@ type Package struct {
 	LatestReleaseAt      *time.Time
 	DependentPackagesURL string
 	Metadata             string `gorm:"type:text"`
+	// RiskFlags is the comma-joined set of supply-chain hygiene warnings the
+	// packages skill reported for this package: single_maintainer,
+	// no_security_policy, native_extension, stale_release,
+	// maintainer_domain_expired. The ids are validated and canonically
+	// ordered on write, so a stored value only ever names known flags. Each
+	// flag's evidence sentence stays in the scan report; promoting only the
+	// ids keeps health scoring and the package page free of JSON decoding.
+	RiskFlags string
 
 	CreatedAt time.Time
 }
--- a/internal/db/repository_health.go
+++ b/internal/db/repository_health.go
@@ -2,7 +2,6 @@ package db
 
 import (
 	"fmt"
-	"slices"
 	"strings"
 	"time"
 
@@ -20,9 +19,11 @@ const (
 	RepositoryHealthAbandoned RepositoryHealth = "abandoned"
 	RepositoryHealthZombie    RepositoryHealth = "zombie"
 
-	healthActiveWindow     = 365 * 24 * time.Hour
-	healthAbandonedWindow  = 2 * 365 * 24 * time.Hour
-	healthZombieDependents = 100
+	healthActiveWindow    = 365 * 24 * time.Hour
+	healthAbandonedWindow = 2 * 365 * 24 * time.Hour
+	// A month is thirty days here, matching how healthAge renders one.
+	healthStaleReleaseWindow = 18 * 30 * 24 * time.Hour
+	healthZombieDependents   = 100
 )
 
 // RepositoryHealthAssessment is the durable classification plus the evidence
@@ -34,10 +35,19 @@ type RepositoryHealthAssessment struct {
 	DependentRepos    int
 	ActiveMaintainers int
 	KnownMaintainers  int
+	// LastReleaseAt is the most recent release across every package the
+	// repository publishes, so a monorepo counts as shipping while any one
+	// of its packages still ships. Nil when no package records a release
+	// date.
+	LastReleaseAt *time.Time
 	// RiskFlags are the supply-chain hygiene warnings the packages skill
 	// reported, unioned across every package the repository publishes and
-	// canonically ordered. They are evidence carried alongside the
-	// classification; only stale_release actually moves it.
+	// canonically ordered. They are surfaced in the summary as evidence and
+	// never move the classification: an absent flag means "not checked"
+	// rather than "checked and clean" (SKILL.md says so). The package set is
+	// also replaced wholesale on every run, so scoring on a flag would let a
+	// scan that skipped the check flip a stored verdict with no upstream
+	// change behind it.
 	RiskFlags []string
 }
 
@@ -54,6 +64,9 @@ func AssessRepositoryHealth(repo Repository, packages []Package, maintainers []M
 		// feed cannot prove those sets are disjoint, so max is conservative.
 		assessment.DependentRepos = max(assessment.DependentRepos, pkg.DependentRepos)
 		flags = append(flags, PackageRiskFlags(pkg.RiskFlags)...)
+		if pkg.LatestReleaseAt != nil && (assessment.LastReleaseAt == nil || pkg.LatestReleaseAt.After(*assessment.LastReleaseAt)) {
+			assessment.LastReleaseAt = pkg.LatestReleaseAt
+		}
 	}
 	// The dropped return is discarded: values in the RiskFlags column were
 	// already validated when the packages skill's row was written, so
@@ -69,6 +82,13 @@ func AssessRepositoryHealth(repo Repository, packages []Package, maintainers []M
 		}
 	}
 
+	var releaseAge time.Duration
+	staleRelease := false
+	if assessment.LastReleaseAt != nil {
+		releaseAge = now.Sub(*assessment.LastReleaseAt)
+		staleRelease = releaseAge >= healthStaleReleaseWindow
+	}
+
 	if !repo.Archived && repo.PushedAt == nil && assessment.KnownMaintainers == 0 {
 		return assessment
 	}
@@ -78,12 +98,12 @@ func AssessRepositoryHealth(repo Repository, packages []Package, maintainers []M
 		age = now.Sub(*repo.PushedAt)
 	}
 
-	assessment.Health = repositoryHealth(repo.Archived, repo.PushedAt != nil, age, assessment)
-	assessment.Summary = healthSummary(repo.Archived, repo.PushedAt, age, assessment)
+	assessment.Health = repositoryHealth(repo.Archived, repo.PushedAt != nil, staleRelease, age, assessment)
+	assessment.Summary = healthSummary(repo.Archived, repo.PushedAt, age, releaseAge, assessment)
 	return assessment
 }
 
-func repositoryHealth(archived, hasPush bool, age time.Duration, assessment RepositoryHealthAssessment) RepositoryHealth {
+func repositoryHealth(archived, hasPush, staleRelease bool, age time.Duration, assessment RepositoryHealthAssessment) RepositoryHealth {
 	abandoned := archived || (hasPush && age >= healthAbandonedWindow && assessment.KnownMaintainers > 0 && assessment.ActiveMaintainers == 0)
 	if abandoned {
 		if assessment.DependentRepos >= healthZombieDependents {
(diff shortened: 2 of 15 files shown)


internal/db/repository_health.go
```diff
@@ -79,7 +91,13 @@ func repositoryHealth(archived, hasPush bool, age time.Duration, assessment Repo
...
 		return RepositoryHealthAbandoned
 	}
-	if hasPush && age <= healthActiveWindow && assessment.ActiveMaintainers > 0 {
+	// A package that has not shipped a release in eighteen months is not
+	// reaching its consumers however busy the repository looks, so the flag
+	// holds the classification at stale even when commits are recent. It
+	// deliberately does not push the repository towards abandoned, which
+	// still requires archival or maintainer evidence.
+	staleRelease := slices.Contains(assessment.RiskFlags, string(PackageRiskStaleRelease))
+	if hasPush && age <= healthActiveWindow && assessment.ActiveMaintainers > 0 && !staleRelease {
```
alexandre-daubois, line 118: Reading the absence of `stale_release` as grounds for `active` treats a missing flag as "checked and clean", which `skills/packages/SKILL.md:60` forbids in as many words, and `parsePackagesOutput` replaces the whole package set each run so nothing preserves the previous verdict. A re-run that omits the check — which the same SKILL.md instructs whenever a flag could not be verified — silently flips the stored `health` back to `active` with no upstream change behind it. Since `Package.LatestReleaseAt` already holds the authoritative date, you can derive the eighteen-month test from the column and leave the reported flag purely advisory.

internal/db/repository_health.go
```diff
@@ -42,11 +48,17 @@ type RepositoryHealthAssessment struct {
...
 	for _, pkg := range packages {
 		// Published packages can share downstream repositories. The package
 		// feed cannot prove those sets are disjoint, so max is conservative.
 		assessment.DependentRepos = max(assessment.DependentRepos, pkg.DependentRepos)
+		flags = append(flags, PackageRiskFlags(pkg.RiskFlags)...)
 	}
+	// The dropped return is discarded: values in the RiskFlags column were
+	// already validated when the packages skill's row was written, so
+	// nothing here should be unknown.
+	assessment.RiskFlags, _ = NormalisePackageRiskFlags(flags)
```
alexandre-daubois, line 74: The union is taken over every package of the repository regardless of `Package.SubprojectID`, so one unreleased gem in a `rails/rails`-shaped tree blocks `active` for every sibling — right after #806 gave sub-packages their own identity and pages. `DependentRepos` uses `max` in the conservative direction whereas this union is conservative in the direction of downgrading, so scoping the flags to the assessed sub-package may be the more defensible read.

internal/db/repository_health.go
```diff
@@ -106,6 +124,9 @@ func healthSummary(archived bool, pushedAt *time.Time, age time.Duration, assess
 	if assessment.DependentRepos > 0 {
 		parts = append(parts, fmt.Sprintf("up to %d dependent repos", assessment.DependentRepos))
 	}
+	if len(assessment.RiskFlags) > 0 {
+		parts = append(parts, "risk flags: "+strings.Join(assessment.RiskFlags, ", "))
```
alexandre-daubois: The flag ids are joined verbatim, so a summary otherwise written as a sentence ends in `risk flags: single_maintainer, stale_release`, and the `package_show.html` badges carry the same raw identifiers. A label map beside `packageRiskFlagOrder` would give both surfaces human text without touching the stored column.

abhinavgautam01: Fixed, a `packageRiskFlagLabels` map now sits beside `packageRiskFlagOrder`, as you suggested.

`PackageRiskFlagLabel` falls back to the id when there is no label, so a value stored before a label existed still renders. Both surfaces go through it: the summary reads `risk flags: single maintainer, no recent release` and the badges use a `risklabel` template helper. The stored column, the API and the skill contract all keep the ids.

Two labels are more than a de-underscoring: `native_extension` renders as "ships native code" while `stale_release` renders as "no recent release", which reads better next to the new "last release N ago" clause in the same summary.

internal/db/repository_health.go
abhinavgautam01: Fixed, though not quite the way you framed it.

Rather than scoping to a sub-package, the classification now takes the newest `LatestReleaseAt` across the repository's packages. `Repository.Health` is a repository-level column with no per-subproject counterpart, so the question it answers is "has this repository shipped anything recently", which parallels how `age` reads `PushedAt`. A rails-shaped tree stays `active` while any one of its packages still ships, which is the case you described.

That also puts it on the same conservative footing as `DependentRepos`: `max` in the direction that avoids the downgrade. The flag union is still repository-wide, but it is now purely descriptive, so its breadth costs nothing.

internal/db/repository_health.go
abhinavgautam01: Fixed.

The rule was making exactly the inference SKILL.md forbids and your flapping scenario is worse than it first looks: `repositoryHealthTick` recomputes health hourly off whatever package rows exist, so a single run that skipped the check would flip the stored verdict within the hour and nothing would record that it had ever been otherwise.

Staleness now comes from `Package.LatestReleaseAt` as you suggested. The reported flag is advisory: surfaced on the package page, in the API and in the health summary, never scored. The reasoning is recorded on the `RiskFlags` field comment so the inference does not creep back in later.

internal/worker/skill_parsers.go
```diff
@@ -164,6 +171,11 @@ func (w *Worker) parsePackagesOutput(scan *db.Scan, report string, emit func(Eve
 				row.Metadata = string(b)
 			}
 		}
+		flagIDs := make([]string, 0, len(p.RiskFlags))
+		for _, f := range p.RiskFlags {
+			flagIDs = append(flagIDs, f.ID)
+		}
+		row.RiskFlags = packageRiskFlags(emit, p.Name, flagIDs)
```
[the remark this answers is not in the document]
abhinavgautam01: Fixed. The event now reads `packages: dropping unknown risk flag(s) made_up_flag for foo (gem)`.

It passes `row.Ecosystem`, the value normalised through `db.EcosystemType`, rather than the raw string from the report, so the identifier in the log matches what the row actually stores. The parser test asserts on the ecosystem too.

internal/db/db.go
```diff
@@ -440,6 +440,14 @@ type Package struct {
...
 	DependentPackagesURL string
 	Metadata             string `gorm:"type:text"`
+	// RiskFlags is the comma-joined set of supply-chain hygiene warnings the
+	// packages skill reported for this package: single_maintainer,
+	// no_security_policy, native_extension, stale_release,
+	// maintainer_domain_expired. The ids are validated and canonically
+	// ordered on write, so a stored value only ever names known flags. Each
+	// flag's evidence sentence stays in the scan report; promoting only the
+	// ids keeps health scoring and the package page free of JSON decoding.
+	RiskFlags string
```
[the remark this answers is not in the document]
abhinavgautam01: Leaving this one as a plain `string`, for consistency rather than disagreement about the outcome.

In this model `type:text` marks blobs and long free text: `Package.Metadata`, `Repository.ThreatModel`, `Repository.ScanConfig`, `Finding.Snippet`, `Finding.Locations`. Short strings do not carry it, including the two nearest analogues to this column, `Repository.Posture` and `Repository.Health`, both of which are also enum-ish values documented as `text` in docs/database.md. Every string column in that document is listed as `text`, so `risk_flags` is not out of step there.

`RiskFlags` holds at most five joined ids, well under a hundred characters. If the Postgres work in #658 turns up a dialect that needs the tag, the right fix is a sweep over all of these columns rather than this one.

skills/packages/schema.json
```diff
@@ -24,7 +24,22 @@
...
+            "items": {
+              "type": "object",
+              "required": ["id", "evidence"],
+              "additionalProperties": false,
+              "properties": {
+                "id": {
+                  "type": "string",
+                  "enum": ["single_maintainer", "no_security_policy", "native_extension", "stale_release", "maintainer_domain_expired"]
+                },
+                "evidence": { "type": "string" }
```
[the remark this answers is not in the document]
abhinavgautam01: Fixed, `evidence` now carries `minLength: 1`.

SKILL.md asks for one sentence naming what the skill actually saw, while the schema accepted `""`. Worth catching because the evidence is the only thing distinguishing a verified flag from a guessed one and the flag ids are all a reader gets in the API and the column.

andrew (changes requested):
Alex's three points and the Copilot ones all look properly addressed — deriving the eighteen-month rule from `max(LatestReleaseAt)` instead of the reported flag is the right call, and the `type:text` reasoning holds. Three small things before this lands:

- `internal/db/package_risk.go` — `PackageRiskFlagLabels` (exported func) and `packageRiskFlagLabels` (map var) sit in the same file differing only by the leading capital, and the func's one caller is `healthSummary` in the same package. Drop the func and inline the loop at the call site; that removes an export and the case collision together. Unexporting isn't an option since it would land on the map's name.
- `internal/worker/skill_parsers.go` — `packageRiskFlags` here is the join-for-write side and `db.PackageRiskFlags` is the split-for-read side: inverse operations with near-identical names. Rename the worker one so it says which direction it goes (`joinPackageRiskFlags` or similar).
- The PR body still describes commit-1 behaviour ("One flag moves the classification: `stale_release` blocks `active`" and the judgement-call paragraph). Commit 2 made the flags advisory and moved the rule onto `LatestReleaseAt`. The docs are already right; update the body so the squash message matches what lands.

abhinavgautam01:
ThankDropped `PackageRiskFlagLabels` and inlined the loop in `healthSummary`. You are right that unexporting was not a way out, it would have collided with the map outright. The singular `PackageRiskFlagLabel` stays exported since the `risklabel` template func in `internal/web` needs it alongside `healthSummary`.

Renamed the worker helper to `joinPackageRiskFlags`, and its doc comment now says it is the write side of `db.PackageRiskFlags` rather than leaving the pairing to be inferred from two near-identical names.

PR body updated: the "one flag moves the classification" section and the judgement-call paragraph are replaced by the `max(LatestReleaseAt)` rule, with the flags described as advisory throughout, so the squash message matches what lands.

andrew (approved):

PR #883, opened 2026-08-19
Add provider-specific OpenCode runners

### LINKED ISSUE #877: Support provider-specific OpenCode runner images and configuration
OpenCode supports providers with several different runtime needs. Scrutineer currently supplies direct credential and egress defaults for Anthropic and OpenAI, while other providers may need OpenCode config, stored auth, an adapter package, another executable, or different API and authentication hosts.

#876 exposed this with `kiro/auto`. The model can be added to Scrutineer's picker, but the stock runner has no Kiro catalog entry, adapter package, `kiro-cli`, credential mapping, or Kiro egress policy. OpenCode then reports an unhelpful server error instead of identifying the missing provider.

The existing `runner_image` and language profile design gives us a useful build order:

```
Scrutineer runner -> OpenCode provider image -> repository language profile
```

Language profiles already build from the configured runner image, so provider images should compose with the existing Ruby, Rust, Go, Node, and other profiles.

Add provider-scoped OpenCode runner configuration that can:

- Select a runner image from the provider prefix in a model ID such as `kiro/auto`.
- Pass only the environment variables and stored auth required by that provider.
- Declare model API and authentication hosts that remain available under hardened mode.
- Supply pinned OpenCode config, catalog entries, plugins, provider packages, and supporting binaries without downloading executable code during every scan.
- Check that the provider and selected model are available before accepting scan work.
- Preserve the selected provider and runner image digest on the scan.
- Return the underlying OpenCode configuration error when a provider is unavailable.

Providers included in OpenCode's built-in catalog should be able to use the stock runner and only declare credentials and egress. Providers with extra runtime dependencies should be able to use a derived image, either published by this project or managed by the operator.

Use Kiro as the first acceptance case. A configured `kiro/auto` scan should pass a provider readiness check and run a skill through an automatically selected language profile. Missing credentials, binaries, catalog data, or egress should each produce a specific error.

## Provider audit

The OpenCode 1.17.12 binary in the current runner already bundles the adapters used by Anthropic, OpenAI, Google Gemini and Vertex, Amazon Bedrock, Azure OpenAI, GitHub Copilot, OpenRouter, xAI, Groq, Vercel AI Gateway, Cloudflare AI Gateway, OpenCode Zen, and the common OpenAI-compatible providers. These providers do not need separate images.

An isolated catalog test with model fetching disabled also confirmed that provider-only `OPENCODE_AUTH_CONTENT` activates the built-in catalogs for those providers. Scrutineer can use that path for ordinary API keys instead of adding every provider's credential variable to the harness. The generated auth document must contain only the provider selected by the model prefix; forwarding a user's whole OpenCode auth file would expose unrelated credentials to the scanned repository.

Some providers need extra handling:

- GitHub Copilot's `GITHUB_TOKEN` name overlaps with the token Scrutineer may use for repository access, so it must never be forwarded implicitly.
- Bedrock and Vertex can use credential chains and credential files. Passing a host file path into the container is not sufficient; these need explicit secret-file handling or narrowly scoped mounts.
- OAuth providers need persistent OpenCode auth state. `OPENCODE_AUTH_CONTENT` is a static override and the current `HOME=/tmp` container loses refreshed auth between jobs.
- Azure, Bedrock, and Vertex use resource- or region-specific endpoints, so their egress hosts must come from resolved provider configuration.
- Local Ollama needs host-service port routing. The current proxy restricts the host gateway to the Scrutineer skill API port, so an egress hostname alone cannot expose port 11434.
- `egress_allow` is ignored under hardened mode. Selected provider hosts therefore need to be part of the resolved harness allowlist rather than treated as ordinary user additions.

Kiro remains a derived-image case. It was removed from models.dev in anomalyco/models.dev@08aa06852395acefd9ae7d7d85f03d1efe9b214b, and first-party OpenCode support is still pending in anomalyco/opencode#26680. Kiro now supports non-interactive `KIRO_API_KEY` authentication, but OpenCode still needs a pinned external adapter or plugin. The provider mechanism should allow an operator-managed image while this project separately decides whether to publish and maintain a Kiro image.

## Proposed configuration shape

The audit suggests a provider-prefix map that keeps ordinary API keys, native credential chains, OAuth state, and derived images separate:

```yaml
opencode:
  providers:
    groq:
      api_key_env: GROQ_API_KEY
      egress_allow:
        - api.groq.com
    kiro:
      runner_image: registry.example/scrutineer-opencode-kiro@sha256:...
      config_file: ./opencode/kiro.json
      pass_env:
        - KIRO_API_KEY
      required_binaries:
        - kiro-cli
      egress_allow:
        - q.us-east-1.amazonaws.com
        - runtime.us-east-1.kiro.dev
        - management.us-east-1.kiro.dev
        - prod.us-east-1.auth.desktop.kiro.dev
```

`api_key_env` can generate selected-provider-only `OPENCODE_AUTH_CONTENT`. `pass_env` should be an explicit allowlist for compound or native credentials, and `required_binaries` gives readiness a way to distinguish a missing helper from a missing catalog model. A separate owner-only `state_dir` can hold rotating OAuth auth, but its `auth.json` must not contain another provider. Provider `egress_allow` entries need to join the resolved harness allowlist so they work under hardened mode; the top-level egress option should keep its current hardened behavior.

Related to #876.

andrew:
Adds provider-scoped OpenCode configuration for credentials, custom runner images, config files, OAuth state, required binaries, and hardened egress.

Provider images compose with the existing language profiles. Each configured provider is checked with `opencode models` before a skill runs, and scans record the selected provider, image, and resolved digest. The operator docs and sample config cover stock providers and operator-managed images such as Kiro.

Closes #877.

### DIFF
--- a/docs/opencode.md
+++ b/docs/opencode.md
@@ -161,7 +161,10 @@ must have owner-only permissions such as `0700`. When `state_dir` is the only
 credential source, its auth file must already contain the selected provider. A
 provider using `pass_env` may initialise the file on first use.
 `state_dir` and `api_key_env` cannot be combined because
-`OPENCODE_AUTH_CONTENT` would override the rotating file.
+`OPENCODE_AUTH_CONTENT` would override the rotating file. Scans that share a
+`state_dir` run one at a time so a token refresh from one scan is never
+overwritten by another; use `api_key_env` or `pass_env` for providers where
+concurrent scans matter more than persisted OAuth.
 
 Without `state_dir`, OpenCode's auth data stays with the scan's retry state and
 is deleted after that scan lineage completes. Logs, repository data, and
--- a/docs/opencode.md
+++ b/docs/opencode.md
@@ -1,104 +1,187 @@
-# Opencode backend
+# OpenCode backend
 
-Scrutineer can drive [opencode](https://opencode.ai) instead of claude-code,
-selected with `-backend opencode` (or `backend: opencode` in
-`scrutineer.yaml`). Unlike claude and codex, opencode is provider-agnostic:
-the model you configure determines which API it talks to, so credential and
-egress handling are looser than for the single-provider backends.
+Scrutineer can run [OpenCode](https://opencode.ai) instead of claude-code. Set
+`-backend opencode`, or use `backend: opencode` in `scrutineer.yaml`. Model ids
+use OpenCode's `provider/model` form, and this backend requires the container
+runner. Scrutineer rejects `--no-container` when it is selected.
 
-## Setup
-
-The runner image bundles the `opencode` binary, so there's nothing to install.
-opencode reads provider credentials from its auth config or from the
-provider's own env var; the harness passes through `ANTHROPIC_API_KEY`,
-`OPENAI_API_KEY`, `OPENCODE_CONFIG_CONTENT` and `OPENCODE_AUTH_CONTENT` from
-the host so the common cases work without extra setup:
+The stock runner contains OpenCode and its common built-in providers. Existing
+Anthropic and OpenAI installations can keep using host credentials without a
+provider block:
 
     export ANTHROPIC_API_KEY=sk-ant-...
     go run ./cmd/scrutineer -skills ./skills -backend opencode
 
-or in `scrutineer.yaml`:
-
     backend: opencode
     default_model: anthropic/claude-sonnet-5
     models:
-      - name: Sonnet (via opencode)
-        id:   anthropic/claude-sonnet-5
-      - name: GPT-5 (via opencode)
-        id:   openai/gpt-5
+      - name: Sonnet via OpenCode
+        id: anthropic/claude-sonnet-5
 
-Model ids are in opencode's `provider/model` form. For providers other than
-Anthropic and OpenAI, set `OPENCODE_CONFIG_CONTENT` (an inline JSON config
-opencode reads at startup) with the provider block, and add the provider's
-API host to `egress_allow:` so the proxy lets it through.
+## Provider configuration
 
-The opencode backend requires the containerised runner. `--no-container` with
-`-backend opencode` is rejected at startup.
+Add an `opencode.providers` entry when a provider needs different credentials,
+egress, config, stored auth, or a derived runner image. The map key must match
+the provider prefix in the model id.
 
-## How the harness maps
+    backend: opencode
+    default_model: groq/llama-3.3-70b-versatile
+    models:
+      - name: Llama 3.3 70B via Groq
+        id: groq/llama-3.3-70b-versatile
 
-| Aspect | claude | opencode |
-| --- | --- | --- |
-| Binary | `claude` | `opencode` |
-| Argv | `claude -p --output-format stream-json ...` | `opencode run --format json --auto ...` |
-| Skill staging | `./.claude/skills/{name}/SKILL.md` | `./.opencode/skill/{name}/SKILL.md` |
-| Project memory | `CLAUDE.md` | `AGENTS.md` |
-| Egress hosts | `*.anthropic.com` | `models.dev`, `api.openai.com`, `*.anthropic.com` |
-| Credential env | `ANTHROPIC_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN` | `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `OPENCODE_CONFIG_CONTENT`, `OPENCODE_AUTH_CONTENT` |
-| State dir env | `CLAUDE_CONFIG_DIR` | `OPENCODE_CONFIG_DIR`, `OPENCODE_DB` |
-| Account-error phrases | claude usage/plan/access messages | union of common provider rate-limit / quota / invalid-key messages |
+    opencode:
+      providers:
+        groq:
+          api_key_env: GROQ_API_KEY
+          egress_allow:
+            - api.groq.com
 
-opencode globs `./.opencode/{skill,skills}/**/SKILL.md` with symlinks enabled,
-so `stageSkill` writes the same files it always has into
-`./.opencode/skill/{name}/`. The activation prompt points at that path
-explicitly (opencode discovers but does not auto-invoke skills in headless
-`run` mode). `PROFILE.md` lands at `AGENTS.md`, which opencode walks from cwd
-up to the project root.
+`api_key_env` names a variable in Scrutineer's host environment. For each
+scan, Scrutineer generates `OPENCODE_AUTH_CONTENT` containing only the selected
+provider and gives the container the variable name rather than putting its
+value in the runtime argv. A configured provider does not inherit
+`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, or a host-wide OpenCode auth document.
 
-`--auto` suppresses opencode's interactive permission prompts; the container
-is the sandbox. A chat turn is therefore held to its prompt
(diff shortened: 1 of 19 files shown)


internal/worker/opencode_provider.go
```diff
@@ -0,0 +1,473 @@
...
+}
+
+func (d ContainerRunner) appendOpencodeStateArgs(args []string, harnessStateDir string, provider opencodeProvider) []string {
+	if harnessStateDir != "" {
+		// Logs, repository metadata, and all other OpenCode data stay scoped to
+		// this scan lineage. A configured provider mounts only auth.json below.
+		args = append(args, "-e", "XDG_DATA_HOME=/harness-state/data")
+	}
+	if provider.StateDir != "" && harnessStateDir != "" {
+		args = append(args, "-v", bindMount(
```
p-linnane, line 240: `auth.json` is mounted read-write into every scan, while OpenCode can refresh OAuth credentials and rewrite it without cross-process locking. Concurrent scans can therefore race rotating refresh tokens and leave stored auth stale. Can we serialize readiness and scan execution per canonical `state_dir`?

internal/worker/opencode_provider.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+	return env
+}
+
+// runnerImageContentDigest records the locally resolved content identity of
+// the provider base. Registry-backed images use RepoDigests; locally built
+// operator images fall back to their immutable image ID.
+func runnerImageContentDigest(ctx context.Context, rt ContainerRuntime, image string) string {
+	const format = `{{if .RepoDigests}}{{index .RepoDigests 0}}{{else}}{{.Id}}{{end}}`
+	out, err := exec.CommandContext(ctx, rt.bin(), "image", "inspect", "--format", format, "--", image).Output()
```
p-linnane, line 490: Apple container 1.2.2 doesn’t support `image inspect --format`, so this returns an empty digest for every Apple scan. Can we parse `configuration.descriptor.digest` from its JSON output, falling back to `id`?

internal/worker/opencode_provider.go
```diff
@@ -0,0 +1,473 @@
...
+			return fmt.Errorf("OpenCode provider %q readiness failed while reaching configured egress host %q: %w: %s", provider.ID, host, err, cappedProviderOutput(out))
+		}
+	}
+	args := d.buildRunArgsForProvider(absWork, image, hnet, harnessStateDir, provider, "/tmp")
+	args = append(args, "opencode", "models", provider.ID)
+	cmd := exec.CommandContext(ctx, d.Runtime.bin(), args...)
+	cmd.Env = environmentWith(os.Environ(), provider.Env)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return classifyOpencodeReadinessError(provider, strings.TrimSpace(string(out)), err)
```
p-linnane: This is the only readiness path that persists uncapped command output. Can we pass `cappedProviderOutput(out)` to the classifier?

p-linnane (changes requested):
This looks good overall, but I think two issues should be fixed before merging: concurrent `state_dir` scans need serialization, and Apple container needs runtime-specific image digest handling. Details inline.

internal/worker/egress.go
```diff
@@ -38,3 +44,24 @@ func WaitHostAPIReachable(ctx context.Context, host, port string) error {
...
+		return 0, noop, errors.New("egress: Token is required")
+	}
+	ln, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return 0, noop, err
+	}
+	srv := &http.Server{Handler: p, ReadHeaderTimeout: scopedEgressReadHeaderTimeout}
+	go func() { _ = srv.Serve(ln) }()
+	closeProxy := func() { _ = srv.Close() }
+	return ln.Addr().(*net.TCPAddr).Port, closeProxy, nil
```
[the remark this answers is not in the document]
p-linnane: I don’t think this applies here. In Go 1.26.6, a wildcard TCP listen either creates a dual-stack socket with `IPV6_V6ONLY=0` or falls back to IPv4, so IPv4 container clients can connect.

internal/worker/container.go
```diff
@@ -209,43 +231,80 @@ func proxyURLWithHost(proxyURL, host string) string {
...
 
+type containerRunErrorState struct {
+	accountText  string
+	providerText string
+	rateLimit    *RateLimitInfo
+}
+
+func (s *containerRunErrorState) observe(event Event, h Harness, configuredProvider bool) {
+	s.accountText = preferAccountErrText(s.accountText, h.AccountErrorText(event.Text))
+	if configuredProvider && event.Kind == KindError && event.Text != "hit max turns" {
```
alexandre-daubois, line 242: Latching `providerText` on every `KindError` event makes `resumeRetryable()` false for any configured-provider run that emitted an error before exiting, so a resume whose session is gone never restarts fresh and `scan.SessionID` keeps a dead id that each later retry replays. It may be worth narrowing the latch to the last error of the run, or to the classes `classifyOpencodeProviderRunError` actually distinguishes, so the fresh-restart fallback stays reachable for a configured provider.

docs/opencode.md
```diff
@@ -1,104 +1,191 @@
...
+then runs `opencode models <provider>` from `/tmp` in the selected
+language-profile image. This keeps repository OpenCode files out of the check.
+The selected `provider/model` must appear exactly in the output. Missing
+OpenCode binaries, adapter packages, supporting binaries, egress, and catalog
+models produce separate errors. OpenCode errors from the skill run are also
+kept in the scan error instead of being replaced by the container exit status.
+
+## Stored auth
+
+OAuth providers can use `state_dir` instead of `api_key_env`. Scrutineer mounts
```
alexandre-daubois, line 154: The rewrite drops the T1 residual paragraph the previous version carried, in the change that widens what enters the sandbox to native credential chains such as `AWS_SECRET_ACCESS_KEY` and a cross-scan `auth.json`. Restating here that in-container code can read those values, and can also write the file under `state_dir`, would keep the residual described in `threatmodel.md` visible at the point where a provider is configured.

internal/worker/opencode_provider.go
andrew: Done in 01c79d2: `OpencodeReadinessCache` now holds a per-`state_dir` mutex acquired before `ensureOpencodeProviderState` reads `auth.json` and released when the scan finishes, so two scans never mount the same file concurrently. `docs/opencode.md` notes the throughput trade-off.

internal/worker/opencode_provider.go
andrew: Done in 01c79d2: the Apple runtime now runs plain `container image inspect` and reads `configuration.descriptor.digest` from the JSON, falling back to `id` for locally built images.

internal/worker/opencode_provider.go
andrew: Done in 01c79d2: the classifier now receives `cappedProviderOutput(out)`.

internal/worker/container.go
andrew: Good catch, done in 01c79d2: `resumeRetryable` gates on account errors only again, so a missing-session error still falls through to the fresh restart. `providerText` is kept for `failure()` classification only.

docs/opencode.md
andrew: Restored in 0beb8e7 in the credential section, covering `api_key_env`/`pass_env` readability and the read-write `state_dir` mount.

p-linnane (approved):

PR #887, opened 2026-08-20
Isolate federation tests from host git config

### LINKED ISSUE #886: federation tests leak host git config (trace2 listener inflates rev-list --all)
`TestExportFeed_picksUpAThirdPartyCommit` fails deterministically, and `TestExportFeed_republishesOntoARecreatedRemote` / `TestRepoDelete_reopensTheRecordsAppliedToIt` fail intermittently, on machines whose global git config sets `trace2.eventtarget` to a listener that writes back into observed repositories (e.g. the `git-ai` background daemon, which attaches notes under `refs/notes/ai`).

### Cause

The federation test helpers shell out to `git` with the host's environment, so global git config leaks in. Two effects:

- `remoteCommitCount` runs `git rev-list --count --all`. `--all` includes `refs/notes/*`, so a daemon-written notes ref inflates the count and the "unchanged record set must add no commit" assertion fires with `got 3`.
- `t.TempDir` cleanup sporadically fails with `unlinkat … .git/objects: directory not empty` because the daemon is still writing loose objects when the test returns.

A second, independent leak: the peer `commit` in the same test does not pass `-c commit.gpgsign=false`, so on a host with `commit.gpgsign=true` and no usable signer the peer commit dies with `gpg failed to sign the data`.

### Repro

```sh
# passes:
GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_NOSYSTEM=1 \
  go test ./internal/web/... -run TestExportFeed_picksUpAThirdPartyCommit

# fails on a host with a trace2 listener that writes refs/notes/*:
go test ./internal/web/... -run TestExportFeed_picksUpAThirdPartyCommit
```

### Fix

- Run every test-side `git` invocation in `federation_test.go` under `testutil.GitEnv()` (already exists for exactly this purpose), so host git config cannot reach the temp repos.
- Count `--branches` instead of `--all` in `remoteCommitCount`, so stray notes/tags can never inflate the count regardless of what else touches the repo.

mechko:
Fixes #886.

## Problem

`federation_test.go` shells out to `git` with the host environment. On a machine whose global git config sets `trace2.eventtarget` to a listener that writes back into observed repos (the `git-ai` daemon attaches notes under `refs/notes/ai`), the bare test remotes grow an extra ref and `remoteCommitCount`'s `rev-list --count --all` over-counts:

```
--- FAIL: TestExportFeed_picksUpAThirdPartyCommit
    federation_test.go:357: an unchanged record set must add no commit on top of the peer's, got 3
```

The other two `remoteCommitCount` callers are timing-dependent on the same mechanism, and `t.TempDir` cleanup sporadically reports `.git/objects: directory not empty` because the daemon is still writing. Separately, a global `commit.gpgsign=true` with no usable signer kills the peer commit.

## Fix

- Add a `testGit` helper that runs git under `testutil.GitEnv()` (already exists for exactly this) and route every test-side git exec through it. Host config no longer reaches the temp repos, and identity is pinned so the per-call `-c user.*` overrides go.
- `remoteCommitCount`: count `--branches` instead of `--all`, so notes/tags can never inflate the assertion regardless of what else touches the repo.

## Verification

On a host with the trace2 listener active:

```
go test ./internal/web/... -run 'TestExportFeed_picksUpAThirdPartyCommit|TestExportFeed_republishesOntoARecreatedRemote|TestRepoDelete_reopensTheRecordsAppliedToIt' -count=5
```
15/15 pass, no `TempDir RemoveAll` noise. Full `./internal/web/...` passes.

### DIFF
--- a/internal/web/federation_test.go
+++ b/internal/web/federation_test.go
@@ -17,8 +17,24 @@ import (
 
 	"scrutineer/internal/db"
 	"scrutineer/internal/interchange"
+	"scrutineer/internal/testutil"
 	"scrutineer/internal/worker"
 )
+
+// testGit runs git under testutil.GitEnv so the host's global git config
+// cannot leak into the temp repositories these tests create. A trace2
+// listener that writes refs/notes/* into observed repos, or a global
+// commit.gpgsign, otherwise changes commit counts and breaks peer commits.
+func testGit(t *testing.T, args ...string) string {
+	t.Helper()
+	cmd := exec.Command("git", args...)
+	cmd.Env = testutil.GitEnv()
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("git %v: %v\n%s", args, err, out)
+	}
+	return string(out)
+}
 
 func feedTime(t *testing.T) time.Time {
 	t.Helper()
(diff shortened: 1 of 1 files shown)


andrew (changes requested):
Please restore `t.Helper()` inside both local `run` closures at lines 348 and 506. `testGit` marks only its own frame as a helper, so Git failures are reported at the closure definition instead of the failing `run(...)` invocation.

mechko:
Restored `t.Helper()` in both `run` closures (eae4204) — thanks for the catch.

andrew (approved):
The updated head restores `t.Helper()` in both closures and addresses the requested change.

PR #889, opened 2026-08-20
Leave repository health empty until required skills have run

### LINKED ISSUE #888: Repository health shows 'stale' before required skills have run
## Problem

The Health column in the repositories table shows **stale** for demonstrably active projects (e.g. `curl/curl`).

`repositoryHealth` in `internal/db/repository_health.go` returns `RepositoryHealthStale` as the fall-through default whenever any one of four conjuncts fails:

```go
if hasPush && age <= healthActiveWindow && assessment.ActiveMaintainers > 0 && !staleRelease {
    return RepositoryHealthActive
}
return RepositoryHealthStale
```

The `maintainers` skill is currently only run manually. So a repository that has completed `metadata` (fresh `pushed_at`) but not `maintainers` has `ActiveMaintainers == 0` and is stamped `stale` — indistinguishable from a genuinely stale project.

The existing "unassessed" escape hatch only fires when `PushedAt == nil` **and** `KnownMaintainers == 0`, so a set `pushed_at` bypasses it.

## Expected

The Health column stays **empty** until every skill the classification depends on (`metadata`, `maintainers`) has a completed run. A partial pipeline must not produce a verdict.

## Scope

- Gate `AssessRepositoryHealth` on scan completion, not on row presence, so "maintainers ran and found nobody" remains distinguishable from "maintainers never ran".
- Archived repositories may still be classified immediately — `archived` is a definitive upstream flag that needs no corroboration.
- The hourly rescore should clear a previously-stamped verdict back to empty when evidence is now known to be incomplete.

mechko:
Fixes #888.

## Problem

`repositoryHealth` returns `stale` as the fall-through default. The `maintainers` skill is currently manual-only, so any repo that has run `metadata` but not `maintainers` gets `ActiveMaintainers == 0` and is stamped **stale** — including very active projects like curl.

The old escape hatch (`PushedAt == nil && KnownMaintainers == 0`) was bypassed as soon as `metadata` set `pushed_at`.

## Change

- `repositoryHealthSkills` names the skills the classification reads from (`metadata`, `maintainers`).
- New `RepositoryHealthEvidenceComplete(gdb, repoID)` reports whether every one of those skills has a `status = done` scan row. This answers "did the skill run", not "did it produce rows", so a maintainers scan that legitimately found nobody still counts.
- `AssessRepositoryHealth` takes an `evidenceComplete` flag; when false it returns the collected evidence (counts, risk flags) with `Health` left empty. `Archived` remains a short-circuit — it's a definitive upstream signal on its own.
- `RefreshRepositoryHealth` and the live repo view both compute the flag from scan rows. The hourly rescore will now also clear a previously-stored verdict back to empty when evidence is incomplete.
- `packages` is deliberately not gated on: its absence softens the verdict (no zombie/stale-release signal) but never inverts it.

## Tests

- New table cases: `recent push without required scans is unassessed` (the curl scenario) and `archived is classified without full evidence`.
- New `TestRefreshRepositoryHealth_leavesHealthEmptyWithoutRequiredScans` covers the persisted path, including clearing a stale value that was already stored.
- Existing `Refresh*` tests seed the required scan rows via a helper.

### DIFF
--- a/internal/db/repository_health.go
+++ b/internal/db/repository_health.go
@@ -190,7 +190,8 @@ func healthAge(age time.Duration) string {
 func RepositoryHealthEvidenceComplete(gdb *gorm.DB, repositoryID uint) (bool, error) {
 	var done []string
 	if err := gdb.Model(&Scan{}).
-		Where("repository_id = ? AND status = ? AND skill_name IN ?", repositoryID, ScanDone, repositoryHealthSkills).
+		Where("repository_id = ? AND kind = ? AND status = ? AND skill_name IN ?",
+			repositoryID, "skill", ScanDone, repositoryHealthSkills).
 		Distinct("skill_name").Pluck("skill_name", &done).Error; err != nil {
 		return false, err
 	}
--- a/internal/db/repository_health.go
+++ b/internal/db/repository_health.go
@@ -26,6 +26,13 @@ const (
 	healthZombieDependents   = 100
 )
 
+// repositoryHealthSkills are the skills whose output the classification reads.
+// A repository is left unassessed until every one of them has a completed run,
+// so a partial pipeline (metadata done, maintainers still manual) cannot stamp
+// a misleading verdict. packages is deliberately absent: its absence softens
+// the verdict (no zombie/stale-release signal) but never inverts it.
+var repositoryHealthSkills = []string{"metadata", "maintainers"}
+
 // RepositoryHealthAssessment is the durable classification plus the evidence
 // used to reach it. Summary is derived at read time for detailed views; only
 // Health is stored on the repository row.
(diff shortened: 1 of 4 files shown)


andrew (changes requested):
Please rebase onto current `main` and remove inherited commit `66f151` from merged PR #885. The current head is conflicting, includes the already-merged `internal/worker/profile*` changes, and has no CI runs, so the health change cannot be reviewed yet.

mechko:
Rebased onto `main` (4a2ab34) and dropped the inherited `66f1510` from #885 — the branch now carries only the health-gate commit. Build, tests, and `golangci-lint run` are clean locally. Ready for another look.

andrew (changes requested):
`RepositoryHealthEvidenceComplete` should restrict the query to `kind = "skill"`. Imported reports create completed `kind = "import"` scans and copy the uploaded tool name into `skill_name`, so imports named `metadata` and `maintainers` can satisfy the gate without either skill running. Please add the predicate and a regression test covering imported scan rows.

andrew (approved):

PR #896, opened 2026-08-21
Support host-local OpenCode providers via host_port

### LINKED ISSUE #895: Support host-local OpenCode providers (Ollama, LM Studio, llama.cpp)
The OpenCode backend can't use a model server running on the host machine. `opencode.providers.<id>.egress_allow` opens named hostnames through the allowlisting proxy, but the proxy only permits scrutineer's own API port on the container-host gateway, so `host.docker.internal:11434` (Ollama) or `:1234` (LM Studio) is refused. The `egress_allow` validator also rejects entries with a port. `docs/opencode.md` currently notes this for Ollama only.

Reported for LM Studio in #876.

A fix needs a way to open a specific host-gateway port to the scan container per provider, without exposing every host service, and needs to work under `--hardened` (per-scan `--internal` network + sidecar proxy) as well as the default host-proxy path. The provider's `config_file` would then point `options.baseURL` at the gateway alias rather than `127.0.0.1`.

andrew:
`opencode.providers.<id>.host_port` opens one operator-named TCP port on the host loopback to the scan container, so a host-local model server (Ollama, LM Studio, llama.cpp) is reachable at `host.docker.internal:<port>` through the egress proxy. The port is carried on the per-scan scoped host proxy (`egress.Proxy.HostPorts`, added in alpha-omega-security/harness#12) and through `SCRUTINEER_PROXY_HOST_PORTS` for the rootless sidecar; scans whose provider has no `host_port` never see the extra port. Readiness probes the port from inside the container on every scan (outside the readiness cache, via a forced `--proxytunnel` CONNECT so a proxy 502/403 fails curl), so a stopped model server or wrong port fails the scan up front.

A provider with only `host_port` needs no `egress_allow` entry. `docs/opencode.md` gains a Host-local providers section with an Ollama example; `threatmodel.md` T13 covers the widened host-loopback surface. Apple's container runtime is documented as unsupported for this option since it has no `host.docker.internal` alias.

Stacked on #894. Bumps `alpha-omega-security/harness` to v0.1.8.

Closes #895.


internal/worker/opencode_provider.go
```diff
@@ -414,6 +423,26 @@ func (d ContainerRunner) checkOpencodeReadiness(ctx context.Context, provider op
...
+// because a host-local server can be stopped or restarted between scans while
+// the image, catalog, and egress hosts the cache covers stay unchanged.
+// --proxytunnel forces CONNECT for the plain-http target so a proxy 403/502
+// (port denied or nothing listening on the host loopback) is a curl exit
+// error; once the tunnel is up, any origin status counts as reachable.
+func (d ContainerRunner) checkOpencodeHostPort(provider opencodeProvider, probe func(...string) ([]byte, error)) error {
+	if provider.HostPort == "" {
+		return nil
+	}
+	target := "http://" + HostGatewayAlias + ":" + provider.HostPort + "/"
```
p-linnane, line 442: This unconditional alias target fails on non-hardened Apple. In that mode the provider proxy recognizes only the gateway IP, so `host.docker.internal` is not gated or rewritten to loopback and the proxy returns 502 during DNS lookup. I reproduced the failure: adding `HostGatewayAlias` to the provider proxy’s `APIHosts` fixes it while an unlisted port still returns 403.

p-linnane (changes requested):
One behavior issue to fix before merge, plus a few documentation corrections Copilot already noted. Otherwise, this looks sound.

internal/worker/opencode_provider.go
alexandre-daubois, line 442: This builds the probe target from `host_port` rather than from the `options.baseURL` the provider's `config_file` sets, so a config pointing at `127.0.0.1` or at a different port passes readiness and then fails inside the OpenCode server, which is the failure this check exists to replace. It may be worth checking the loaded `OPENCODE_CONFIG_CONTENT` for a `baseURL` naming `host.docker.internal:<host_port>` at startup, so a mismatch is refused before a scan runs.

docs/opencode.md
```diff
@@ -93,13 +93,51 @@ The top-level `egress_allow` remains ignored in hardened mode. Use hostnames
...
+`host_port` and `egress_allow` may be combined when a provider needs both a
+host-local endpoint and external hostnames. `host_port` requires the docker or
+podman runtime; Apple's container runtime has no `host.docker.internal` alias
+and is not supported for this option.
+
+Under the sidecar egress path (rootless podman with `--hardened`) the request
+goes container → sidecar → host gateway → host loopback, which requires the
+rootless network backend to forward the host gateway to the host loopback. The
+sidecar already refuses to start when that forwarding is unavailable for the
+skill API port; the same check applies here.
```
alexandre-daubois: The sidecar's startup gate probes the skill API at `-addr`, which defaults to `127.0.0.1:8080` but accepts any address, so an operator on `-addr 0.0.0.0:8080` gets a sidecar that starts without the backend forwarding the host gateway to the host loopback. A loopback-bound model server is then out of reach behind a green sidecar, so you may want to qualify this as holding only for a loopback `-addr`.

threatmodel.md
```diff
@@ -160,6 +160,8 @@ Under **rootless** podman the host proxy lives across the `--internal` boundary
...
...the skill API port plus the one operator-named port", for scans using that provider only (a per-scan scoped proxy on the host path; the per-scan sidecar receives the port through `SCRUTINEER_PROXY_HOST_PORTS`). The proxy rewrites `host.docker.internal:<port>` to the host loopback the same way it rewrites the skill API port, so no other host loopback service is reachable and the process-wide proxy that scans without a `host_port` provider go through never receives the extra port. The residual is that in-container code a hostile repo runs can reach whatever listens on that port: Ollama and LM Studio serve an unauthenticated OpenAI-compatible API, so such code could enumerate loaded models, pull a model into the server, or send its own prompts, and under `--hardened` the response is one of the few places findings text could be exfiltrated to. An operator who sets `host_port` accepts that surface for the named port. Bind the model server to loopback only (the Ollama and LM Studio default) so the grant is the sole path to it, and prefer a server whose API can be scoped to inference only. Apple's container runtime has no `host.docker.internal` alias, so `host_port` is docker/podman only.
```
alexandre-daubois: T13's rootless-sidecar paragraph above names the `host.docker.internal`-only-on-the-API-port restriction as what stops a proxy-using workload from reaching other host ports through the sidecar, and this paragraph is the change that widens it. Qualifying that sentence with the `host_port` grant would keep the two halves of the same section from disagreeing.

internal/worker/opencode_provider.go
```diff
@@ -171,6 +174,11 @@ func (d ContainerRunner) configureOpencodeProviderEgress(provider opencodeProvid
 	}
 	d.Egress.Allow = appendUniqueStrings(d.Egress.Allow, provider.EgressHosts...)
 	d.ProviderProxy.Allow = appendUniqueStrings(d.ProviderProxy.Allow, provider.EgressHosts...)
+	var hostPorts []string
+	if provider.HostPort != "" {
+		hostPorts = []string{provider.HostPort}
+		d.Egress.HostPorts = appendUniqueStrings(d.Egress.HostPorts, provider.HostPort)
```
alexandre-daubois, line 178: Opening an extra host-loopback port leaves no trace in the logs: neither the scoped proxy nor the sidecar records which port was granted, so the config file is the only evidence of a widened host surface. A `Log.Info` here naming the provider and the port would put the grant in the same audit trail that carries the proxy's denials.

scrutineer.sample.yaml
```diff
@@ -118,15 +118,19 @@
 # instead. A derived runner_image can add a pinned adapter or supporting binary;
 # it must be based on the Scrutineer runner so language profiles still compose.
 # config_file and state_dir are host paths; relative paths start beside this
-# config file. Every configured provider must declare its API/auth hostnames.
-# Only the selected provider's hosts are added to a scan, including in hardened
-# mode.
+# config file. Every configured provider must declare its API/auth hostnames
+# under egress_allow, or a host_port for a model server on the host loopback
```
alexandre-daubois, line 122: This comment picks up `host_port` while the `opencode.providers` sentence in `README.md:412` lists only credentials, hardened-mode egress hosts, OAuth state, OpenCode config, and a derived runner image. Extending that sentence with the host-local port would stop the README describing the pre-`host_port` shape of the feature.

internal/worker/opencode_provider.go
```diff
@@ -171,6 +174,11 @@ func (d ContainerRunner) configureOpencodeProviderEgress(provider opencodeProvid
 	}
 	d.Egress.Allow = appendUniqueStrings(d.Egress.Allow, provider.EgressHosts...)
 	d.ProviderProxy.Allow = appendUniqueStrings(d.ProviderProxy.Allow, provider.EgressHosts...)
+	var hostPorts []string
```
alexandre-daubois: `hostPorts` and `d.Egress.HostPorts` end up holding the same one-element list, since nothing populates `Egress.HostPorts` before this point. Passing `d.Egress.HostPorts` to the scoped proxy below would drop the local and leave one place where the port is set.

internal/worker/opencode_provider.go
```diff
@@ -414,6 +423,26 @@ func (d ContainerRunner) checkOpencodeReadiness(ctx context.Context, provider op
...
+}
+
+// checkOpencodeHostPort probes the provider's host-local model server from
+// inside the container. It runs on every scan (outside the readiness cache)
+// because a host-local server can be stopped or restarted between scans while
+// the image, catalog, and egress hosts the cache covers stay unchanged.
+// --proxytunnel forces CONNECT for the plain-http target so a proxy 403/502
+// (port denied or nothing listening on the host loopback) is a curl exit
+// error; once the tunnel is up, any origin status counts as reachable.
+func (d ContainerRunner) checkOpencodeHostPort(provider opencodeProvider, probe func(...string) ([]byte, error)) error {
```
alexandre-daubois: Nothing in this body reads the receiver, which is what forces the failure-case test to build a `ContainerRunner` it never uses. Dropping `d` and making this a plain function would match what the code does and leave callers passing only `provider` and `probe`.

internal/worker/opencode_provider_test.go
```diff
@@ -295,6 +366,47 @@ func TestCheckOpencodeReadinessFindsExactModel(t *testing.T) {
...
+	data, _ = os.ReadFile(log)
+	if strings.Contains(string(data), "opencode models") {
+		t.Errorf("cached call re-ran the catalog probe: %s", data)
+	}
+	if !strings.Contains(string(data), "http://"+HostGatewayAlias+":11434/") {
+		t.Errorf("cached call skipped the host-port probe: %s", data)
+	}
+
+	runtime = writeFakeRuntime(t, "#!/bin/sh\necho 'curl: (56) CONNECT tunnel failed, response 502' >&2; exit 56\n")
+	d = ContainerRunner{Harness: OpencodeHarness{}, Runtime: ContainerRuntime{Bin: runtime}}
```
alexandre-daubois: Nothing reads this `ContainerRunner`: the closure below runs `exec.Command(runtime, ...)` and `checkOpencodeHostPort` ignores its receiver, so the assignment is scaffolding the failure case does not need. Dropping it leaves the same coverage with one variable less.

internal/worker/opencode_provider.go
andrew: Fixed in 0d52b36: `apiHosts` now always includes `host.docker.internal` alongside the runtime-specific container host, so the scoped proxy recognises the alias on Apple and applies the port gate + loopback rewrite. The docker/podman-only wording is dropped from `docs/opencode.md` and `threatmodel.md` in 0b8be03. Thanks for reproducing.

andrew: 0d52b36: `loadOpencodeProviders` refuses a `host_port` provider at startup when the loaded `config_file` does not contain `host.docker.internal:<port>`, with tests for the no-config, `127.0.0.1`, and wrong-port cases. Substring rather than JSON parse so it does not depend on the config's exact shape.

internal/worker/opencode_provider.go
andrew: 2e7e459: added `d.ProviderProxy.Log.Info("host-local provider port granted on host loopback", "provider", provider.ID, "port", provider.HostPort)` at the grant site, nil-guarded so tests without a logger stay green.

internal/worker/opencode_provider.go
andrew: 2e7e459: dropped the local; the scoped proxy now takes `d.Egress.HostPorts`.

internal/worker/opencode_provider.go
andrew: 2e7e459: `checkOpencodeHostPort` is now a plain function.

internal/worker/opencode_provider_test.go
andrew: 2e7e459: gone along with the `writeFakeRuntime` and `os/exec` import; the failure case now passes a closure that returns the curl output and error directly.

docs/opencode.md
andrew: 0b8be03: qualified for a non-loopback `-addr`. The per-scan readiness probe still fails the scan in that case, so the qualification is about which check catches it rather than a gap.

threatmodel.md
andrew: 0b8be03: the rootless-sidecar sentence now names the `host_port` grant alongside the API-port restriction. Also split T13's four long paragraphs at their topic breaks while I was in there.

scrutineer.sample.yaml
andrew: 0b8be03: `README.md:412` now lists a host-local model server port among the `opencode.providers` fields.

docs/opencode.md
```diff
@@ -93,13 +93,51 @@ The top-level `egress_allow` remains ignored in hardened mode. Use hostnames
...
-so allowing a hostname does not expose a host-local model server's port.
-
 A model whose provider prefix is neither `anthropic`, `openai`, nor a key under
 `opencode.providers` is refused before the container starts.
 
+## Host-local providers
+
+For a model server running on the host machine's loopback (Ollama, LM Studio,
+llama.cpp, or another OpenAI-compatible endpoint), set `host_port` instead of
+`egress_allow`. The egress proxy opens exactly that port on
```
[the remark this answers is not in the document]
andrew: 0b8be03: reworded to "at least one `egress_allow` hostname or a `host_port`", and the `--hardened` credential-reach sentence three lines above is fixed too.

alexandre-daubois (approved):

p-linnane (approved):

PR #901, opened 2026-08-24
Add ocaml runner profile

andrew:
Adds an OCaml profile so opam/dune projects get a working toolchain, dependency install, and reproducer path. Prompted by https://anil.recoil.org/notes/rumour-is-the-exploit which name-checks scrutineer for the OCaml ecosystem; the cohttp 6.3.0 path-traversal case (OSEC-2026-16) is the concrete target.

The image carries opam 2.5.2 (upstream static binary, sha256-pinned per arch; trixie's 2.3.0 predates the CVE-2026-57825 symlink fix), a compiled OCaml 5.5.0 switch, dune, ocamlfind, and the C headers for the common `conf-*` depexts. The opam package index is the default `opam.ocaml.org` HTTP index, whose `repo` file carries `archive-mirrors: "cache"`, so scan-time `opam update` and every archive fetch go through `opam.ocaml.org` regardless of where a package's primary URL points; that host is on the harness egress allowlist as of v0.1.9 (alpha-omega-security/harness#13). `OPAMROOT` lives under `/opt` and is world-writable so `opam install . --deps-only` works under the worker's `--user <host-uid>` runtime.

Profile detection keys on brief's `opam` package_manager with a `Dune` build and `OCaml` language fallback, registered before `c-cpp` so an opam package that also ships a compat Makefile still routes here.

Also: `*.opam.locked` / `_build` / `_opam` in `BuiltinSkipPaths`, `probe.ml` in the poc bundle map, opam solver-failure markers in `depResolutionMarkers`, and harness bumped to v0.1.9.

Verified against a live cohttp scan: profile detection routes there, the image builds, and `opam install fmt` as the host uid with `erratique.ch` / `github.com` / `release-assets.githubusercontent.com` blackholed still fetches everything from `opam.ocaml.org/cache`.

Separate follow-up (not OCaml-specific): `release-assets.githubusercontent.com` is where GitHub 302s release-attachment downloads and it's not in `DefaultAllow`; every profile that fetches from GitHub releases hits it.


docker/profiles/ocaml/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+        "https://github.com/ocaml/opam/releases/download/${OPAM_VERSION}/opam-${OPAM_VERSION}-${opam_arch}-linux" \
+ && echo "${opam_sha}  /usr/local/bin/opam" | sha256sum -c - \
+ && chmod +x /usr/local/bin/opam \
+ # bwrap doesn't nest in a container. Default HTTP index: opam.ocaml.org
+ # is on the egress allowlist and its repo file carries archive-mirrors,
+ # so scan-time `opam update` and every archive fetch go through that one
+ # host regardless of where a package's primary URL points.
+ && opam init --disable-sandboxing --bare --no-setup \
+ && opam switch create default "${OCAML_VERSION}" \
+ && opam install dune ocamlfind \
```
avsm:
```suggestion
 && opam install dune ocamlfind ocaml-lsp-server odoc merlin \
```

and the agent will have more useful tools available as well.

PR #913, opened 2026-08-26
feat(interchange): ask federation peers before reporting a finding

### LINKED ISSUE #14: Interchange and federation

alexandre-daubois:
Fix #14 once for all 🙂


andrew (changes requested):
This is careful work — redirect refusal with the loopback-replay reasoning, credential/query/fragment rejection at startup, response body cap, concurrent peer queries so one hanging peer can't hide another, opt-out and `file://` repos silent, VINCE and outreach-skill enqueue both gated before any outreach happens, and the queued-scan window called out as a known gap in the docs rather than papered over. threatmodel.md T14 updated to match. One small hole and two optional tidy-ups:

**A recorded claim survives a reject/duplicate and is treated as acknowledgement on reopen.** `WriteFindingField` clears `federation_claim_contacts` only when the status write is `reported` (`internal/db/finding_helpers.go:92`). If the analyst sees the banner and clicks Reject instead, the claim stays on the row. If the finding is later reopened to `ready`, `claimPeerHold` sees `FederationClaimContacts != ""` (`internal/web/claim_check_out.go:168`) and returns immediately as already-acknowledged, so the report goes through without asking peers again — stale coordination standing in for a fresh check. Clearing on any status change (drop the `== FindingReported` clause; the no-op guard at `finding_helpers.go:67` already stops a same-value write from reaching it) closes this without adding a second call site.

**Redundant repository load on the outreach-enqueue path** (optional). `enqueueSkillWith` loads `repo` with `url, federation_opt_out_at` at `server.go:3288` and returns early on opt-out; `refuseClaimedOutreach` → `claimPeerHold` → `peerClaims` then loads the same row with the same columns at `claim_check_out.go:86`. One extra `First` per manually-triggered enqueue — cheap, but `peerClaims` could take the repo (or its URL) and let the two other callers load it.

**`TestPeerClaims_oneHangingPeerDoesNotHideAnother` waits ~2s** (optional). The 2s outer context is what unblocks the hanging goroutine; a 200ms context would prove the same thing.

alexandre-daubois:
Comments addressed

PR #914, opened 2026-08-26
chore(deps): update go modules and upgrade to Go 1.27

alexandre-daubois:
Supersedes #912


andrew (changes requested):
Thanks for this. Since it was opened, main merged the renovate Go/Docker/module bumps (#918, #920, #922), so most of the diff is now redundant and merging as-is would revert claude-code 2.1.237 → 2.1.234 in `Dockerfile`, `Dockerfile.runner`, and `threatmodel.md`.

The parts renovate doesn't cover, and that are still worth having:

- `docker/profiles/go/Dockerfile`: `GO_VERSION=1.27.0` and both `GO_SHA256_*` (checksums verified against go.dev), plus `GOVULNCHECK_VERSION=v1.7.0`
- `docker/profiles/go/PROFILE.md`: Go 1.26 → Go 1.27
- `README.md`: Go 1.26+ → Go 1.27+
- `threatmodel.md`: the four `go1.26.5` references → `go1.27.0`

Could you rebase onto main and trim to just those? Everything in `go.mod`/`go.sum` and the `FROM golang:` lines can be dropped — renovate manages them (the trixie images are just waiting on the 7-day `minimumReleaseAge`), and `git-pkgs/clone` v0.5.0→v0.7.0 in particular should come through on its own.

I've opened #925 to bring the govulncheck profile pin and the threatmodel Go-version prose under renovate so they stop drifting.

alexandre-daubois:
Here you go!

Issue #725, opened 2026-07-27
Cap finding severity by attacker prerequisites


andrew:
`security-deep-dive` emits a severity per finding and most confirmed sinks come out `High`. There's no rule set bounding severity by what the attacker actually gains.

Add a set of hard caps applied after severity is assigned, keyed on the attacker's starting position and the finding's preconditions. The principle is that impact is bounded by the new capability the attacker gains over what they already needed to trigger the bug. Examples from the mantis rule set:

- attacker already needs a shell on the host: force `Low` (a shell-equivalent finding adds nothing)
- attacker needs long-term physical access: force `Low`
- attack vector is local only: cap at `Medium`
- outcome depends on a probabilistic LLM response: cap at `High`
- authenticated internal user attacking another tenant's data they could already request through support: cap at `Medium`
- `Critical` reserved for unauthenticated, no-interaction, remote code execution or equivalent

Record which caps fired on the finding (a string list is enough) so an analyst can see why a `High` became a `Medium`. `UNKNOWN` on any input never lowers severity; it flags the finding as incompletely calibrated instead.

This can live in `security-deep-dive`'s rating step, or as a separate `calibrate` skill that runs over all open findings after `revalidate`. Separate skill is cleaner since it also applies to imported and semgrep findings.

Source: [google/mantis](https://github.com/google/mantis) `mantis-calibrate/references/calibration_rules.md` has 27 rules in three tiers (force-Low, cap-High, cap-Medium), Apache-2.0. We'd adapt a subset; the full list includes memory-safety and confidential-computing rules that don't apply to most of what we scan.

andrew:
The mantis caps above key on attacker starting position, which the model currently asserts in prose. Structured design controls (separate issue, to land in `skills/threat-model/schema.json`) give a machine-checkable input: if the finding's file matches a `controls[]` entry of `kind: authorization` and #111's `control_bypass` row records no bypass, the "authenticated internal user" cap applies without trusting the model's own precondition claim. Same for `kind: sandbox` → local-only cap.

[VVAH](https://github.com/visa/visa-vulnerability-agentic-harness/blob/d91b28d3af7ee46f9cc1827a9377cbdedd1f6b7e/vvaharness/pipeline/stages/s6_verify.py#L235-L276) does the glob-match half of this (control → path) but only uses it to force `FALSE_POSITIVE` when fully mitigated; using a partial mitigation to cap severity rather than reject is the addition here.

PR #938, opened 2026-08-29
Apply control-backed finding severity caps

abhinavgautam01:
Part of #725

## Summary

Builds on the structured `control_bypass` evidence introduced in #936 to apply deterministic finding-severity caps from host-reconciled authorization and sandbox controls.

When verification establishes that a matched authorization or sandbox control held, the finding is capped at `Medium`. Unknown, unresolved, unavailable or unsupported control evidence never lowers severity and instead marks the finding as incompletely calibrated.

This implements the machine-checkable control-backed portion of #725. Broader prerequisite rules based on shell access, physical access, probabilistic model behavior and other prose-derived attacker conditions remain follow-up work.

## Changes

- Add finding projections for:
  - Current severity-cap reasons
  - Incomplete severity calibration
- Cap `Critical` and `High` findings at `Medium` when a matched:
  - `authorization` control is verified as held
  - `sandbox` control is verified as held
- Leave severity unchanged when controls are:
  - `bypassed`
  - `not_applicable`
- Mark calibration incomplete without lowering severity when:
  - Control resolution is unavailable
  - An assessment is unresolved or not attempted
  - A held control has an unsupported kind
  - The existing severity is unknown
- Apply severity changes through `db.WriteFindingField` with system provenance.
- Persist severity changes, cap projections, verification records and notes in one transaction.
- Preserve the previous authoritative calibration when a verification report is ungraded or fails host reconciliation.
- Keep parser retries idempotent.
- Expose cap reasons and incomplete state through:
  - Finding API responses
  - OpenAPI documentation
  - Finding detail UI
- Document the database fields and verify behavior.
- Add unit, integration, rollback, API and UI coverage.

## Tests

- `go mod tidy -diff`
- `go test -race ./...`
- `go test -tags evals ./...`
- `go test -race ./internal/worker -run 'Test(ParseVerifySeverityCalibrationRollsBackWithVerification|
ParseVerifyAppliesControlSeverityCapAtomically|ParseVerifyUngradedReportPreservesSeverityCalibration|CalibrateControlSeverity)$' -count=1`
- `go vet ./...`
- `go vet -tags podman ./...`
- `git diff --check`


openapi.yaml
```diff
@@ -1584,6 +1584,8 @@ components:
         sinks:         { type: string, description: Comma-joined sink ids. }
         title:         { type: string }
         severity:      { type: string, enum: [Critical, High, Medium, Low] }
+        severity_caps: { type: array, items: { type: string }, description: "Deterministic reasons that currently cap severity, derived from the latest reconciled verification control assessment." }
+        severity_calibration_incomplete: { type: boolean, description: "True when unknown or unavailable control evidence prevented complete severity calibration; unknown inputs never lower severity." }
```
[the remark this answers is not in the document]
abhinavgautam01: The OpenAPI description now includes unknown severity, unavailable resolution, unresolved or unattempted assessments and unsupported held-control kinds.

internal/worker/severity_calibration.go
```diff
@@ -0,0 +1,91 @@
...
+			continue
+		}
+		result.Caps = append(result.Caps, reason)
+		result.Severity = capFindingSeverity(result.Severity, controlSeverityCap)
+	}
+
+	slices.Sort(result.Caps)
+	if len(result.Caps) > 0 && !knownFindingSeverity(current) {
+		result.Incomplete = true
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: An unknown current severity now always marks calibration incomplete, even when no cap rule fires. Added coverage for the no-controls case.

internal/worker/skill_parsers.go
```diff
@@ -1175,6 +1192,19 @@ func (w *Worker) recordVerifyOutput(
 				return fmt.Errorf("update status: %w", err)
 			}
 		}
+		if calibration.Evaluated {
+			if calibration.Severity != f.Severity {
+				if err := db.WriteFindingField(tx, f.ID, "severity", calibration.Severity, db.SourceSystem, verifySkillName); err != nil {
+					return fmt.Errorf("apply severity cap: %w", err)
+				}
+			}
```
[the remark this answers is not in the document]
abhinavgautam01: Cap application now reads the latest stored severity through a retry/CAS helper, applies only downward changes, and never raises an already-lower severity. Added database and stale-snapshot regression coverage.

internal/worker/skill_parsers.go
```diff
@@ -1175,16 +1194,32 @@ func (w *Worker) recordVerifyOutput(
...
+		if calibration.Evaluated {
+			effectiveSeverity, err := db.ApplyFindingSeverityCap(
+				tx, f.ID, calibration.Maximum, db.SourceSystem, verifySkillName,
+			)
+			if err != nil {
+				return fmt.Errorf("apply severity cap: %w", err)
+			}
+			calibration = calibration.withSeverity(effectiveSeverity)
+			if err := tx.Model(&db.Finding{}).Where("id = ?", f.ID).Updates(map[string]any{
+				"severity_caps":                   strings.Join(calibration.Caps, "\n"),
```
alexandre-daubois: This overwrites `severity_caps` on every evaluated run while the cap on `severity` is one-way, so a finding lowered to `Medium` because an `authorization` control held keeps that `Medium` after a later run reports the same control `bypassed` and returns `confirmed`, with no cap reason left on the row to account for it. You may want to restore the severity recorded in `FindingHistory` when `calibration.Maximum` is empty, or leave the previous reasons in place so the row still explains the value it carries.

internal/worker/severity_calibration.go
```diff
@@ -0,0 +1,87 @@
...
+		switch control.Kind {
+		case threatmodel.KindAuthorization:
+			reason = fmt.Sprintf("authorization control %q held; severity capped at %s", control.ID, controlSeverityCap)
+		case threatmodel.KindSandbox:
+			reason = fmt.Sprintf("sandbox control %q held; severity capped at %s", control.ID, controlSeverityCap)
+		default:
+			result.Incomplete = true
+			continue
+		}
+		result.Caps = append(result.Caps, reason)
```
alexandre-daubois, line 68: This appends a reason for every held `authorization` or `sandbox` control whatever the finding's own severity, so a `Low` finding carries a `severity_caps` line reading `severity capped at Medium` in the API, in the verify note and under Severity on the finding page. It may be worth dropping the reasons in `withSeverity` once the effective severity turns out to sit below `Maximum`.

internal/db/db.go
```diff
@@ -800,14 +800,20 @@ type Finding struct {
...
-	Location   string
+	FindingID string // e.g. F1, F2 within the report
+	Sinks     string // comma-joined sink IDs
+	Title     string
+	Severity  string `gorm:"index"`
+	// SeverityCaps is the newline-delimited set of deterministic cap reasons
+	// applied by the latest authoritative verification. Unknown calibration
+	// inputs never lower Severity; SeverityCalibrationIncomplete records that
+	// an analyst still needs to resolve them.
+	SeverityCaps                  string `gorm:"type:text"`
```
alexandre-daubois, line 821: `findingExport` in `internal/web/api_export.go` mirrors every other `Finding` column and misses both of these, so the unauthenticated JSONL export never emits them while `FindingExport` inherits both through the `FindingDetail` and `FindingSummary` `allOf` chain in `openapi.yaml`. Adding the two keys there keeps the export in line with the contract it declares.

internal/worker/severity_calibration.go
```diff
@@ -0,0 +1,87 @@
...
+		result.Maximum = controlSeverityCap
+	}
+
+	slices.Sort(result.Caps)
+	return result
+}
+
+func (calibration findingSeverityCalibration) withSeverity(severity string) findingSeverityCalibration {
```
alexandre-daubois: `Severity` is set here and read nowhere outside this package's own test, and the branch below repeats the cap `ApplyFindingSeverityCap` already applied to the value it returns, which leaves `Incomplete` as the only field this changes in production. Folding the unknown-severity check into the caller and dropping the field would reduce this to one condition.

abhinavgautam01: Removed the unused calibration Severity field and the duplicate cap application logic. The caller now only handles unknown-severity completeness and suppression of inactive cap reasons after database reconciliation.

internal/db/db.go
abhinavgautam01: findingExport now includes severity_caps as an array and severity_calibration_incomplete as a boolean, with export coverage matching the OpenAPI contract.

internal/worker/severity_calibration.go
abhinavgautam01: Cap reasons are now omitted when the effective severity is below the cap, and also for unknown severities that were not lowered. Unknown severities remain marked as calibration-incomplete. Added coverage for both Low and unknown severities.

internal/worker/skill_parsers.go
abhinavgautam01: Removing a cap now restores the pre-cap severity only when the latest severity history entry is the system-owned verify cap. Any later analyst or model severity change remains authoritative. I added direct and end-to-end cap-removal tests.

andrew (changes requested):
Thanks — the calibration path and restore logic look right after the earlier rounds. A few things before merge.

**Blocking**

`internal/worker/severity_calibration.go:65-68` — a `held` control whose kind is neither `authorization` nor `sandbox` sets `result.Incomplete = true`. `internal/threatmodel/threatmodel.go:42-47` names four more kinds (`input-validation`, `csrf`, `rate-limit`, `other`) and it is an open enum. On a repo whose threat model matches, say, a rate-limit control to a finding location, every verified finding shows "Calibration incomplete" permanently with nothing for an analyst to resolve. Drop `result.Incomplete = true` from the kind-`default` case (just `continue`), remove the "unsupported held-control kind" clause from `openapi.yaml` and `docs/database.md`, and adjust the `"unknown held control is incomplete"` test case.

**Non-blocking**

- `internal/web/api_reads.go:17-18` — `severityCapsKey` / `severityCalibrationKey` consts. None of the other keys in `findingSummary` or `findingExport` use consts; inline the string literals to match the surrounding code.
- `internal/db/finding_helpers.go:127` — the comment says "retry-protected read". `retryFindingWrite` runs one savepointed attempt with no retry when handed an open transaction (`finding_helpers.go:249-253`), and the only production caller passes `tx`, so there is no retry in this path. Behaviour is fine (a CAS conflict rolls the outer transaction back); trim the comment to "same read that decided the write".
- `internal/worker/skill_parsers.go:1172-1182` — `recordVerifyOutput` grew to 9 positional params. Move the calibration mutation and `verifyNote` call back into `parseVerifyOutput` and pass the finished `note` string as before, or bundle `result`/`rubric`/`score`/`gradingError`/`calibration` in a struct.

**CI**

`tests (ubuntu-24.04)` is red on `TestWriteFindingField_concurrentUpdatesKeepHistoryChain` (`SQLITE_BUSY` after 5 retries). That test is unchanged from main and macOS passed, so it looks like a pre-existing flake rather than this change. Please re-run; if it recurs on this branch and not on main we can look closer.

abhinavgautam01:
Thanks, Fixed. Held non-cap control kinds are now ignored without marking calibration incomplete and the schema, docs, and tests are aligned. I also inlined the API keys, corrected the helper comment, and bundled verify persistence inputs while keeping reconciliation transactional. Full race tests, vet, lint and 20 repeated runs of the flaky SQLite test pass locally.

PR #942, opened 2026-08-30
Reap orphaned scan workspaces on startup

### LINKED ISSUE #933: SweepRunning leaves scan workspaces on disk after unclean restart
`db.SweepRunning` marks scans that were running when the previous process died as failed, but only touches the database row. The dead process never reached the `os.RemoveAll(w.scanWorkRoot(scan))` at the end of `wrap()`, and the new process does not remove anything, so `data/work/scan-{id}/` and `data/work/harness-state/scan-{id}/` are left behind for every scan that was in flight at the time of the crash.

https://github.com/alpha-omega-security/scrutineer/blob/6f4da3b/internal/db/db.go#L1998-L2006
https://github.com/alpha-omega-security/scrutineer/blob/6f4da3b/cmd/scrutineer/main.go#L625
https://github.com/alpha-omega-security/scrutineer/blob/6f4da3b/internal/worker/worker.go#L875-L878

This is separate from the pre-#446 leftovers: before a7b4eba (21 Jun) the cleanup was gated on `status == ScanDone`, so every failed or cancelled scan kept its workspace. #446 widened that to `status.Terminal()` for the normal path but did not sweep the existing directories, and the restart path still bypasses it entirely.

Proposed fix: a startup reaper alongside the existing orphan sidecar/network sweep in `main.go` that walks `data/work/scan-*`, looks each id up, and removes the directory when the scan is terminal and no queued scan has it as `resumed_from_scan_id`. That covers both the crash case and any historical orphans on first run. Alternatively `SweepRunning` could return the affected ids for `RemoveScanArtifacts`, but the worker is not constructed yet at that point.

abhinavgautam01:
Fixes #933

## Summary

Adds startup cleanup for scan workspaces left behind when _Scrutineer_ exits before a running scan reaches its normal terminal cleanup.

The reaper runs after legacy harness-state migration and before queue handlers are registered.

## Changes

- Inspect exact `scan-<id>` directories under the worker data directory during startup.
- Remove workspaces belonging to terminal scans.
- Remove the corresponding persisted harness-state directory at the same time.
- Preserve workspace roots referenced by queued, running or paused resume scans.
- Preserve state-only directories used by normally failed resumable scans.
- Ignore malformed directory names, unrelated files and scans that are not terminal.
- Continue processing other entries when an individual lookup or removal fails, while reporting the accumulated error.
- Log the number of orphaned scan artifacts removed.

## Tests

- Add coverage for workspaces left by scans that were running during a restart.
- Add coverage for historical failed-scan workspaces.
- Verify queued and paused resume lineages remain intact.
- Verify active scan workspaces and state-only resume stores are preserved.
- Verify malformed entries and regular files are ignored.
- Verify a missing workspace root is a no-op.

## Verification

- `go mod tidy -diff`
- `go test ./...`
- `go test -race ./internal/worker -count=1 -timeout=5m`
- `go test -race ./internal/web -count=1 -timeout=5m`
- `go vet ./...`
- `go vet -tags podman ./...`
- `golangci-lint v2.12.2`
- `git diff --check`

### DIFF
--- a/internal/worker/artifact_sweep.go
+++ b/internal/worker/artifact_sweep.go
@@ -1,0 +1,95 @@
+package worker
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"scrutineer/internal/db"
+)
+
+// sweepOrphanScanArtifacts removes workspaces left behind when a process exits
+// before finalizeScan reaches its terminal cleanup. A workspace that still
+// backs an active resume must survive so the resumed harness can find its
+// source tree and session store.
+func (w *Worker) sweepOrphanScanArtifacts() (int, error) {
+	if w.DB == nil || w.DataDir == "" {
+		return 0, nil
+	}
+	entries, err := os.ReadDir(w.DataDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return 0, nil
+		}
+		return 0, fmt.Errorf("read scan workspace root: %w", err)
+	}
+
+	var removed int
+	var sweepErr error
+	for _, entry := range entries {
+		scanID, ok := scanWorkspaceEntryID(entry)
+		if !ok {
+			continue
+		}
+		reap, err := w.scanArtifactsReapable(scanID)
+		if err != nil {
+			sweepErr = errors.Join(sweepErr, err)
+			continue
+		}
+		if !reap {
+			continue
+		}
+		if err := w.RemoveScanArtifacts(scanID); err != nil {
+			sweepErr = errors.Join(sweepErr, fmt.Errorf("remove scan %d artifacts: %w", scanID, err))
+			continue
+		}
+		removed++
+	}
+	return removed, sweepErr
+}
+
+func scanWorkspaceEntryID(entry os.DirEntry) (uint, bool) {
+	if !entry.IsDir() {
+		return 0, false
+	}
+	rawID, ok := strings.CutPrefix(entry.Name(), "scan-")
+	if !ok || rawID == "" {
+		return 0, false
+	}
+	id, err := strconv.ParseUint(rawID, 10, strconv.IntSize)
+	if err != nil || id == 0 {
+		return 0, false
+	}
+	return uint(id), true
+}
+
+func (w *Worker) scanArtifactsReapable(scanID uint) (bool, error) {
+	var scan struct {
+		ID     uint
+		Status db.ScanStatus
+	}
+	if err := w.DB.Model(&db.Scan{}).
+		Select("id, status").
+		Where("id = ?", scanID).
+		Limit(1).
+		Find(&scan).Error; err != nil {
+		return false, fmt.Errorf("load scan %d for artifact sweep: %w", scanID, err)
+	}
+	if scan.ID == 0 || !scan.Status.Terminal() {
+		return false, nil
+	}
+
+	var activeResumes int64
+	if err := w.DB.Model(&db.Scan{}).
+		Where("resumed_from_scan_id = ? AND status IN ?", scanID, []db.ScanStatus{
+			db.ScanQueued,
+			db.ScanRunning,
+			db.ScanPaused,
+		}).
+		Count(&activeResumes).Error; err != nil {
+		return false, fmt.Errorf("check scan %d resume references: %w", scanID, err)
+	}
+	return activeResumes == 0, nil
+}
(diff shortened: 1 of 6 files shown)


internal/worker/artifact_sweep.go
```diff
@@ -0,0 +1,95 @@
...
+			sweepErr = errors.Join(sweepErr, err)
+			continue
+		}
+		if !reap {
+			continue
+		}
+		if err := w.RemoveScanArtifacts(scanID); err != nil {
+			sweepErr = errors.Join(sweepErr, fmt.Errorf("remove scan %d artifacts: %w", scanID, err))
+			continue
+		}
```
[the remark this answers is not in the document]
abhinavgautam01: The sweep now separates workspace removal from harness-state cleanup and preserves state for resumable failed or max-turns-hit scans. Active queued, running and paused resume roots remain completely untouched.

internal/worker/artifact_sweep_test.go
```diff
@@ -0,0 +1,124 @@
...
+		t.Fatal(err)
+	}
+	if removed != 2 {
+		t.Fatalf("removed = %d, want 2", removed)
+	}
+
+	for _, id := range []uint{crashed.ID, historical.ID} {
+		assertPathMissing(t, w.workRoot(id))
+		assertPathMissing(t, w.harnessStateDirID(id))
+	}
```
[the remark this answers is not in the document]
abhinavgautam01: Updated the tests. Failed scans and max-turns-hit scans with a captured session now retain their harness state while their orphan workspace is removed. Failed scans without a resumable session still have unusable state cleaned up.

internal/worker/artifact_sweep.go
```diff
@@ -0,0 +1,109 @@
...
+// sweepOrphanScanArtifacts removes workspaces left behind when a process exits
+// before finalizeScan reaches its terminal cleanup. A workspace that still
+// backs an active resume must survive so the resumed harness can find its
+// source tree and session store. A resumable terminal scan can shed its stale
+// workspace, but keeps the harness state needed by a future retry.
+func (w *Worker) sweepOrphanScanArtifacts() (int, error) {
+	if w.DB == nil || w.DataDir == "" {
+		return 0, nil
+	}
+	entries, err := os.ReadDir(w.DataDir)
```
alexandre-daubois: This walks workspace directories only, so a `harness-state/scan-<id>` with no workspace beside it stays out of reach: `finalizeScan` clears the session store on plain `done` alone and then removes the workspace for every terminal status, which leaves a cancelled scan's harness state on disk for good. Walking `<DataDir>/harness-state` in the same pass would catch those entries.

internal/worker/artifact_sweep.go
```diff
@@ -0,0 +1,109 @@
...
+		MaxTurnsHit bool
+	}
+	if err := w.DB.Model(&db.Scan{}).
+		Select("id, status, session_id, max_turns_hit").
+		Where("id = ?", scanID).
+		Limit(1).
+		Find(&scan).Error; err != nil {
+		return false, false, fmt.Errorf("load scan %d for artifact sweep: %w", scanID, err)
+	}
+	if scan.ID == 0 || !scan.Status.Terminal() {
```
alexandre-daubois: A workspace whose scan row is gone lands in the same branch as a live scan, so it survives every restart: `repoDelete` commits the row deletion first and reclaims the disk afterwards through `RemoveScanArtifacts` calls that only log a failure. Treating `scan.ID == 0` as reapable would pick up those historical orphans alongside the crash case.

abhinavgautam01: Artifacts whose scan row no longer exists are now reapable after checking that no active resume references the ID. Added coverage for removing both workspace and harness-state residue in this case.

internal/worker/artifact_sweep.go
abhinavgautam01: The sweep now inspects both workspace and harness-state directories, with coverage proving that state-only residue from cancelled scans is removed.

andrew (changes requested):
Thanks for iterating on this — the sweep logic and coverage look right now.

One thing before merge: `artifact_sweep.go:131-132` duplicates the resumable predicate from `internal/web/scans.go:369-370` (`Status == ScanFailed || (Status == ScanDone && MaxTurnsHit)` plus `SessionID != ""`). This predicate decides whether harness-state survives, so the two copies drifting would mean the sweep deletes state the retry path still considers resumable.

Please add `func (s *Scan) Resumable() bool` on `db.Scan` (in `internal/db/db.go`, next to `MaxTurnsHit`), load into a `db.Scan` instead of the anonymous struct at `artifact_sweep.go:97-102` (keep the `Select`), and call the method from both `scanArtifactSweepDecision` and `resumeOpts`.

abhinavgautam01:
Thanks, Fixed. Added db.Scan.Resumable() as the shared source of truth and updated both the artifact sweep and retry path to use it. The sweep still performs a narrow column selection and table coverage now verifies all resumable and non-resumable combinations.

PR #948, opened 2026-08-31
fix(skills): accept release_watch in the triage report schema

alexandre-daubois:
`skills/triage/SKILL.md` tells the model to record ids in a `release_watch` field, but `schema.json` sets `additionalProperties: false` without declaring it, so `ValidateSkillReport` rejects any triage run that enqueues one. The schema declares the key with the same shape as `verify`, and the release-watch paragraph spells out the empty case the way the `verify` one does.


andrew (approved):
LGTM, real bug — `additionalProperties: false` with the field missing would fail every triage run on a repo with a `fixed` finding.

PR #949, opened 2026-08-31
unbreak Dockerfile

avsm:
The Python dep was upgraded to 3.15rc, but a later stage (pydantic) then broke as there's no wheel for it.

With this change I can deploy Scrutineer!


andrew (approved):

andrew:
Thanks!

PR #956, opened 2026-09-01
Handle fingerprint backfill database errors

### LINKED ISSUE #915: Return and Handle Errors from the Finding Fingerprint Backfill
### Summary

`db.BackfillFindingFingerprints` silently ignores database errors while loading findings that need fingerprints and while updating each matching row. A locked, unavailable, or failing database can therefore leave all or part of the fingerprint backfill incomplete without preventing startup or producing an actionable error.

### Location

- `internal/db/fingerprint.go`, in `BackfillFindingFingerprints`
- `cmd/scrutineer/main.go`, where the backfill is called during startup

### Current Behavior

The backfill executes a raw query and scans its results without checking the returned GORM error:

```go
gdb.Raw(`
        SELECT f.id, f.sub_path, f.cwe, f.location, f.title, s.skill_name
        FROM findings f JOIN scans s ON s.id = f.scan_id
        WHERE f.fingerprint IS NULL OR f.fingerprint = ''
`).Scan(&rows)
```

It then updates every selected finding without checking whether any update failed:

```go
gdb.Model(&Finding{}).Where("id = ?", r.ID).Updates(map[string]any{
        "fingerprint":       fp,
        "last_seen_scan_id": gorm.Expr("COALESCE(NULLIF(last_seen_scan_id, 0), scan_id)"),
        "last_seen_commit":  gorm.Expr(`COALESCE(NULLIF(last_seen_commit, ''), "commit")`),
        "seen_count":        gorm.Expr("CASE WHEN seen_count = 0 THEN 1 ELSE seen_count END"),
})
```

The function has no error return and its startup caller invokes it unconditionally without any way to determine whether the backfill completed successfully.

### Impact

If the initial query fails, the function continues with an empty or incomplete result set and startup proceeds as if the backfill succeeded. If an update fails partway through the loop, earlier rows may be updated while later rows remain without fingerprints. Those findings cannot reliably participate in fingerprint-based deduplication during the current process, which can result in duplicate findings or incomplete last-seen metadata.

The incomplete state is not necessarily permanent because the backfill runs again on subsequent startups. However, there is no retry, warning, or failure signal during the affected startup and recurring database errors can leave the rows unpopulated indefinitely.

### Proposed Resolution

Change `BackfillFindingFingerprints` to return an error, check the error from the initial query and check every update result. Run the backfill in a transaction if partial completion is not acceptable. The startup caller should handle the returned error explicitly, either by failing startup with context or by logging a prominent warning according to the project's migration policy.

### Acceptance Criteria

- `BackfillFindingFingerprints` returns an error when its selection query fails.
- An update failure is returned with the affected finding ID in the error context.
- Startup no longer silently treats a failed fingerprint backfill as successful.
- Tests inject failures into both the query and update paths and verify that the errors reach the caller.
- Existing successful and idempotent backfill behavior remains unchanged.

abhinavgautam01:
Fixes #915

## Summary

Makes the finding fingerprint backfill fail explicitly and atomically when database operations cannot be completed.

Startup now receives and returns fingerprint backfill failures instead of continuing with partially or entirely unpopulated fingerprint data.

## Changes

- Change `BackfillFindingFingerprints` to return an error.
- Check and return errors from the initial finding selection query.
- Include the affected finding ID in update-error context.
- Run selection and updates in one transaction so a later failure rolls back earlier updates.
- Process findings in deterministic ID order.
- Propagate backfill failures through application startup.
- Preserve successful and idempotent backfill behavior.
- Add tests covering:
  - Successful fingerprint and last-seen metadata backfill.
  - Idempotent repeated execution.
  - Selection-query failure propagation.
  - Update failure propagation with finding context.
  - Transaction rollback after a partial update attempt.

## Tests

- `go mod tidy -diff`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `git diff --check`

### DIFF
--- a/cmd/scrutineer/main.go
+++ b/cmd/scrutineer/main.go
@@ -617,7 +617,9 @@ func run(log *slog.Logger) error {
 	}
 	db.BackfillFindings(gdb)
 	db.BackfillFindingRepository(gdb)
-	db.BackfillFindingFingerprints(gdb)
+	if err := db.BackfillFindingFingerprints(gdb); err != nil {
+		return fmt.Errorf("backfill finding fingerprints: %w", err)
+	}
 	db.BackfillStatusPriority(gdb)
 	worker.BackfillRepoDiskUsage(gdb, f.dataDir)
 	if err := db.SeedDefaultLabels(gdb); err != nil {
--- a/internal/db/fingerprint.go
+++ b/internal/db/fingerprint.go
@@ -3,6 +3,7 @@ package db
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"strings"
 
 	"gorm.io/gorm"
@@ -66,8 +67,9 @@ func normaliseLocation(loc string) string {
 // BackfillFindingFingerprints fills Finding.Fingerprint for rows created
 // before the column existed, joining through Scan for skill_name. It does
 // not merge existing duplicates; it just sets the column so future scans
-// dedupe against them. Safe to call on every startup.
-func BackfillFindingFingerprints(gdb *gorm.DB) {
+// dedupe against them. The transaction prevents a failed row update from
+// leaving a partially completed backfill. Safe to call on every startup.
+func BackfillFindingFingerprints(gdb *gorm.DB) error {
 	type row struct {
 		ID        uint
 		SubPath   string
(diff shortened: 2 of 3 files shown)


andrew (changes requested):
The error return and `.Error` checks are the right fix for #915; the transaction wrapper is not.

**`internal/db/fingerprint.go:81`** — wrapping the backfill in `gdb.Transaction` makes recovery worse. The operation is already idempotent (`WHERE fingerprint IS NULL OR ''` skips completed rows on the next run), so partial completion is the correct failure mode: rows before the failing one stay backfilled, the returned error names the bad finding ID, the operator fixes that row, and the next startup finishes the remainder. With the transaction, a persistent failure on any single row rolls back every row already processed, and since `main.go` now aborts on the error, the instance can't start at all — every restart re-attempts the full set and hits the same row. #915 made the transaction conditional on partial completion being unacceptable; here it's acceptable. Please drop the `Transaction` wrapper and keep the error checks and `error` return.

**`internal/db/fingerprint_test.go:139`** — with the transaction gone, `TestBackfillFindingFingerprintsRollsBackOnUpdateError` becomes a partial-completion test: assert `findings[0]` is backfilled, `findings[1]` isn't, and the error names `findings[1].ID`. Rename accordingly.

**`internal/db/fingerprint.go:87`** — `ORDER BY f.id` only exists so the test can predict which row fails second. Fine to keep for deterministic error output, but if it stays please note in the comment that it's for test determinism, not correctness.

abhinavgautam01:
Thanks, Fixed. The backfill now preserves successful earlier updates, returns the failing finding ID and retries only unfinished rows on the next run. The test now covers partial completion and retry behavior and the ordering comment clarifies its deterministic purpose. All checks pass.

PR #974, opened 2026-09-02
Gate the patch on a pristine checkout

p-linnane:
The patch gate ran host Git inside the per-scan workspace, which is bind-mounted read-write into the runner as the host uid. A repository that gets code execution during a scan can append to `src/.git/config`, and the `git reset --hard HEAD` at the top of `gitApplyCheck` then runs a `filter.<name>.smudge` command as the operator when it re-materialises a path bound by `.gitattributes`. No Git flag ignores local repository config, so hardening the invocation does not help: the gate has to run somewhere the runner never had write access.

Validation now happens in a temporary copy staged from the host-only clone cache under its per-URL lock. Local repositories are cloned with `--no-local` rather than copied, so a linked worktree's `gitdir:` pointer cannot lead Git back to the operator's real index and discard their staged work. The HEAD used for the `base_commit` comparison now comes from that pristine copy too, rather than from the runner's checkout where both sides of the comparison were attacker-controlled.

Codex was used to implement this, Opus 5 to review it.

### DIFF
--- a/internal/worker/patch_gate.go
+++ b/internal/worker/patch_gate.go
@@ -3,6 +3,7 @@ package worker
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -31,7 +32,7 @@ type patchReport struct {
 // on pass, appends an immutable RemediationAttempt and updates the
 // Finding.SuggestedFix projection through WriteFindingField. A gate failure is
 // not a scan error: the scan completed and the diff remains in Scan.Report.
-func (w *Worker) parsePatchOutput(scan *db.Scan, report string, emit func(Event)) error {
+func (w *Worker) parsePatchOutput(ctx context.Context, scan *db.Scan, report string, emit func(Event)) error {
 	var rep patchReport
 	if err := json.Unmarshal([]byte(report), &rep); err != nil {
 		return fmt.Errorf("parse patch: %w", err)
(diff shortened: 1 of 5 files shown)


andrew (changes requested):
The security fix is right: host git never touches the runner-writable tree again, the cache `.git/config` is host-populated so a filter cannot land there, and `--no-local` gives a local clone independent metadata so `reset --hard` cannot reach the operator's index through a `gitdir:` pointer. The filter-injection test exercises the real attack.

One regression to sort before merge:

`gatePatch` now validates against the current cache HEAD rather than the commit the scan ran on. `clone.EnsureWithOptions` does `fetch` + `reset --hard FETCH_HEAD` in the shared cache on every scan start, so a second scan on the same URL that begins while this scan's skill is still running moves the tree `stagePatchGateSrc` later copies. A valid patch then fails the `base_commit` check at patch_gate.go:72 and, if the touched file drifted upstream, `git apply --check` too. The old code was pinned to the workspace snapshot (attacker-writable, hence this PR); the new code is host-owned but time-varying.

`scan.Commit` is already the host-recorded pin, written from `gitHead` before the runner launches (skill.go:186/198, claude.go:215). Compare `rep.BaseCommit` against `scan.Commit` directly, and reset the staged tree to `scan.Commit` instead of `HEAD` in `gitApplyCheck` so applicability is checked against what the skill actually saw. The object is still in the cache store after a fetch. The same applies to the local-repo branch, where the operator may have committed between scan start and gate.

Smaller, while you're in there:

- `stagePatchSkillEdits` and its three call sites populate `scanWorkRoot(scan)/src`, which nothing in `parsePatchOutput` reads any more. `TestParsePatchOutput_resumedScanUsesLineageWorkspace` now tests a `scanWorkRoot`/`workRoot` distinction the gate no longer touches. Drop the helper and its call sites, and either remove the resumed-scan test or cut it to the cache-path assertion.
- `git clone --no-local` of a linked worktree detached at a commit no branch reaches yields an empty checkout, which surfaces as `diff targets missing file`. Resetting to `scan.Commit` per the fix above turns that into a clear `unknown revision` instead, so no separate change needed.

andrew (approved):
Round-2 addresses everything: base_commit compared against scan.Commit, staged tree reset to the scan's commit before the file-exists and apply checks, EnsureCommit deepens a shallow cache first, local-repo path pins the same way. stagePatchSkillEdits and the resumed-scan test are gone. TestPatchGate_acceptsPatchAfterCacheAdvances covers the concurrent-scan case directly.

PR #979, opened 2026-09-03
Fix OpenCode provider egress under Podman

### LINKED ISSUE #963: OpenCode: Egress whitelist ignored
With the latest Scrutineer version **v2026.08.25.1** and Podman as runtime, when using a custom OpenCode provider, the initial _scan_ task for a repository fails because the OpenCode container cannot connect to the provider:

```
$ cp -r /mnt/data/workspace ./src
2 agent-directive item(s) stripped from ./src
27275 file(s) excluded by path filters
profile: building scrutineer-profile-c-cpp:c64259a3fe39 (first build can take several minutes)
profile: built scrutineer-profile-c-cpp:c64259a3fe39 in 21s
profile: c-cpp (scrutineer-profile-c-cpp:c64259a3fe39)
profile guide: /home/user/scrutineer/data/bundled-profiles/f7c891a06e602783ae4341e360e76a3cba48ec19c1e7cb6ad25a6da2ec2a3dc9/c-cpp/PROFILE.md -> /work/AGENTS.md
[error] OpenCode provider "leo" readiness failed while reaching configured egress host "leo.example.com": exit status 1: curl: (56) CONNECT tunnel failed, response 403
scrutineer-readiness-fail: leo.example.com
```

This is my `scrutineer.yaml`:
```yaml
runtime: podman
backend: opencode
models:
- name: Qwen3.8 27B
  id: leo/Qwen/Qwen3.8-27B-FP8
  tier: high
opencode:
  providers:
    leo:
      config_file: ./opencode/leo/opencode/opencode.jsonc
      state_dir: ./opencode/leo/
      egress_allow:
      - leo.example.com
```

`opencode/leo/opencode/opencode.jsonc`:
```
{
  "$schema": "https://opencode.ai/config.json",
  "provider": {
    "leo": {
      "name": "LEO",
      "npm": "@ai-sdk/openai-compatible",
      "models": {
        "cyankiwi/GLM-4.7-Flash-AWQ-4bit": {
          "name": "GLM-4.7-Flash",
          "family": "glm-flash",
          "release_date": "2026-01-19",
          "attachment": false,
          "reasoning": true,
          "temperature": true,
          "tool_call": true,
          "limit": {
            "context": 200000,
            "output": 131072
          },
          "modalities": {
            "input": ["text"],
            "output": ["text"]
          }
        },
        "RedHatAI/gemma-4-31B-it-FP8-Dynamic": {
          "name": "Gemma 4 31B IT",
          "family": "gemma",
          "release_date": "2026-04-02",
          "attachment": true,
          "reasoning": true,
          "temperature": true,
          "tool_call": true,
          "limit": {
            "context": 262144,
            "output": 32768
          },
          "modalities": {
            "input": ["text", "image"],
            "output": ["text"]
          }
        },
        "Qwen/Qwen3.8-27B-FP8": {
          "name": "Qwen3.8 27B",
          "family": "qwen",
          "release_date": "2026-08-14",
          "attachment": true,
          "reasoning": true,
          "temperature": true,
          "tool_call": true,
          "limit": {
            "context": 262144,
            "output": 32768
          },
          "modalities": {
            "input": ["text", "image", "video"],
            "output": ["text"]
          }
        }
      },
      "options": {
        "baseURL": "https://leo.example.com/api/v1"
      }
    }
  }
}
```

`opencode/leo/opencode/auth.json`:
```
{
  "leo": {
    "type": "api",
    "key": "<redacted>"
  }
}
```

abhinavgautam01:
Fixes #963

## Summary

Fixes configured OpenCode providers failing hardened egress checks under Podman because Podman automatically inherited host proxy environment variables.

Scrutineer now disables Podman's automatic host proxy injection and explicitly supplies both uppercase and lowercase forms of its controlled proxy variables. This prevents inherited lowercase variables from overriding the per-scan egress proxy.

## Changes

- Pass `--http-proxy=false` to Podman container runs to prevent automatic host proxy inheritance.
- Set both uppercase and lowercase variants of:
  - `HTTPS_PROXY`
  - `HTTP_PROXY`
  - `ALL_PROXY`
- Clear both `NO_PROXY` and `no_proxy` for proxied scan containers.
- Preserve Docker behavior.
- Extend runtime and container tests to cover:
  - Podman proxy-inheritance suppression.
  - Uppercase and lowercase controlled proxy variables.
  - Fail-closed networking behavior.
  - Configured OpenCode provider run arguments.

## Verification

- `go test ./...`
- `go test -tags podman ./internal/worker -count=1`
- `go vet ./...`
- `go vet -tags podman ./...`
- `golangci-lint run`
- `git diff --check`

Live verification was also completed with Podman 6.1.1 using a rootless ARM64 Linux VM. The hardened proxy-sidecar integration, host-gateway resolution, keep-ID ownership and egress-blocking tests all passed.

### DIFF
--- a/internal/worker/container.go
+++ b/internal/worker/container.go
@@ -982,16 +982,16 @@ func (d ContainerRunner) sidecarNetworkIP(name, network string) (string, error) 
 // an unreachable host API lingers long enough for verifyHardenedNetwork to
 // capture its logs.
 func (d ContainerRunner) proxySidecarRunArgs(name, network string) []string {
-	args := []string{
-		"run", "-d",
+	args := runtimeRunArgs(d.Runtime,
+		"-d",
 		"--name", name,
 		"--network", network,
 		"--cap-drop", "ALL",
 		"--security-opt", "no-new-privileges",
 		"--read-only",
 		"--tmpfs", "/tmp:rw,noexec,nosuid,size=16m",
-		"--add-host", HostGatewayAlias + ":" + d.Egress.GatewayIP,
-	}
+		"--add-host", HostGatewayAlias+":"+d.Egress.GatewayIP,
+	)
 	for _, e := range EgressSidecarEnv(d.Egress, SidecarListenFirstIface+":"+proxySidecarPort) {
 		args = append(args, "-e", e)
 	}
@@ -1073,8 +1073,9 @@ func VerifyProxyBinary(ctx context.Context, rt ContainerRuntime, image string) e
 	if image == "" || !imageExistsLocally(ctx, rt, image) {
 		return nil
 	}
-	out, err := exec.CommandContext(ctx, runtimeBin(rt), "run", "--rm", "--pull", "never",
-		"--", image, "scrutineer", "proxy", "-h").CombinedOutput()
+	args := runtimeRunArgs(rt, "--rm", "--pull", "never",
+		"--", image, "scrutineer", "proxy", "-h")
+	out, err := exec.CommandContext(ctx, runtimeBin(rt), args...).CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("runner image %q is missing the scrutineer binary required for the "+
 			"hardened egress proxy sidecar (rebuild it from Dockerfile.runner): %w: %s",
--- a/internal/worker/container.go
+++ b/internal/worker/container.go
@@ -1158,7 +1158,7 @@ func (d ContainerRunner) verifyProxySidecarReachable(hn hardenedNet, image strin
 	deadline := time.Now().Add(proxySidecarReadyTimeout)
 	var last string
 	for {
-		out, err := exec.Command(runtimeBin(d.Runtime), sidecarReachArgs(hn.name, hn.proxyEndpoint, image)...).CombinedOutput()
+		out, err := exec.Command(runtimeBin(d.Runtime), sidecarReachArgs(d.Runtime, hn.name, hn.proxyEndpoint, image)...).CombinedOutput()
 		last = strings.TrimSpace(string(out))
 		if err == nil && strings.Contains(last, "REACHED") {
 			return nil
@@ -1233,13 +1233,13 @@ func hardenedProxyReachArgs(rt ContainerRuntime, network, gatewayIP, proxyPort, 
 // answers, e.g. 407 without auth) means the in-network path to the sidecar is
 // open, which by the sidecar's readiness gate also means the host API is
 // reachable through it.
-func sidecarReachArgs(network, endpoint, image string) []string {
+func sidecarReachArgs(rt ContainerRuntime, network, endpoint, image string) []string {
 	target := "http://" + endpoint + "/"
 	script := "curl -s -m 5 -o /dev/null " + target + " && echo REACHED || echo UNREACHABLE"
-	return []string{
-		"run", "--rm", "--cap-drop", "ALL", "--network", network,
+	return runtimeRunArgs(rt,
+		"--rm", "--cap-drop", "ALL", "--network", network,
 		"--entrypoint", "sh", "--", image, "-c", script,
-	}
+	)
 }
 
 // proxyPortFromURL extracts the port from a proxy URL of the shape ProxyURL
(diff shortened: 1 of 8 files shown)


internal/worker/container.go
```diff
@@ -513,12 +513,19 @@ func (d ContainerRunner) buildRunArgsForProvider(absWork, image string, hnet har
...
-			"-e", "HTTPS_PROXY="+proxyURL,
-			"-e", "HTTP_PROXY="+proxyURL,
-			"-e", "ALL_PROXY="+proxyURL,
-			"-e", "NO_PROXY=",
-		)
+		// Set both cases. Podman normally inherits both variants from its
+		// host, and curl prefers lowercase https_proxy over HTTPS_PROXY.
+		// runtimeRunArgs disables that Podman inheritance, while these explicit
+		// assignments also override image-provided proxy variables on every
+		// supported runtime.
```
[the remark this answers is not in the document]
abhinavgautam01: The proxy sidecar and proxy-binary check now use the shared runtime argument builder, so Podman consistently receives `--http-proxy=false`. I also added regression coverage and verified the sidecar path with a live rootless Podman integration run.

internal/worker/container.go
```diff
@@ -1066,8 +1073,9 @@ func VerifyProxyBinary(ctx context.Context, rt ContainerRuntime, image string) e
 	if image == "" || !imageExistsLocally(ctx, rt, image) {
 		return nil
 	}
-	out, err := exec.CommandContext(ctx, runtimeBin(rt), "run", "--rm", "--pull", "never",
-		"--", image, "scrutineer", "proxy", "-h").CombinedOutput()
+	args := runtimeRunArgs(rt, "--rm", "--pull", "never",
```
alexandre-daubois, line 1076: `sidecarReachArgs` (`internal/worker/container.go:1240`) is the one production run path left building a literal `"run"` argv, so the sidecar reach probe still inherits the host proxy variables that every other probe now suppresses. On a host exporting `http_proxy`, its `curl` dials that inherited proxy instead of `<sidecar-ip>:3128`, the address is unreachable from the `--internal` network, and `verifyProxySidecarReachable` burns the whole `proxySidecarReadyTimeout` before refusing the scan. You could thread the runtime into it and build through `runtimeRunArgs` like its two sibling probes.

internal/worker/container.go
```diff
@@ -513,12 +513,19 @@ func (d ContainerRunner) buildRunArgsForProvider(absWork, image string, hnet har
...
-			"-e", "ALL_PROXY="+proxyURL,
-			"-e", "NO_PROXY=",
-		)
+		// Set both cases. Podman normally inherits both variants from its
+		// host, and curl prefers lowercase https_proxy over HTTPS_PROXY.
+		// runtimeRunArgs disables that Podman inheritance, while these explicit
+		// assignments also override image-provided proxy variables on every
+		// supported runtime.
+		for _, key := range []string{
+			"HTTPS_PROXY", "https_proxy",
```
alexandre-daubois, line 522: These lowercase names become sandbox-owned, but `reservedOpencodePassEnv` (`internal/config/config.go:256`) still lists only the uppercase ones. An operator writing `pass_env: [https_proxy]` on an OpenCode provider therefore passes config validation and then has the value silently overridden by the `-e https_proxy=` assignment here, instead of getting the `is managed by scrutineer` error at startup. It may be worth adding `https_proxy`, `http_proxy`, `all_proxy` and `no_proxy` to that map so the reserved set matches what the run args write.

internal/worker/container.go
abhinavgautam01: The sidecar reachability probe now uses runtimeRunArgs, so Podman receives --http-proxy=false. I also verified the complete sidecar path with invalid host proxy variables in a live rootless Podman run.

internal/worker/container.go
abhinavgautam01: Thanks, fixed
Lowercase https_proxy, http_proxy, all_proxy and no_proxy are now reserved alongside their uppercase variants, with coverage confirming that OpenCode pass_env rejects all managed proxy names.

alexandre-daubois:
Thank you!

PR #996, opened 2026-09-05
Add coverage completeness filtering to the Scans page

abhinavgautam01:
Part of #771

## Summary

Adds coverage completeness indicators and filtering to the Scans page, allowing operators to distinguish execution status from how much of the intended scope was covered.

Uses the existing indexed `Scan.Completeness` field. Legacy scans with empty or NULL values display as `unknown` and are included in the unknown filter.

## Changes

- Add a completeness column and `complete`, `partial` and `unknown` filter options.
- Combine completeness with existing skill and status filters.
- Preserve the selection through sorting, pagination and live refresh.
- Apply the completeness filter to “Retry all failed” and preserve it on redirect.
- Return HTTP 400 for invalid completeness filter values.
- Show “No matching scans” when filters produce an empty result.
- Allow the toolbar to wrap and the table to scroll horizontally on narrow screens.
- Document filter behavior and add regression tests.

This change exposes existing worker-derived verdicts; it does not change coverage reconciliation or establish completeness for full and focus-area scans.

## Validation

- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `git diff --check`
- Local HTTP checks confirmed valid filters return 200, invalid values return 400 and live-refresh URLs preserve the selection.

### DIFF
--- a/docs/diff-based-rescans.md
+++ b/docs/diff-based-rescans.md
@@ -69,6 +69,10 @@ other files that affect how untrusted input reaches security-sensitive code.
 
 ## Coverage Metadata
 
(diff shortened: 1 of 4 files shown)


p-linnane (approved):
Looks good. Thanks @abhinavgautam01!

PR #1002, opened 2026-09-06
feat: reporting page for scan activity and cost averages

dkw-oss:
# feat: reporting page for scan activity and cost averages

Adds `/reporting`, plus a sidebar entry: repositories scanned, scans started,
scans completed and findings raised over a rolling **All time / Month / Week /
Day** window, with an optional **minimum-severity floor** on the findings.

Alongside the totals it shows the per-scan cost and token averages from
`docs/cost_averages.sql` for the selected period next to the all-time figures.
Those averages are aggregated in SQL through GORM, transcribing that file's
`WHERE status = 'done' AND cost_usd > 0` population directly; day bucketing
stays in Go because `date()`/`date_trunc` are dialect-specific and
`internal/db` keeps the schema portable.

The severity floor reuses `db.SeverityOrderSQL()`, so it can never disagree
with how the finding lists rank severity. It applies to findings only — scan
counts and cost averages are properties of scans — and is carried through the
period links and both download links.

The snapshot downloads from `/reporting/report.csv` and
`/reporting/report.json`. The CSV is a single rectangular table (one header,
uniform columns, no blank separator) with a `row_type` column marking the
period total, the all-time averages and the daily rows, so it opens as one
clean sheet in Excel. The JSON is named for what it holds: `period`,
`filters`, `activity_in_period`, `cost_averages_per_scan.in_period` /
`.all_time`, and `activity_by_day`. Both carry aggregate counts only — no
repository names, finding titles or API tokens.

Overlap with `/usage` is limited to the averages, which use a different
population (`done` + `failed` there, completed-and-costed here). All three
routes are read-only GETs on the existing browser mux and add no auth surface.

### DIFF
--- a/internal/web/reporting.go
+++ b/internal/web/reporting.go
@@ -124,6 +124,31 @@ type reportTotals struct {
 	Findings     int
 }
 
+// ScansPerRepo is the mean number of scan runs each scanned repository
+// accounted for. Display-only: a reader seeing 510 runs against 49
+// repositories needs the ratio to know the figure is a fan-out, not an
+// inflated count. One repository scan enqueues a run per skill (see
+// enqueueDiffRescanGroup), so this is normally well above 1.
+//
+// Deliberately a method rather than a field: it is derived presentation,
+// and the exports build their payloads from the fields explicitly, so
+// keeping it off the struct stops it leaking into the CSV or JSON.
+func (t reportTotals) ScansPerRepo() float64 {
+	if t.ReposScanned == 0 {
+		return 0
+	}
+	return float64(t.Scans) / float64(t.ReposScanned)
+}
+
+// CompletionRate is the share of scan runs that reached "done", as a
+// 0..1 fraction for the pct template helper.
+func (t reportTotals) CompletionRate() float64 {
+	if t.Scans == 0 {
+		return 0
+	}
+	return float64(t.ScansDone) / float64(t.Scans)
+}
+
 // reportAverages is one column of the cost-averages table: the per-scan
 // means from docs/cost_averages.sql. Runs is the denominator, exposed so a
 // small-sample average is recognisable as one.
--- a/internal/web/reporting.go
+++ b/internal/web/reporting.go
@@ -359,10 +359,8 @@ func (s *Server) buildReport(iv reportInterval, minSeverity string) reportData {
 		}
 		acc := at(sc.FinishedAt.UTC().Format(reportDateLayout))
 		active(acc, sc.RepositoryID)
-		if sc.Status == db.ScanDone {
-			data.Totals.ScansCompleted++
-			acc.completed++
-		}
+data.Totals.ScansCompleted++
+acc.completed++
 		// Spend lands on the finish day for any terminal status: the worker
 		// writes the cost and token columns when a run finalises, so an
 		// in-flight run has nothing to attribute yet and a failed one still
--- a/internal/web/reporting.go
+++ b/internal/web/reporting.go
@@ -34,6 +34,12 @@ import (
 // are UTC so a report generated in two timezones bucket-aligns.
 const reportDateLayout = "2006-01-02"
 
+// findingsField is the "findings" column name, shared by the CSV header,
+// the CSV summary block and the JSON payload. Named rather than repeated
+// so the three spellings of the export cannot drift apart, and so this
+// package's count of the bare literal stays under goconst's threshold.
+const findingsField = "findings"
+
 // reportInterval is one selectable rolling window. Dur of 0 means all
 // time. Rolling rather than calendar: the totals are a running figure, so
 // they should not collapse to near-zero at midnight or on the 1st.
@@ -301,7 +307,7 @@ func (s *Server) reportingCSV(w http.ResponseWriter, r *http.Request) {
 		{"repos_scanned", strconv.Itoa(data.Totals.ReposScanned)},
 		{"scans", strconv.Itoa(data.Totals.Scans)},
 		{"scans_done", strconv.Itoa(data.Totals.ScansDone)},
-		{"findings", strconv.Itoa(data.Totals.Findings)},
+		{findingsField, strconv.Itoa(data.Totals.Findings)},
 	}
 	summary = append(summary, averageCSVRows("window", data.Window)...)
 	summary = append(summary, averageCSVRows("alltime", data.AllTime)...)
@@ -314,7 +320,7 @@ func (s *Server) reportingCSV(w http.ResponseWriter, r *http.Request) {
 	cw.Flush()
 	_, _ = w.Write([]byte("\n"))
 
-	_ = cw.Write([]string{"date", "repos_scanned", "scans", "findings", "cost_usd", "total_tokens"})
+	_ = cw.Write([]string{"date", "repos_scanned", "scans", findingsField, "cost_usd", "total_tokens"})
 	for _, d := range data.Days {
 		_ = cw.Write([]string{
 			d.Date,
@@ -350,7 +356,7 @@ func (s *Server) reportingJSON(w http.ResponseWriter, r *http.Request) {
 			"date":          d.Date,
 			"repos_scanned": d.ReposScanned,
 			"scans":         d.Scans,
-			"findings":      d.Findings,
+			findingsField:   d.Findings,
 			"cost_usd":      d.CostUSD,
 			"total_tokens":  d.TotalTokens,
 		})
@@ -364,7 +370,7 @@ func (s *Server) reportingJSON(w http.ResponseWriter, r *http.Request) {
 			"repos_scanned": data.Totals.ReposScanned,
 			"scans":         data.Totals.Scans,
 			"scans_done":    data.Totals.ScansDone,
-			"findings":      data.Totals.Findings,
+			findingsField:   data.Totals.Findings,
 		},
 		"averages": map[string]any{
 			"window":   averageJSON(data.Window),
--- a/internal/web/reporting.go
+++ b/internal/web/reporting.go
@@ -115,21 +115,27 @@ func minSeverityRank(level string) (int, bool) {
 }
 
 // reportTotals is the running-total panel. ReposScanned counts distinct
-// repositories with at least one scan in the window, so a repo rescanned
-// ten times still counts once. Findings honours the severity filter; the
-// scan counts do not, since severity is a property of findings only.
+// repositories with at least one run that began or ended in the window, so
+// a repo rescanned ten times still counts once. ScansStarted and
+// ScansCompleted are read on their own clocks (see reportActivitySQL), so
+// they are not a total and a subset of it. Findings honours the severity
+// filter; the scan counts do not, since severity is a property of findings
+// only.
 type reportTotals struct {
-	ReposScanned int
-	Scans        int
-	ScansDone    int
-	Findings     int
+	ReposScanned   int
+	ScansStarted   int
+	ScansCompleted int
+	Findings       int
 }
 
-// ScansPerRepo is the mean number of scan runs each scanned repository
+// ScansPerRepo is the mean number of runs each repository with activity
 // accounted for. Display-only: a reader seeing 510 runs against 49
 // repositories needs the ratio to know the figure is a fan-out, not an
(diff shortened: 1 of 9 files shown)


internal/web/reporting.go
```diff
@@ -0,0 +1,587 @@
...
+	}
+
+	// Every status counts toward the activity totals, so unlike the
+	// averages this read is not narrowed to completed scans.
+	var scans []db.Scan
+	sq := s.DB.Model(&db.Scan{}).
+		Select("repository_id", "status", "cost_usd", "input_tokens", "output_tokens",
+			"cache_read_tokens", "cache_write_tokens", "finished_at", "created_at")
+	if data.Since != nil {
+		sq = sq.Where(reportAnchorSQL+" >= ?", *data.Since)
```
p-linnane: This uses finish/create time for `scans_started` and counts queued rows as started. Please derive starts from non-null `started_at` and completions from `finished_at`.

internal/web/templates/reporting.html
```diff
@@ -0,0 +1,167 @@
...
+          <td class="text-right">{{usd .CostUSD}}</td>
+          <td class="text-right text-muted-foreground">{{bignum .TotalTokens}}</td>
+        </tr>
+      {{end}}
+      </tbody>
+    </table>
+  </div>
+{{else}}
+  {{template "empty" (dict "Icon" "calendar-off" "Title" "No activity in this window" "Body" "Widen the interval, or run a scan to start populating the report.")}}
+{{end}}
```
p-linnane, line 174: This page never invokes `{{template "foot" .}}`, so shared dialogs, toaster, and closing markup are omitted. Please add it after this block.

p-linnane (changes requested):
Two correctness issues need fixing before merge.

dkw-oss:
> Two correctness issues need fixing before merge.

Good pickups, ty, actioned

internal/web/reporting.go
```diff
@@ -0,0 +1,647 @@
...
+		}
+		if !data.inWindow(sc.FinishedAt) {
+			continue
+		}
+		acc := at(sc.FinishedAt.UTC().Format(reportDateLayout))
+		active(acc, sc.RepositoryID)
+		if sc.Status == db.ScanDone {
+			data.Totals.ScansCompleted++
+			acc.completed++
+		}
```
p-linnane, line 380: Final nit.

```suggestion
data.Totals.ScansCompleted++
acc.completed++
```

p-linnane: This is already validated immediately above.

internal/web/reporting.go
```diff
@@ -0,0 +1,645 @@
...
+			data.Totals.ScansStarted++
+			acc.started++
+			active(acc, sc.RepositoryID)
+		}
+		if !data.inWindow(sc.FinishedAt) {
+			continue
+		}
+		acc := at(sc.FinishedAt.UTC().Format(reportDateLayout))
+		active(acc, sc.RepositoryID)
+data.Totals.ScansCompleted++
```
alexandre-daubois: This counts every run that carries a `finished_at` as a completion, so a `failed`, `cancelled` or `skipped` row inflates `ScansCompleted` and pushes `CompletionRate` to 100% of runs started. Five tests in `reporting_test.go` fail on this head (`ScansCompleted = 2, want 1` for `day`), and the JSON export's `measured_by` string still says completions are counted "on runs that reached done". Restoring the `sc.Status == db.ScanDone` guard around both increments is what those tests and that export text expect.

internal/web/reporting.go
```diff
@@ -0,0 +1,645 @@
...
+acc.completed++
```
alexandre-daubois: These two statements sit at column 0 instead of the loop body's indentation, which `gofmt -l internal/web/` reports. `.golangci.yml` enables `gofmt` as a formatter and `tests.yml` runs `golangci-lint`, so the lint job fails on this file. A `gofmt -w internal/web/reporting.go` after restoring the status guard clears it.

internal/web/reporting.go
```diff
@@ -0,0 +1,645 @@
...
+	// begun or ended, which is every row that has anything to attribute.
+	var scans []db.Scan
+	sq := s.DB.Model(&db.Scan{}).
+		Select("repository_id", "status", "cost_usd", "input_tokens", "output_tokens",
+			"cache_read_tokens", "cache_write_tokens", "started_at", "finished_at").
+		Where(reportActivitySQL)
+	if data.Since != nil {
+		sq = sq.Where(reportWindowSQL, *data.Since, *data.Since)
+	}
+	sq.Find(&scans)
```
alexandre-daubois: Neither this `Find` nor the `Scan` calls at 269 and 396 read their error, so a locked or failing database produces a report of zeros rather than a failure. An operator can then download that CSV or JSON and file the zeros as a real activity figure. Returning the error to the handler and rendering a failure would keep an empty report from passing as a true one.

internal/web/reporting.go
```diff
@@ -0,0 +1,645 @@
...
+			return canonical
+		}
+	}
+	return ""
+}
+
+// minSeverityRank converts a canonical severity to its severityOrder rank.
+// severityOrder ranks the most severe LOWEST (Critical 0 … Low 3, unknown
+// last), so "at least this severe" is `rank <= threshold`.
+func minSeverityRank(level string) (int, bool) {
```
alexandre-daubois: This re-derives `len(db.SeverityLevels) - 1 - i`, the same rank `db.SeverityOrderSQL` builds into its `CASE`, in a second package. The two agree today only because `TestMinSeverityRankMatchesSeverityOrder` hardcodes 0 and 3, so a change to the `ELSE` slot or to the list order leaves this filter selecting the wrong end of the scale. Exporting a `db.SeverityRank` that `SeverityOrderSQL` also consumes would tie them together.

internal/web/templates/layout.html
```diff
@@ -57,6 +57,7 @@
           <li><a href="/audit"{{if eq $nav "audit"}} aria-current="page"{{end}}><i data-lucide="clipboard-check"></i><span>Audit</span></a></li>
           <li><a href="/skills"{{if eq $nav "skills"}} aria-current="page"{{end}}><i data-lucide="sparkles"></i><span>Skills</span></a></li>
           <li><a href="/usage"{{if eq $nav "usage"}} aria-current="page"{{end}}><i data-lucide="coins"></i><span>Usage</span></a></li>
+          <li><a href="/reporting"{{if eq $nav "reporting"}} aria-current="page"{{end}}><i data-lucide="file-chart-column"></i><span>Reporting</span></a></li>
```
alexandre-daubois, line 60: This adds a sidebar section that `README.md`'s own sidebar list never gains, next to its `- **Usage**` entry. `CLAUDE.md` keeps that list in review scope: "A new UI surface does still belong in the `README.md` list that covers it (the repo tab list, the sidebar sections)". A one-line `- **Reporting**` entry beside `**Usage**` would match.

internal/web/templates/reporting.html
```diff
@@ -0,0 +1,171 @@
...
+  <div class="overflow-x-auto">
+    <table class="table w-full">
+      <thead><tr>
+        <th>Date</th>
+        <th class="text-right">Repositories</th>
+        <th class="text-right">Started</th>
+        <th class="text-right">Completed</th>
+        <th class="text-right">Findings{{if $r.MinSeverity}} ({{$r.MinSeverity}}+){{end}}</th>
+        <th class="text-right">Cost</th>
+        <th class="text-right">Total tokens</th>
```
alexandre-daubois, line 155: The daily table stops at `Total tokens`, while both exports carry `scans_averaged`, `avg_cost_usd` and `avg_total_tokens` per day. `reporting.go`'s header comment states that "a downloaded report always matches the screen it was downloaded from", which three export-only columns break. Adding them to this row, or dropping them from the exports, would restore that.

internal/web/reporting.go
dkw-oss: @p-linnane your change broke the tests, have a look at 6610d4e77e7503f3840e4e4e2dff0e23759867a6 as a more explicit test + resolution and let me know what you think

dkw-oss: cc @alexandre-daubois

dkw-oss:
@alexandre-daubois - thank you

> This counts every run that carries a finished_at as a completion, so a
> failed, cancelled or skipped row inflates ScansCompleted and pushes
> CompletionRate to 100% of runs started. Five tests in reporting_test.go fail
> on this head (ScansCompleted = 2, want 1 for day), and the JSON export's
> measured_by string still says completions are counted "on runs that reached
> done". Restoring the sc.Status == db.ScanDone guard around both increments is
> what those tests and that export text expect.

Fixed in 6610d4e — guard restored with a comment, plus
TestBuildReportCountsOnlyDoneRunsAsCompleted so a re-suggestion fails with an
explanation rather than five opaque count mismatches. Worth adding that paused
also stamps finished_at, so without the guard a run that has not finished at
all counts as a completion now and again when it does.

> These two statements sit at column 0 instead of the loop body's indentation,
> which gofmt -l internal/web/ reports. .golangci.yml enables gofmt as a
> formatter and tests.yml runs golangci-lint, so the lint job fails on this
> file. A gofmt -w internal/web/reporting.go after restoring the status guard
> clears it.

Fixed with the same change; gofmt -l is clean across the tree.

> Neither this Find nor the Scan calls at 269 and 396 read their error, so a
> locked or failing database produces a report of zeros rather than a failure.
> An operator can then download that CSV or JSON and file the zeros as a real
> activity figure. Returning the error to the handler and rendering a failure
> would keep an empty report from passing as a true one.

Fixed: reportAveragesFor and buildReport now return errors and all three
handlers 500 before writing anything, with TestReportingFailsRatherThanReportingZeros
dropping the scans table to prove it. Note errcheck could not have caught this
— gorm returns *gorm.DB with the error in a field.

> This re-derives len(db.SeverityLevels) - 1 - i, the same rank
> db.SeverityOrderSQL builds into its CASE, in a second package. The two agree
> today only because TestMinSeverityRankMatchesSeverityOrder hardcodes 0 and 3,
> so a change to the ELSE slot or to the list order leaves this filter selecting
> the wrong end of the scale. Exporting a db.SeverityRank that SeverityOrderSQL
> also consumes would tie them together.

Fixed: exported db.SeverityRank, SeverityOrderSQL now numbers its WHEN and ELSE
clauses from it, and TestSeverityOrderSQLIsNumberedBySeverityRank parses the
generated CASE instead of pinning literals. The criticism of the old test was
right — it pinned two numbers, not the relationship.

> This adds a sidebar section that README.md's own sidebar list never gains,
> next to its - **Usage** entry. CLAUDE.md keeps that list in review scope: "A
> new UI surface does still belong in the README.md list that covers it (the
> repo tab list, the sidebar sections)". A one-line - **Reporting** entry beside
> **Usage** would match.

Fixed: added

internal/web/reporting.go
dkw-oss: Please see https://github.com/alpha-omega-security/scrutineer/pull/1002#issuecomment-5567381468

internal/web/reporting.go
dkw-oss: Please see https://github.com/alpha-omega-security/scrutineer/pull/1002#issuecomment-5567381468

internal/web/reporting.go
dkw-oss: Please see https://github.com/alpha-omega-security/scrutineer/pull/1002#issuecomment-5567381468

internal/web/reporting.go
dkw-oss: Please see https://github.com/alpha-omega-security/scrutineer/pull/1002#issuecomment-5567381468

internal/web/templates/layout.html
dkw-oss: Please see https://github.com/alpha-omega-security/scrutineer/pull/1002#issuecomment-5567381468

internal/web/templates/reporting.html
dkw-oss: 5f9be69a3f117084da3461c3e35fd79076270ffb

p-linnane (approved):
Looks great. Thanks @dkw-oss!

Issue #727, opened 2026-07-27
Occasionally schedule an audit that ignores the threat model


andrew:
`triage` enqueues audits based on what `recon` and `threat-model` identified as in scope. If the threat model rules something out too early, no later scan ever looks at it.

On each `triage` run, with 25-50% probability, enqueue one extra `security-deep-dive` scan that ignores the plan:

- adversarial sweep: pick a component the threat model marked safe or out of scope and audit it anyway
- random dig: pick a random source directory, pass a minimal open-ended question, and give it no threat-model context

One or the other, not both, and only when there's already at least one planned audit for the repo (so a repo with nothing to scan doesn't get only a random dig). Findings go through the normal `revalidate` gate, so noise costs a revalidate run rather than analyst time.

Cheap to add (`internal/worker` picks the extra job when building the triage queue) and the cost is one deep-dive per triage at most.

Source: [google/mantis](https://github.com/google/mantis) `mantis-plan/SKILL.md` "Exploratory Investigations" (Apache-2.0).

Related: #20 (huge projects; this is the opposite problem but shares the "where to look" machinery).

- [x] random dig (#1004)
- [ ] adversarial sweep

PR #1004, opened 2026-09-06
Add independent random-dig audits to eligible triage runs

abhinavgautam01:
Part of #727

## Summary

Adds independent random-dig audits alongside planned security audits to investigate code without relying on the threat model’s assumptions.

This implements the random-dig option from #727. It does not add adversarial sweeps targeting components explicitly marked safe or out of scope.

## Changes

- Select approximately one third of eligible triage invocations using a stable hash of the triage scan ID.
- Schedule at most one automatic extra `security-deep-dive` after a successful full-mode threat model has queued a planned audit.
- Preserve normal planned audits and exclude manual threat-model runs, failed threat models and diff rescans from automatic exploration.
- Select a source directory from the filtered checkout, respecting repository exclusions and subproject scope.
- Stage independent audit instructions without threat-model, recon, or scan-config guidance.
- Restrict the exploratory scan’s callback token to validating its own report schema.
- Preserve the normal findings schema, ingestion, deduplication and High/Critical revalidation pipeline.
- Record exploration mode, selected directory and triage lineage; preserve these inputs on explicit retries.
- Prevent exploratory scans from becoming diff baselines.
- Document the behavior and expose exploration metadata in scan API responses.

## Safeguards

- Completion-hook redelivery does not reroll selection or enqueue duplicate extras, including after failure or cancellation.
- Missing source candidates, unavailable instructions and retry targets that are no longer eligible fail visibly before running the audit.
- Existing operator path exclusions remain authoritative.
- Reports must describe their limited audit scope rather than claim whole-repository coverage.

## Tests

- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `git diff --check`


internal/worker/exploration.go
```diff
@@ -0,0 +1,160 @@
...
+		return fmt.Errorf("read random-dig instructions: %w", err)
+	}
+	// The loaded skill is scan-local. Update it too so the logged prompt and
+	// any report-repair invocation describe the instructions actually staged.
+	skill.Body = string(body)
+	if err := stageSkill(skill, workRoot, skillDir); err != nil {
+		return err
+	}
+	// Keep repository identity and scope, not any model-derived guidance.
+	// apiAuth limits this scan's token to validating its own output schema.
```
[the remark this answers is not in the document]
abhinavgautam01: Clarified that staging omits model-derived guidance while retaining the token; the web server’s apiAuth middleware separately enforces the token restriction.

internal/worker/exploration_test.go
```diff
@@ -0,0 +1,226 @@
...
+		{name: "root", skill: deepDiveSkillName, mode: ExplorationRandomDig, target: ".", valid: true},
+		{name: "subdir", skill: deepDiveSkillName, mode: ExplorationRandomDig, target: "lib/parser", valid: true},
+		{name: "wrong skill", skill: "verify", mode: ExplorationRandomDig},
+		{name: "unknown mode", skill: deepDiveSkillName, mode: "other"},
+		{name: "orphan path", skill: deepDiveSkillName, target: "lib"},
+		{name: "escape", skill: deepDiveSkillName, mode: ExplorationRandomDig, target: "../lib"},
+		{name: "absolute", skill: deepDiveSkillName, mode: ExplorationRandomDig, target: "/lib"},
+		{name: "focus", skill: deepDiveSkillName, mode: ExplorationRandomDig, focus: "{}"},
+		{name: "diff", skill: deepDiveSkillName, mode: ExplorationRandomDig, rescan: db.ScanRescanModeDiff},
+		{name: "finding", skill: deepDiveSkillName, mode: ExplorationRandomDig, finding: new(uint(1))},
```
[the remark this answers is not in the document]
abhinavgautam01: Both usages are valid under the repository’s Go 1.27 requirement. The validation and recipe tests pass on Go 1.27.1, including race detection, so I’ve retained them.

internal/web/server_test.go
```diff
@@ -4310,6 +4310,15 @@ func TestRetry_preservesScanFields(t *testing.T) {
 				t.Errorf("retry lost focus area: %q", f.FocusArea)
 			}
 		}},
+		{"exploration", func(sc *db.Scan) {
+			sc.TriageScanID = new(uint(17))
```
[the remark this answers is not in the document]
abhinavgautam01: This is valid since Go 1.26 introduced new(expression). This repository requires Go 1.27 and the retry tests pass with Go 1.27.1, including race detection. No change needed.

internal/web/scans.go
```diff
@@ -473,6 +476,9 @@ func (s *Server) scansRetryFailed(w http.ResponseWriter, r *http.Request) {
 			DiffBaseScanID:       sc.DiffBaseScanID,
 			ScanGroup:            sc.ScanGroup,
 			FocusArea:            sc.FocusArea,
+			TriageScanID:         sc.TriageScanID,
+			ExplorationMode:      sc.ExplorationMode,
```
alexandre-daubois, line 480: The bulk retry rebuilds `ScanOpts` from a row whose `Select` list never gained `triage_scan_id`, `exploration_mode` or `exploration_path`, so all three read back as zero values here. A retried random-dig audit then returns as an ordinary unscoped `security-deep-dive` with the threat model staged and the whole callback API open, which is what `ValidateExploration` and the new `apiAuth` gate exist to prevent. `CLAUDE.md` covers this shape directly: "A new `ScanOpts` / `Scan` field must be threaded through every enqueue path, retries included, and added to each retry query's `Select`".

internal/web/api.go
```diff
@@ -78,6 +78,10 @@ func (s *Server) apiAuth(next http.Handler) http.Handler {
 			writeAPIError(w, http.StatusInternalServerError, err.Error())
 			return
 		}
+		if scan.ExplorationMode != "" && (r.Method != http.MethodPost || r.URL.Path != fmt.Sprintf("/scans/%d/validate-report", scan.ID)) {
+			writeAPIError(w, http.StatusForbidden, "exploratory audits may only validate their own report")
```
alexandre-daubois, line 82: This gate lives in `apiAuth`, which wraps the `/api/` mux alone, so the `/api/v1/` export mux (`GET /findings`, `GET /scans`, `GET /audit/queue`, `POST /import`) never sees it. Those routes gate on a loopback `Host` only, and `apiBaseFor` hands a `host_skills` deep dive the loopback `HostAPIBase`, so a random-dig audit on the host runner can read every finding and prior scan report that `docs/exploratory-audits.md` says are withheld. It may be worth refusing an exploratory token in `exportHandler` as well, or recording `host_skills` as incompatible with `random-dig`.

internal/worker/exploration.go
```diff
@@ -0,0 +1,161 @@
...
+
+func exploratorySourceFile(name string) bool {
+	return slices.Contains(exploratorySourceExtensions, strings.ToLower(path.Ext(name)))
+}
+
+func stageExploratoryWorkspace(workRoot, skillDir, apiBase string, scan *db.Scan, skill *db.Skill) error {
+	if err := ValidateExploration(skill.Name, scan.ExplorationMode, scan.ExplorationPath, scan.FocusArea, scan.RescanMode, scan.FindingID); err != nil {
+		return err
+	}
+	if scan.ExplorationPath == "" || skill.SourcePath == "" {
```
alexandre-daubois: Staging requires `references/random-dig.md` under `skill.SourcePath`, which is empty for a `security-deep-dive` authored on `/skills` and absent from any `-skills` or `skills_repo` override that does not ship the file. `autoEnqueueExploratoryAudit` picks the skill on `name` and `active` alone, so on those instances a third of triage runs queue a scan that always dies at staging. Checking for the reference pack before the enqueue would stop the audit being scheduled there at all.

internal/worker/exploration.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+			if clean != "" && name != "." && name != clean && !strings.HasPrefix(clean, name+"/") && !strings.HasPrefix(name, clean+"/") {
+				return fs.SkipDir
+			}
+			return nil
+		}
+		if !entry.Type().IsRegular() || (clean != "" && !strings.HasPrefix(name, clean+"/")) {
+			return nil
+		}
+		if exploratorySourceFile(name) {
```
alexandre-daubois, line 150: Candidates are every directory holding one of `exploratorySourceExtensions`, and `skills.BuiltinSkipPaths` covers `node_modules` and `dist` but not `vendor`, `third_party`, `testdata`, `tests` or `examples`. A repository with many fixture and vendored directories spends most of its max-tier random digs outside first-party source, which `SKILL.md` names as the only valid target. Excluding those trees from the candidate set, or weighting the pick by first-party file count, would put the sample where the audit is meant to look.

docs/skills.md
```diff
@@ -37,7 +37,7 @@ These live in `skills/` and are embedded in the Scrutineer executable. At startu
...
 | `threat-model` | Derives the project's security contract from source and docs: components, entry-point trust table, claimed and disclaimed properties, and disposition labels. Loaded by `security-deep-dive` so it does not re-derive boundaries per run. |
-| `security-deep-dive` | The model-driven audit. Inventories trust boundaries and sinks, then runs a six-step trace/boundary/validate/prior-art/reach/rate analysis on each. After threat-model completes, Scrutineer fans configured focus areas out into parallel deep-dive scans. |
+| `security-deep-dive` | The model-driven audit. Inventories trust boundaries and sinks, then runs a six-step trace/boundary/validate/prior-art/reach/rate analysis on each. After threat-model completes, Scrutineer fans configured focus areas out into parallel deep-dive scans. One third of eligible triage runs also receive one independent [random-dig audit](exploratory-audits.md). |
```
alexandre-daubois, line 40: This row documents the random-dig mode, but the `context.json` section below lists `scrutineer.exploration` neither in its sample nor in the paragraph that enumerates every key from `focus_area` through `scan_config`. A skill author reading that section finds no record of the key `stageContextWithInputs` now writes. `CLAUDE.md` asks for the `context.json` paragraph to move with a schema change, so adding `exploration` to the sample and that paragraph would close it.

internal/web/exploratory_audit_test.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func TestExploratoryAuditFanoutIdempotent(t *testing.T) {
+	s, done := newTestServer(t)
+	defer done()
+	parent, skill := exploratoryFixture(t, s)
+	s.autoEnqueueFocusAreaDeepDives(&parent)
+	var wg sync.WaitGroup
+	for range 6 {
+		wg.Go(func() { s.autoEnqueueFocusAreaDeepDives(&parent) })
```
alexandre-daubois: This spawns six goroutines against the same fan-out and asserts once they finish, which passes on an unguarded build most runs. `CLAUDE.md` asks a concurrency test to force the interleaving instead, and names the seam: a registered GORM `Create` callback that commits a competing exploratory row inside the transaction under test fails every run on the unguarded code. It may be worth reaching for that seam here.

internal/worker/exploration.go
```diff
@@ -0,0 +1,161 @@
...
+	"strings"
+
+	"scrutineer/internal/db"
+)
+
+const ExplorationRandomDig = "random-dig"
+
+// ValidateExploration rejects combinations that would silently turn a blind
+// source audit back into a planned, finding-scoped, or diff-only audit.
+func ValidateExploration(skill, mode, target, focus, rescan string, findingID *uint) error {
```
alexandre-daubois: Five adjacent `string` parameters carry `skill`, `mode`, `target`, `focus` and `rescan`, so a transposed pair at any of the three call sites compiles and then validates the wrong field. Two of those callers already hold the `*db.Scan` every value comes from. Taking the scan plus the skill name instead would remove the ordering hazard.

abhinavgautam01: Changed ValidateExploration to accept *db.Scan and the skill name, updating all callers and validation tests.

internal/web/exploratory_audit_test.go
abhinavgautam01: Replaced the goroutine-only check with a forced competing insert using a GORM callback. Added a transaction-level duplicate recheck and verified rollback without another queue job.

docs/skills.md
abhinavgautam01: Added scrutineer.exploration to the sample and documented its fields, omitted guidance, callback restrictions and runtime requirements.

internal/worker/exploration.go
abhinavgautam01: Excluded vendor, third-party, test, fixture and example trees from directory sampling at any depth. Added root-level, nested and case-insensitive exclusion tests.

internal/worker/exploration.go
abhinavgautam01: Added instruction-pack validation before enqueueing. Skills without a nonempty references/random-dig.md retain planned audits but do not queue exploratory scans. Added regression coverage.

internal/web/api.go
abhinavgautam01: Random-dig now rejects host execution at enqueue and worker preparation, including host_skills. Blocking bearer tokens alone would not prevent unauthenticated loopback access. Added tests and documented the restriction.

internal/web/scans.go
abhinavgautam01: Fixed the retry query’s selected columns and added a bulk-retry regression test covering triage lineage, exploration mode, target directory and scan group.

alexandre-daubois (approved):

PR #1009, opened 2026-09-08
fix(web): stop the repository list reading every cached ecosyste.ms payload

p-linnane:
`GET /` and `/repositories` timed out once repository rows carried large cached ecosyste.ms payloads. `repoList` hydrated whole `db.Repository` rows and the default `updated_at desc` order had no index, so SQLite dragged every row's `metadata` and six ecosyste.ms blobs through a temporary sort: 404ms for the list query on a 600 repository, 528 MiB fixture.

The list now selects only the six columns `repo_list.html` renders, and `updated_at` is indexed, which plans the default order as an index scan and takes that query to 1.5ms. Orders keyed on `disk_bytes` or a scan subquery still sort the table, so this covers the landing page rather than every column header.

Opus 5 was used to implement this.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -141,7 +141,10 @@ type Repository struct {
 	NextScheduledScanAt *time.Time `gorm:"index"`
 
 	CreatedAt time.Time
-	UpdatedAt time.Time
+	// UpdatedAt drives the repository list's default newest-first order. Keep
+	// it indexed so SQLite can find the first page without scanning through
+	// every repository's large cache columns and building a temporary sort.
+	UpdatedAt time.Time `gorm:"index"`
 
 	Scans            []Scan            `gorm:"constraint:OnDelete:CASCADE"`
 	ExpectedFindings []ExpectedFinding `gorm:"constraint:OnDelete:CASCADE"`
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -743,6 +743,43 @@ func TestWithPragmas_joinsOnExistingQuery(t *testing.T) {
 	}
 }
 
+func TestOpen_addsRepositoryUpdatedAtIndexToExistingDatabase(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "existing.db")
+	gdb, err := Open(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	sqldb, err := gdb.DB()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() { _ = sqldb.Close() }()
+	if err := gdb.Migrator().DropIndex(&Repository{}, "idx_repositories_updated_at"); err != nil {
+		t.Fatal(err)
+	}
+	// Prove the precondition: without this the reopen below could pass on an
+	// index the drop never removed, and the test would assert nothing.
+	if gdb.Migrator().HasIndex(&Repository{}, "idx_repositories_updated_at") {
+		t.Fatal("DropIndex left idx_repositories_updated_at in place")
+	}
+	if err := sqldb.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	reopened, err := Open(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	reopenedSQL, err := reopened.DB()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() { _ = reopenedSQL.Close() }()
+	if !reopened.Migrator().HasIndex(&Repository{}, "idx_repositories_updated_at") {
+		t.Fatal("Open did not add idx_repositories_updated_at to an existing database")
+	}
+}
+
 // foreign_keys and busy_timeout are per-connection in SQLite. Setting them
 // via a single gdb.Exec only configures whichever pooled connection that
 // Exec lands on, leaving the rest at the defaults (#457). Open now folds
--- a/internal/web/list_performance_test.go
+++ b/internal/web/list_performance_test.go
@@ -77,13 +77,14 @@ func BenchmarkListPagesLargeDataset(b *testing.B) {
 	s, done := newTestServer(b)
 	defer done()
 	seedListPagePerfFixtures(b, s, 500)
+	handler := s.Handler()
 
 	for _, path := range []string{"/", "/findings", "/scans"} {
 		b.Run(path, func(b *testing.B) {
 			b.ReportAllocs()
 			for i := 0; i < b.N; i++ {
 				w := httptest.NewRecorder()
-				s.Handler().ServeHTTP(w, localReq("GET", path))
+				handler.ServeHTTP(w, localReq("GET", path))
 				if w.Code != 200 {
 					b.Fatalf("%s status %d: %s", path, w.Code, w.Body)
 				}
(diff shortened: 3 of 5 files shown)


internal/web/server.go
```diff
@@ -847,6 +847,12 @@ type repoRow struct {
 	Branches []string
 }
 
+// repoListColumns is the complete Repository surface rendered by
+// repo_list.html. Repository also carries multiple potentially large cached
+// JSON payloads; projecting the row keeps the default updated_at sort from
+// copying those blobs through SQLite's temporary sort table.
+const repoListColumns = "id, url, languages, health, clone_error, disk_bytes"
```
alexandre-daubois: Nothing ties this string to `repo_list.html`, so a field added to the template renders its zero value and every test still passes. Swapping `Select` for an `Omit` of `metadata`, `threat_model`, `scan_config` and the six `ecosystems_*_data` columns keeps the blobs out of the row while a newly rendered column arrives on its own.

internal/web/list_performance_test.go
```diff
@@ -141,6 +340,13 @@ func countDBQueries(t testing.TB, s *Server) func() int64 {
 	if err := s.DB.Callback().Query().Before("gorm:query").Register(name+":query", callback); err != nil {
 		t.Fatal(err)
 	}
+	// GORM routes statements through three processors: Find/First/Pluck use
+	// Query, Scan (including Raw(...).Scan) uses Row, and Exec uses Raw.
+	// Register on all three so the bound counts every statement a handler
+	// issues rather than whichever processor it happens to reach for.
+	if err := s.DB.Callback().Row().Before("gorm:row").Register(name+":row", callback); err != nil {
```
alexandre-daubois: Widening the counter to the `Row` processor takes `/` from 4 statements to 8 against a `max` of 8, and `/scans` from 5 to 6 against a `max` of 6, so both pages sit on their ceiling. Raising those two thresholds in `TestListPagesQueryCountsStayBounded` would keep the bound meaningful for the next handler that needs one more statement.

internal/web/server.go
```diff
@@ -928,7 +934,7 @@ func (s *Server) repoList(w http.ResponseWriter, r *http.Request) {
 	page := paginate(r, total)
 
 	var repos []db.Repository
-	q.Limit(perPage).Offset((page.N - 1) * perPage).Find(&repos)
+	q.Select(repoListColumns).Limit(perPage).Offset((page.N - 1) * perPage).Find(&repos)
```
alexandre-daubois: `loadRepoMap` runs the same unprojected `Find(&repos)` for `/findings`, `/advisories`, `/orgs/{owner}` and `/maintainers/{login}`, and those templates read only `.ID` and `.Name` off the map. On the fixture shape this PR benchmarks, `/findings` costs 74.6 MB and 37ms per render against 0.25 MB and 0.57ms for the fixed repository list, so it may be worth projecting in that helper too.

internal/web/server.go
```diff
@@ -847,6 +847,12 @@ type repoRow struct {
 	Branches []string
 }
 
+// repoListColumns is the complete Repository surface rendered by
```
alexandre-daubois: The index in the sibling commit removes the temporary sort from the default order, and `TestRepoList_defaultSortUsesUpdatedAtIndex` asserts `USE TEMP B-TREE FOR ORDER BY` is gone from that plan, so the mechanism named here no longer applies to the order it names. What the projection buys on that path is the 20 result rows no longer carrying the blobs, 9.4ms and 73 MB down to 0.57ms and 0.25 MB, which you could state instead.

internal/web/list_performance_test.go
```diff
@@ -116,6 +116,205 @@ func TestRepoList_diskBadgeReadsCachedColumn(t *testing.T) {
...
+		// or a live database.
+		go func() {
+			defer close(done)
+			for i := 0; ; i++ {
+				select {
+				case <-stop:
+					return
+				default:
+				}
+				_ = s.DB.Model(&db.Repository{}).
```
alexandre-daubois: The loop has no pause between updates, so `with-enrichment-writes` measures the read path against a 256 KiB write storm rather than the single background refresh the comment above describes, and the discarded `.Error` lets a failing UPDATE spin at full speed. A short sleep between iterations, plus recording the first error for the cleanup to surface, would make the number reproducible.

internal/db/db_test.go
```diff
@@ -743,6 +743,42 @@ func TestWithPragmas_joinsOnExistingQuery(t *testing.T) {
 	}
 }
 
+func TestOpen_addsRepositoryUpdatedAtIndexToExistingDatabase(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "existing.db")
+	gdb, err := Open(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := gdb.Migrator().DropIndex(&Repository{}, "idx_repositories_updated_at"); err != nil {
```
alexandre-daubois, line 757: The first `Open` handle reaches `sqldb.Close()` only on the happy path, so any of the three `t.Fatal` calls before it leaves the database file open while `t.TempDir` cleanup runs. Deferring the close right after the first `Open` covers every exit.

internal/db/db.go
```diff
@@ -141,7 +141,10 @@ type Repository struct {
 	NextScheduledScanAt *time.Time `gorm:"index"`
 
 	CreatedAt time.Time
-	UpdatedAt time.Time
+	// UpdatedAt drives the repository list's default newest-first order. Keep
+	// it indexed so SQLite can find the first page without scanning through
+	// every repository's large cache columns and building a temporary sort.
+	UpdatedAt time.Time `gorm:"index:idx_repositories_updated_at"`
```
alexandre-daubois: `idx_repositories_updated_at` is the name GORM derives for this field on this table, and every other indexed field in `Repository` carries a bare `gorm:"index"`. Dropping the explicit name matches the rest of the struct and yields the same index the two new tests look for.

internal/web/list_performance_test.go
```diff
@@ -116,6 +116,205 @@ func TestRepoList_diskBadgeReadsCachedColumn(t *testing.T) {
...
+		if err := s.DB.Create(&repo).Error; err != nil {
+			b.Fatal(err)
+		}
+		repoIDs = append(repoIDs, repo.ID)
+	}
+
+	// Build the mux once, as main does. Re-registering ~200 routes per
+	// iteration costs ~200us and ~2200 allocations, which would otherwise
+	// swamp the allocation figure this benchmark exists to watch.
+	handler := s.Handler()
```
alexandre-daubois, line 283: `BenchmarkListPagesLargeDataset` in this file still calls `s.Handler()` inside its iteration loop, which is the cost this comment documents. Hoisting that call there as well would put both benchmarks on the same footing.

internal/web/list_performance_test.go
```diff
@@ -116,6 +116,205 @@ func TestRepoList_diskBadgeReadsCachedColumn(t *testing.T) {
...
+	run := func(b *testing.B) {
+		b.Helper()
+		b.ReportAllocs()
+		b.ReportMetric(float64(repositories*7*blobBytes)/(1<<20), "cache-MiB")
+		for i := 0; i < b.N; i++ {
+			w := httptest.NewRecorder()
+			handler.ServeHTTP(w, localReq("GET", "/"))
+			if w.Code != 200 {
```
alexandre-daubois, line 300: A regression that renders the empty-repositories state returns 200 and measures fast, so this would report an improvement on a page that stopped listing anything. Checking that one seeded row appears in the body once before the loop would pin what the numbers cover.

internal/web/server.go
p-linnane: Went with a narrow `repoListFields` struct as the query dest instead: GORM derives the SELECT from the type, so a template reference to anything outside it fails the render instead of rendering a zero. `Omit` would still hydrate a seventh `ecosystems_*_data` column the day it lands.

alexandre-daubois (approved):

PR #1011, opened 2026-09-08
Add operator feedback for verification reruns

### LINKED ISSUE #943: Rerun verify skill with feedback
On some codebases, I've seen many findings that are poorly verified, like running scripts that don't interact with the real codebase. I'd like an option to rerun the verification skill, ideally with user feedback

abhinavgautam01:
Fixes #943

## Summary

Allows operators to rerun finding verification with optional feedback, such as a missed entry point or an incorrect reproduction assumption, without changing the original finding.

## Changes

- Add an optional feedback textarea to the finding verification form, using the existing form styling.
- Store feedback on the individual verification scan and pass it into the verify skill's context.
- Require valid UTF-8 and limit feedback to 4,000 bytes.
- Restrict feedback to finding-scoped verify scans.
- Preserve feedback when retrying scans while keeping new runs independent.
- Display submitted feedback on the scan page and alongside its verification history entry.
- Include feedback in scan API responses, exports and execution recipes.
- Prevent duplicate verification submissions while a verification is pending.
- Instruct the verify skill to evaluate feedback against current evidence without overriding safety requirements or the grading rubric.
- Document the new field and add validation, retry, context-staging and UI regression tests.

## Validation

- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `git diff --check`
- Browser checks at desktop and mobile widths in light and dark themes, including focused and unfocused feedback styling.


andrew (changes requested):
The mechanism is right — per-scan column, staged into `context.json`, kept on retry, in the recipe, validated at each layer — and the SKILL.md wording on treating feedback as steer rather than evidence is good.

A few UI/wiring things to sort:

- **Two verify buttons on the finding page.** The new form in `finding_show.html` posts to `/findings/{id}/verify` right after the workflow section, which already has a Verify button (`finding_workflow.html:98`) posting to the same endpoint without feedback. Consolidate to one entry point — either drop the workflow button now that there is a dedicated section, or fold the textarea into the workflow form.
- **`scan_show.html` renders the feedback section before `{{template "crumbs"}}`**, so it appears above the breadcrumb bar. Move it below the crumbs into the page body alongside the other scan-detail blocks.
- **The new form has no `hx-post`**, so it does a full page reload where every other action on the page uses htmx. Add `hx-post="/findings/{{$f.ID}}/verify" hx-swap="none"`.
- **`VerificationFeedback` was inserted as the first field of `skillContextScrutineer`**, which reorders the `scrutineer` block in every `context.json`. Move it next to `FindingID`/`DependentID` where the docs example places it.
- The `agentEnqueueMu` you added around `findingVerify` closes the check-then-enqueue race, but `findingCritic`/`findingDisclose`/`findingReattackRun` have the same `skipOpen=true` pattern without it. Taking the lock inside `runFindingSkill` when `skipOpen` covers all of them and removes the special case in `findingVerify`.

abhinavgautam01:
Thanks, Addressed all five points: verification now has one feedback-enabled htmx form, scan feedback follows the breadcrumbs, the context field sits beside FindingID and skipOpen locking is shared across all affected actions. Added UI, htmx and concurrent-enqueue regression tests.

andrew (approved):
Thanks — all five round-1 points addressed.

PR #1017, opened 2026-09-10
feat(codex): run scans on a ChatGPT account credential

p-linnane:
Codex can authenticate with a ChatGPT subscription, but every scan gets a fresh private `CODEX_HOME` where the interactive login cannot be completed and no account credential survives, so `CODEX_API_KEY` was the only credential that reached a scan. Mounting the operator's whole Codex home would share thread databases and history with untrusted scan containers, and mounting one rotating credential without coordination would let concurrent token refreshes race.

`codex.auth_file` now names a dedicated `auth.json` in the config file only, and just that credential is mounted read-write into each scan's otherwise private Codex home. Scrutineer requires mode `0600` and a ChatGPT refresh token, revalidates before every run, and refuses to start alongside `CODEX_API_KEY` or `OPENAI_API_KEY` so an account run cannot silently spend Platform credits; a cancellation-aware semaphore and a queue-level admission cap keep refreshes from racing. threatmodel.md records what stays exposed, since a read-write mount lets a hostile scan replace a credential later scans depend on.


internal/web/theme.go
```diff
@@ -323,15 +323,30 @@ func (s *Server) settingsUpdateConcurrency(w http.ResponseWriter, r *http.Reques
...
+}
+
 // settingsRestartRunner rebuilds the runner at the saved concurrency, applying
 // it live. In-flight scans are cancelled by the swap; queued scans survive.
 func (s *Server) settingsRestartRunner(w http.ResponseWriter, r *http.Request) {
 	n := db.SettingInt(s.DB, db.SettingConcurrency)
 	if n <= 0 {
 		n = s.Queue.Concurrency()
 	}
+	if eff := s.Queue.EffectiveConcurrency(n); eff == s.Queue.Concurrency() {
```
alexandre-daubois, line 343: This returns before `Reconfigure`, so a restart whose saved value already equals the applied one rebuilds nothing, and on an install where `SettingConcurrency` was never written `n` falls back to `s.Queue.Concurrency()` and the button stops restarting anything at all. The banner it answers still promises to cancel the running scans, and an operator clicking it to rebuild a wedged runner reads `Runner unchanged`. Gating on the cap itself, along the lines of `eff < n && eff == s.Queue.Concurrency()`, would keep the account-auth no-op without disarming the plain restart.

internal/web/settings_handlers_test.go
```diff
@@ -191,6 +191,53 @@ func TestSettingsRestartRunner(t *testing.T) {
...
+
+	if err := db.SetSetting(s.DB, db.SettingConcurrency, "16"); err != nil {
+		t.Fatal(err)
+	}
+	w := postForm(t, s, "/settings/runner/restart", nil)
+	if w.Code != http.StatusSeeOther {
+		t.Fatalf("status %d: %s", w.Code, w.Body)
+	}
+	if got := s.Queue.Concurrency(); got != 1 {
+		t.Errorf("runner concurrency = %d, want account-auth cap 1 after restart", got)
```
alexandre-daubois, line 207: `SetMaxConcurrency(1)` already drives `Concurrency()` to 1 before the request runs, so this holds whether the handler capped the restart or returned without touching the runner. Reconfiguring above the cap first, or asserting on the `Runner unchanged` flash, would separate those two outcomes.

README.md
```diff
@@ -405,7 +405,20 @@ Scrutineer can drive OpenAI's [codex](https://github.com/openai/codex) CLI inste
...
+credits. Create an isolated file-backed login with `codex login --device-auth`,
+then configure its credential file:
+
+    backend: codex
+    codex:
+      auth_file: ~/.config/scrutineer/codex-rubygems/auth.json
+
+The credential must be mode `0600`. Scrutineer refuses this configuration while
+`CODEX_API_KEY` or `OPENAI_API_KEY` is set, mounts only `auth.json` into each
+scan's otherwise private Codex home, and serializes account-authenticated scans
```
alexandre-daubois, line 418: `codex.auth_file` pins the runner at one slot and discards a configured `-concurrency`, but the `-concurrency` row at `README.md:355` still describes a free choice with a 1.5x chat ceiling. Noting the cap on that row would match how every other concurrency ceiling in this repo is documented.

README.md
```diff
@@ -405,7 +405,20 @@ Scrutineer can drive OpenAI's [codex](https://github.com/openai/codex) CLI inste
     export CODEX_API_KEY=sk-...
     go run ./cmd/scrutineer -skills ./skills -backend codex
 
-The container, egress proxy, language profiles and skill staging stay the same; only the agent CLI inside the container changes. The egress allowlist picks up `api.openai.com` automatically, and the model pick list defaults to codex's own catalog with tier tags already set -- override with `models:` in the config if you want a different set. Use `-model-base-url` or `model_base_url:` for a custom OpenAI-compatible endpoint; under codex it is passed as `openai_base_url` to `codex exec`. The codex backend requires the containerised runner; `--no-container` with `-backend codex` is rejected at startup.
+It can also use a ChatGPT subscription login without consuming Platform API
+credits. Create an isolated file-backed login with `codex login --device-auth`,
```
alexandre-daubois, line 409: Run as written, `codex login --device-auth` writes `~/.codex/auth.json` rather than the `~/.config/scrutineer/codex-rubygems/auth.json` that `auth_file` points at four lines below, so startup fails on `inspect codex.auth_file`. Carrying the `CODEX_HOME=...` form from `docs/codex.md` into this snippet would make the two halves agree.

internal/worker/container.go
```diff
@@ -322,37 +330,57 @@ func (d ContainerRunner) RunSkill(ctx context.Context, sj SkillJob, emit func(Ev
 		return result, err
 	}
 	runBase := d.buildRunArgsForProvider(absWork, image, hnet, absConfig, provider, "/work")
+	h := d.harness()
+	unlockCodexAuth := func() {}
```
alexandre-daubois, line 334: The anonymous function and the `terminalErr` relay buy nothing but dropping the credential lock before `readCappedReport`, a few milliseconds on a queue already pinned at one slot. `defer unlockCodexAuth()` straight after the acquire releases ahead of the earlier defers all the same and removes the extra scope and the error hand-off.

internal/queue/queue.go
```diff
@@ -80,6 +83,30 @@ func (q *Queue) Concurrency() int {
 	return int(q.concurrency.Load())
 }
 
+// EffectiveConcurrency reports the runner limit that would be applied for a
+// requested value after process-level caps. It does not rebuild the runner.
+func (q *Queue) EffectiveConcurrency(requested int) int {
```
alexandre-daubois, line 88: This forwards to `effectiveConcurrency` and adds nothing, while the unexported form keeps its one internal caller in `Reconfigure`. Exporting the single method and deleting the wrapper would stop the two names drifting apart.

PR #1018, opened 2026-09-10
fix(runner): install the codex code-mode host beside the CLI

p-linnane:
The pinned Codex release was installed from its standalone executable archive, which contains `codex` but not the version-matched `codex-code-mode-host` that the CLI spawns from the same directory, so every scan reported Code Mode as unavailable and ran on the degraded analysis path. Renovate followed those same assets from #1014, so no version update could supply the helper. Both binaries now come from the release package archive under one per-architecture digest lock, and the custom datasources follow it so the tag and both digests move together.

`CostFromUsage` prices `gpt-daybreak-blue-latest` as the GPT-5.6 Sol alias OpenAI documents it to be, so an approved operator who adds it to `models:` gets a real cost figure rather than zero. It stays out of the seeded defaults: Codex marks it `visibility: hide`, and the first entry of that list becomes the implicit default whenever `default_model` is unset.


internal/worker/pricing_test.go
```diff
@@ -19,3 +19,16 @@ func TestCostFromUsage_gpt6Astra(t *testing.T) {
...
+
+func TestCostFromUsage_daybreakBlueUsesGPT56SolPricing(t *testing.T) {
+	usage := Usage{
+		InputTokens:      1_000_000,
+		OutputTokens:     1_000_000,
+		CacheReadTokens:  100_000,
+		CacheWriteTokens: 200_000,
+	}
+	want := CostFromUsage(modelGPT56SolID, usage)
+	if got := CostFromUsage(modelDaybreakBlueID, usage); math.Abs(got-want) > 1e-9 {
```
alexandre-daubois, line 34: `harness.CostFromUsage` returns 0 for any model missing from its table, so this comparison passes when the alias and `gpt-5.6-sol` both price at zero, which is the exact outcome the change set out to remove. Asserting `want > 0` next to the equality would pin a real figure instead of an agreement between two blanks.

internal/worker/harness_test.go
```diff
@@ -169,7 +169,16 @@ func TestDefaultModelsFor_codexMatchesPinnedCatalog(t *testing.T) {
 		{Name: "GPT-5.5", ID: "gpt-5.5"},
 		{Name: "GPT-5.2", ID: "gpt-5.2"},
 	}
-	if got := DefaultModelsFor(CodexHarness{}); !reflect.DeepEqual(got, want) {
+	got := DefaultModelsFor(CodexHarness{})
+	if got[0].ID != modelGPT56SolID {
```
alexandre-daubois: `want` already fixes `gpt-5.6-sol` at index 0 and carries no Daybreak entry, so the `reflect.DeepEqual(got, want)` on line 181 subsumes both checks, and `got[0]` panics rather than fails on an empty list. Dropping the two loops keeps the coverage and the panic risk with it.

PR #1022, opened 2026-09-11
fix(db): keep retrying finding writes while the lock is held

p-linnane:
`WriteFindingField` reads then updates the row in one deferred transaction, and SQLite skips the busy handler when such a transaction takes the write lock, so `busy_timeout(5000)` never applied. The only protection was five attempts over 15 ms, so anything holding the lock longer failed the write: CI hit it in the concurrent history test, and a worker mid-transaction turned analyst edits into HTTP 422s.

`FindingWriteTransaction` now retries the whole transaction under a five-second, cancellation-aware budget matching the busy timeout, with jittered backoff for an active writer and an immediate retry for a stale snapshot. The browser, API, verification, remediation and VINCE transactions go through it too, since a nested call gets one savepoint-backed attempt; callbacks can run more than once, so they stay database-only and reset per-attempt state. Browser and API edits use the request context; VINCE persistence does not, because the submission has already succeeded.

### DIFF
--- a/internal/db/finding_helpers.go
+++ b/internal/db/finding_helpers.go
@@ -1,6 +1,7 @@
 package db
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"regexp"
@@ -8,6 +9,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"scrutineer/internal/retry"
 
 	"github.com/git-pkgs/vulns"
 	"gorm.io/gorm"
@@ -24,9 +27,11 @@ const GHSAIDPattern = `(?i)GHSA(-[0-9a-z]{4}){3}`
 var ghsaIDRE = regexp.MustCompile("^" + GHSAIDPattern + "$")
 
 const (
-	findingWriteMaxAttempts  = 5
+	findingWriteRetryTimeout = 5 * time.Second
+	findingWriteMaxDelay     = 100 * time.Millisecond
 	findingCapHistoryMaxRows = 20
 	sqliteBusyCode           = 5
+	sqliteBusySnapshotCode   = 517
 )
 
 var errFindingWriteConflict = errors.New("finding changed concurrently")
@@ -64,7 +69,7 @@ func WriteFindingField(gdb *gorm.DB, findingID uint, field, newValue string, sou
 	// sync must commit together: a failure between them would change the
 	// stored value with no matching history row (breaking the audit
 	// trail) or leave cvss_score inconsistent with cvss_vector.
-	return retryFindingWrite(gdb, findingID, func(tx *gorm.DB) error {
+	return FindingWriteTransaction(gdb, findingID, func(tx *gorm.DB) error {
 		var f Finding
 		if err := tx.First(&f, findingID).Error; err != nil {
 			return fmt.Errorf("load finding %d: %w", findingID, err)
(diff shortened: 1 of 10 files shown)


internal/db/finding_helpers.go
```diff
@@ -354,12 +370,20 @@ func conditionalFindingUpdate(gdb *gorm.DB, findingID uint, column string, oldVa
 	return nil
 }
 
-// SQLite reports a stale WAL read transaction as SQLITE_BUSY_SNAPSHOT (an
-// extended SQLITE_BUSY code) instead of returning zero rows from the compare-
-// and-swap. The whole owned transaction must be restarted to get a new snapshot.
-func isSQLiteBusy(err error) bool {
+func findingWriteRetryDelay(err error, attempt int) (time.Duration, bool) {
+	if errors.Is(err, errFindingWriteConflict) {
```
alexandre-daubois, line 374: A conflict returns a zero delay, so `retry.Sleep` falls straight through and the loop reruns the whole transaction with no pause for the full five-second budget, where the old loop backed off from 1 ms. Two writers trading a hot finding turn one edit into thousands of transactions against the same row, so routing this branch through `retry.BackoffDelay` the way the plain busy branch does would bound the churn.

internal/web/api_finding_writes.go
```diff
@@ -55,7 +55,7 @@ func (s *Server) apiPatchFinding(w http.ResponseWriter, r *http.Request) {
 		fields = append(fields, field)
 	}
 	sort.Strings(fields)
-	if err := s.DB.Transaction(func(tx *gorm.DB) error {
+	if err := db.FindingWriteTransaction(s.DB.WithContext(r.Context()), uint(id), func(tx *gorm.DB) error {
```
alexandre-daubois, line 58: A write that exhausts the five-second budget or loses its request context still comes back as `422` carrying `context deadline exceeded`, which tells a skill its field values are bad when the row was locked. Mapping the `context.DeadlineExceeded` and `context.Canceled` cases to `503` would let a client tell a retry from a rejection.

internal/web/finding_write_contention_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func testFindingEditWaitsForWriter(t *testing.T, api bool) {
+	t.Helper()
+	s, cleanup := newTestServer(t)
+	t.Cleanup(cleanup)
+	f, token, _ := seedFindingForAPI(t, s)
+	scopeAPITokenToFinding(t, s, token, f.ID)
+	ctx, cancel := context.WithCancel(t.Context())
+	t.Cleanup(cancel)
+	s.DB = s.DB.WithContext(ctx)
```
alexandre-daubois: Both handlers rebind the write to `r.Context()` and `httptest.NewRequest` hands them a background context, so `cancel()` reaches the reads alone and the cleanup sits on `<-done` until the retry budget expires whenever an assertion fails. Threading `ctx` into the request the helpers build would make the cleanup effective and cover the client-disconnect path the change describes.

internal/worker/skill_parsers.go
```diff
@@ -1202,7 +1202,8 @@ type verifyRecord struct {
 }
 
 func (w *Worker) recordVerifyOutput(scan *db.Scan, f db.Finding, record verifyRecord) error {
-	return w.DB.Transaction(func(tx *gorm.DB) error {
+	return db.FindingWriteTransaction(w.DB, f.ID, func(tx *gorm.DB) error {
```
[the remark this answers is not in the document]
p-linnane: Pre-existing and should be handled in a follow up.

PR #1023, opened 2026-09-11
Show "focus area" in scan lists and list filters

edorian:
Clicking a focus-area label opens /scans?group=<batch-id>, showing only scans from that batch.

The batch filter stays active when changing skill, status, completeness, or sorting. “Retry all failed” also stays scoped to that batch.

# Before

All on the same (omitted) repo, making the /scans overview quite samey and how to parse

Security-deep-dive

# After

I paused

### DIFF
--- a/internal/web/scans.go
+++ b/internal/web/scans.go
@@ -33,6 +33,10 @@ func (s *Server) jobs(w http.ResponseWriter, r *http.Request) {
 	if status != "" {
 		q = q.Where("status = ?", status)
 	}
+	group := r.URL.Query().Get("group")
+	if group != "" {
+		q = q.Where("scan_group = ?", group)
+	}
 
 	sortCol, dir := splitSort(r.URL.Query().Get("sort"))
 	switch sortCol {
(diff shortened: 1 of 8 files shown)


internal/web/scans.go
```diff
@@ -448,6 +452,7 @@ func (s *Server) scansRetryFailed(w http.ResponseWriter, r *http.Request) {
 			  AND COALESCE(n.sub_path, '') = COALESCE(scans.sub_path, '')
 			  AND COALESCE(n.ref, '') = COALESCE(scans.ref, '')
 			  AND COALESCE(n.finding_id, 0) = COALESCE(scans.finding_id, 0)
+			  AND COALESCE(n.focus_area, '') = COALESCE(scans.focus_area, '')
```
alexandre-daubois, line 455: This conjunct puts `focus_area` into the supersession key, so nothing but another unscoped run can supersede the failed unscoped deep-dive that `autoEnqueueFocusAreaDeepDives` enqueues when a threat-model fails, and a repository with `focus_areas` configured never produces one again. Bulk retry re-runs that row on every click, which is the monotonic growth this `NOT EXISTS` was written to stop, so it may be worth treating an empty `focus_area` on either side as a match.

andrew (changes requested):
Nice — the batch filter and per-area rows make the split runs much easier to read.

One behaviour change to sort in `scansRetryFailed`: adding `focus_area` to the supersession key means an unscoped fallback deep-dive (the one `autoEnqueueFocusAreaDeepDives` enqueues when threat-model fails) is no longer suppressed once later scoped runs exist. It is not unbounded growth — the retry itself creates a new `focus_area = ""` row that supersedes the old one, so each click still retries at most one per tuple — but it does mean bulk retry keeps re-running a stale unscoped scan the scoped set has already covered. Treating empty on either side as a match closes it:

```sql
AND (COALESCE(n.focus_area, "") = COALESCE(scans.focus_area, "")
     OR COALESCE(n.focus_area, "") = ""
     OR COALESCE(scans.focus_area, "") = "")
```

i.e. an unscoped run supersedes and is superseded by anything; scoped runs only supersede the same area.

Small one: the rewritten comment above that block dropped the "Cancelled is deliberately absent" line — that is the non-obvious part worth keeping.

edorian:
Thanks to both of you. Makes total sense but I didn't have the needed context/depth of understanding to see it.

I hope https://github.com/alpha-omega-security/scrutineer/pull/1023/changes/03e5d04843ac9a19682289d3b4ca1aa0f8c97de4 addresses this sufficiently.

andrew (approved):
The supersession clause and restored comment are exactly right, and the three-case test pins the behaviour.

PR #1024, opened 2026-09-11
Fix repository deletion blocked by paused scans

edorian:
Currently, a repository with paused scans cannot be deleted. The delete handler rejects the request, but the UI hides the error, making the button appear unresponsive. Paused scans cannot be cancelled through the UI either.

Delete the repository’s paused scans as part of the deletion transaction, and display any deletion failures on the page.

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -3037,9 +3037,14 @@ func (s *Server) deleteRepository(repo db.Repository) (deletedRepository, error)
 		if err := deleteFindingChildren(tx, repo.ID); err != nil {
 			return err
 		}
+		if err := tx.Where("sbom_upload_id IN (SELECT id FROM sbom_uploads WHERE repository_id = ?)", repo.ID).
+			Delete(&db.SBOMPackage{}).Error; err != nil {
+			return err
+		}
 		for _, child := range []any{
 			&db.Finding{}, &db.Scan{}, &db.Subproject{}, &db.Dependency{},
 			&db.Dependent{}, &db.Package{}, &db.Advisory{}, &db.SBOMUpload{},
+			&db.PackageAlternative{}, &db.ExpectedFinding{},
 		} {
 			if err := tx.Where("repository_id = ?", repo.ID).Delete(child).Error; err != nil {
 				return err
@@ -3161,6 +3166,8 @@ func deleteFindingChildren(tx *gorm.DB, repoID uint) error {
 	for _, child := range []any{
 		&db.FindingNote{}, &db.FindingCommunication{}, &db.FindingReference{},
 		&db.FindingHistory{}, &db.FindingDependent{}, &db.FindingReview{},
+		&db.FindingVerification{}, &db.FindingAttackPath{},
+		&db.RemediationValidation{}, &db.RemediationAttempt{},
 	} {
 		if err := tx.Where(findingsOfRepo, repoID).Delete(child).Error; err != nil {
 			return err
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -2971,11 +2971,14 @@ func (s *Server) repoDelete(w http.ResponseWriter, r *http.Request) {
 
 	deleted, err := s.deleteRepository(repo)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/web/server.go
```diff
@@ -3002,13 +3005,18 @@ func (s *Server) deleteRepository(repo db.Repository) (deletedRepository, error)
 		// the finding it points at, and any scan referencing a doomed finding
 		// must have its NO ACTION link cleared or the finding delete 787s.
 		var inFlight int64
+		// Paused scans owned by this repo are removed in this transaction.
```
alexandre-daubois, line 3008: A paused scan can still hold a live goqite entry, because `pauseQueuedOnAccountError` flips every `queued` row to `paused` while its job stays on the queue, and `wrap` drops that job as stale only after it loads the row. Deleting the repository removes the row first, so the redelivered job fails with `load scan N: record not found` and goqite retries it up to `MaxReceive` instead of dropping it, which treating `gorm.ErrRecordNotFound` in `wrap` as a stale job would close.

internal/web/server.go
```diff
@@ -2991,7 +2994,7 @@ type deletedRepository struct {
 	ConversationIDs []uint
 }
 
-var errRepositoryDeleteInFlight = errors.New("repository has queued, running, or paused scans")
+var errRepositoryDeleteInFlight = errors.New("repository has queued or running scans, or a linked paused scan on another repository")
```
alexandre-daubois, line 2997: The `409` description in `openapi.yaml` still reads `Repository has queued, running, or paused scans`, which no longer matches `DELETE /repositories/{id}` now that a paused scan owned by the repository is deleted rather than refused. A generated client reading the contract expects a conflict on a repository whose only in-flight work is paused and gets a `204`, so you may want to reword that line to cover queued, running and cross-repository paused scans.

edorian:
Thank you for the review and the insights @alexandre-daubois, much appreciated.

Hope this makes sense now

internal/web/server.go
```diff
@@ -3029,9 +3037,14 @@ func (s *Server) deleteRepository(repo db.Repository) (deletedRepository, error)
...
 			return err
 		}
+		if err := tx.Where("sbom_upload_id IN (SELECT id FROM sbom_uploads WHERE repository_id = ?)", repo.ID).
+			Delete(&db.SBOMPackage{}).Error; err != nil {
+			return err
+		}
 		for _, child := range []any{
 			&db.Finding{}, &db.Scan{}, &db.Subproject{}, &db.Dependency{},
 			&db.Dependent{}, &db.Package{}, &db.Advisory{}, &db.SBOMUpload{},
+			&db.PackageAlternative{}, &db.ExpectedFinding{},
```
alexandre-daubois, line 3047: `db.AdvisoryAudit` carries a `repository_id` with no `OnDelete:CASCADE` from `Repository` and is absent from this list, so every repository delete leaves its `advisory_audits` rows in place pointing at a row that is gone. Adding `&db.AdvisoryAudit{}` next to the two new entries closes the last `repository_id` table, and the assertion map in `TestRepoDelete_removesRepoAndAllLinkedData` would then cover it in both `foreign_keys` modes.

internal/web/server.go
```diff
@@ -3153,6 +3166,8 @@ func deleteFindingChildren(tx *gorm.DB, repoID uint) error {
 	for _, child := range []any{
 		&db.FindingNote{}, &db.FindingCommunication{}, &db.FindingReference{},
 		&db.FindingHistory{}, &db.FindingDependent{}, &db.FindingReview{},
+		&db.FindingVerification{}, &db.FindingAttackPath{},
```
alexandre-daubois, line 3169: The four tables added here are missing from `deleteFinding`'s own child loop, which still lists only notes, communications, references, history, dependents and reviews. `DELETE /findings/{id}` leaves `finding_verifications`, `finding_attack_paths`, `remediation_attempts` and `remediation_validations` behind on any connection where the `foreign_keys` pragma is off, the case this PR now adds a test matrix for, so it may be worth sharing one list between the two functions.

edorian:
Resolved merge conflicts against main

PR #1033, opened 2026-09-12
Add static runner capability preflight for skills

abhinavgautam01:
Part of #772

## Summary

Checks declared skill runtime requirements before spending model turns, allowing scans to fail early or proceed with explicitly reduced coverage.

This implements the static preflight portion of #772. Cached live backend probes and their recipe linkage remain follow-up work.

## Changes

- Add optional skill metadata:
  - `scrutineer.requires_commands`
  - `scrutineer.requires_features`
  - `scrutineer.degraded_mode`
- Validate and persist capability declarations, rejecting invalid executable names, unknown features, duplicate requirements and oversized lists.
- Check executable availability without running the required tools.
- Run container probes using the selected image and the same user, mounts, working directory, environment configuration and network policy as the agent.
- Check the local runner's own environment when running without a container.
- Persist worker-owned `ready`, `degraded` or `blocked` preflight results before agent launch.
- Stage the result in both copies of `context.json`.
- Fail blocked scans before model execution; permit missing capabilities only when degraded mode is explicitly enabled.
- Keep blocked or degraded coverage partial, including after coverage reconciliation and subsequent repair turns.
- Preserve existing behavior for skills without requirements.
- Document the metadata, coverage fields, and capability limitations.

## Capability Semantics

`network-egress` indicates configured network-policy access, not verified destination reachability, credentials or model availability. Existing allowlists remain authoritative.

Current runners report `docker-in-docker` and `fuse` as unavailable because they do not provision the necessary runtime support. Declaring requirements never grants privileges, mounts host sockets or relaxes network restrictions.

Probe execution failures, cancellation, malformed output and persistence failures stop execution even when degraded mode is enabled.

## Validation

- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `gofmt` and `git diff --check`
- Live Docker integration checks covering ready, blocked, degraded, network-policy and unsupported-feature cases.


internal/worker/capability_preflight.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		return stageCapabilityPreflight(sj.WorkRoot, sj.SkillDir, result)
+	}
+}
+
+func stageCapabilityPreflight(workRoot, skillDir string, result coverage.Preflight) error {
+	// Keep both copies in sync, including after whole-tree restaging or a retry.
+	for _, dir := range []string{workRoot, skillDir} {
+		path := filepath.Join(dir, "context.json")
+		data, err := os.ReadFile(path)
```
alexandre-daubois, line 172: `os.ReadFile` and the `os.WriteFile` below follow symlinks, and on the second `RunSkill` of `runSkillWithFallback` the `context.json` at `workRoot` is whatever the agent left in the read-write `/work` mount, since `reStageWholeTree` restages `src/` alone. A skill that swapped it for a symlink redirects this rewrite outside the workspace, and one that removed it turns the whole-tree fallback into a `record capability preflight` failure, so it may be worth lstat-ing the path and refusing a non-regular file.

andrew (changes requested):
This matches the #772 spec well — probe inside the same container config as the agent, record the result in coverage, cap completeness. The validation on command names and the closed feature set are right.

One security issue blocks merge, and it is the case Alexandre already flagged inline:

**`stageCapabilityPreflight` writes through a path the agent controls between runs.** Both `workRoot/context.json` and `skillDir/context.json` are under the RW `/work` mount (`skillDir` resolves to `workRoot/.claude/skills/<name>`). `runSkillWithFallback` calls `RunSkill` twice; between the two, the first agent has had write access and `reStageWholeTree` only replaces `src/`. `os.ReadFile`/`os.WriteFile` both follow symlinks, so a skill that swaps `context.json` for a symlink gets a host-side overwrite of whatever it points at when the second preflight fires. A skill that just deletes the file makes the fallback fail where it succeeds on main.

Simplest fix: run the probe once per scan. Nil out `RecordPreflight` (or the requirement lists) after it fires — the second run uses the same image, env, and network, so the answer cannot differ. That also removes a redundant container spin-up per fallback. For belt-and-braces on the write side, open with `O_WRONLY|O_CREATE|O_TRUNC|O_NOFOLLOW` (or `Lstat` and refuse a non-regular file) rather than `os.WriteFile`, and build the JSON from worker state instead of re-reading a file the agent may have touched.

Smaller:
- `docs/skills.md` — the new section is much denser than its neighbours. Trim to what the three keys do, the feature names and what `network-egress` means, and blocked-fails / degraded-caps-partial. The probe-failure and per-invocation caveats fit in one sentence each.
- The probe reuses `runBase` in full, which includes the model API credential env. The shell script does not need it; a slimmer arg set (image, user, mounts, workdir, network) would keep the token out of the probe container. Not blocking, but worth doing while you are in here.
- Drop the "checked again on … whole-tree fallbacks" line from the docs once the once-per-scan change is in.

internal/worker/capability_preflight.go
abhinavgautam01: Sorry for the delay. Fixed by running preflight once across fallback and repair, building context from worker-owned state and using rooted exclusive file replacements. Added regression tests for symlinks, hard links, deleted files and malformed context.

abhinavgautam01:
Thanks Addressed the security issue and smaller notes: preflight now runs once per attempt, context writes use worker-owned data and safe replacements, probe containers omit model credentials/state mounts, and the docs are trimmed.

andrew (changes requested):
Security fix looks right — `sync.Once` on the shared `*capabilityPreflightState` means the second `RunSkill` never touches the filesystem, `writeSkillContext` goes through `replaceWorkspaceFile` (`os.OpenRoot` → `RemoveAll` → `O_CREATE|O_EXCL`) so an agent-planted link is unlinked rather than followed, and `RecordPreflight` closes over the worker-built `skillContext` instead of re-reading. The probe container args are correctly slimmed to `buildContainerBaseArgs` with no model creds or harness-state mount.

Two cleanups from the refactor:

- `internal/worker/skill.go` — `stageContext` and `stageContextWithInputs` have no production callers now that `stageExploratoryWorkspace`, `doExposure` and `stageWorkspaceWithInputs` all call `buildSkillContext` + `writeSkillContext` directly. Delete both, and drop the three comments that still name `stageContext` (`skill.go` around 1245 and 1511, `exposure.go` around 125).
- `docs/skills.md` — drop "Cached live backend probes remain separate work."; docs shouldn't reference planned work.

Please also squash the fix-up and merge commit into one when you push.

abhinavgautam01:
Removed both unused wrappers and stale comments, updated the tests to exercise the build/write functions directly and dropped the planned-work sentence. Squashed into one commit.

andrew (approved):

PR #1041, opened 2026-09-14
feat: record the producing model on findings

dkw-oss:
Denormalise the producing scan's model onto Finding (backfilled on startup), expose it in the JSONL findings exports and sharing bundles (preserved across import), show it on the findings list and detail pages, and add a per-model breakdown to /reporting and its CSV/JSON exports. All additive; no interface reshaped.

cc @p-linnane implemented as discussed recently.


dkw-oss:
Example:

```
$ curl -s http://127.0.0.1:8080/api/v1/findings | jq -c '{model}' | sort | uniq
{"model":""}
{"model":"claude-opus-4-6"}
{"model":"claude-opus-5"}
{"model":"claude-sonnet-4-6"}
```

internal/web/reporting.go
```diff
@@ -564,6 +670,23 @@ func (s *Server) reportingCSV(w http.ResponseWriter, r *http.Request) {
 		"avg_cost_usd":     num(data.AllTime.CostUSD),
 		"avg_total_tokens": num(data.AllTime.TotalTokens),
 	}))
+	// The per-model rows sit with the summary rows, above the daily table:
+	// they slice the same period total by model, not by day. An empty model
+	// cell on a model row is real data — activity from before models were
+	// recorded — not an unfilled column.
+	for _, m := range data.Models {
+		_ = cw.Write(row("model", "", map[string]string{
+			"model":            m.Model,
```
p-linnane: **[P2] Protect imported model values in the CSV export**

`Model` can contain externally supplied bundle text. Importing `"model":"=1+1"` emits an unchanged formula into the reporting CSV. Please protect this cell against spreadsheet formula interpretation and add import-to-CSV regression coverage. CSV quoting alone is insufficient.

internal/web/server.go
```diff
@@ -1234,18 +1236,21 @@ func (s *Server) findings(w http.ResponseWriter, r *http.Request) {
 
 	reposByID := loadRepoMap(s.DB, rows, findingRepoID)
 	anySubPath := false
+	anyModel := false
 	for _, r := range rows {
 		if r.SubPath != "" {
 			anySubPath = true
-			break
+		}
+		if r.Model != "" {
```
p-linnane, line 1248: **[P3] Keep the model column visible while sorting by it**

With 20 unattributed findings and one newer attributed finding, clicking Model sorts empty values onto page one and hides the column, including its direction toggle. Please keep it visible whenever model sorting is active.

internal/web/templates/reporting.html
```diff
@@ -137,6 +137,46 @@ <h3 class="text-lg font-semibold mb-1">Cost averages</h3>
   </div>
 {{end}}
 
+<h3 class="text-lg font-semibold mb-1">By model</h3>
+<p class="text-sm text-muted-foreground mb-3">
+  The period's activity sliced by model. Scan runs and cost group by the model each
+  run used; findings group by the model that first produced them, so findings imported
+  from another instance's bundle count under the model that found them there, not the
+  local import run. Rows without a model predate model recording.
```
p-linnane: **[P3] Distinguish missing model attribution from legacy data**

Empty models don't necessarily predate model recording: deterministic imports still produce them today. Suggest "Rows without a model have no model attribution recorded." Related documentation should distinguish those imports from the LLM ingest fallback, which records its ingest model.

p-linnane (changes requested):
One CSV export issue should be fixed before merging. Two smaller comments below.

internal/web/templates/reporting.html
dkw-oss: Actioned, ty 66ba9d3de2b516a4a023849f48118679d0bf1955

internal/web/server.go
dkw-oss: Actioned, ty 66ba9d3de2b516a4a023849f48118679d0bf1955

internal/web/reporting.go
dkw-oss: Good pickup, fixed c53944e65fec4249303369c364e94d35695892c3

internal/ingest/ingest.go
```diff
@@ -43,6 +43,66 @@ func normaliseSeverity(s string) string {
...
+func normaliseModel(s string) string {
+	s = strings.TrimSpace(s)
+	if s == "" || len(s) > maxModelLen {
+		return ""
+	}
+	if !isAlphanumeric(rune(s[0])) {
+		return ""
+	}
+	for _, r := range s {
+		if !isAlphanumeric(r) && !strings.ContainsRune("._:/@+-_", r) {
```
p-linnane: The earlier three comments are addressed. One regression remains in `normaliseModel`: the character allowlist rejects `[` and `]`, so built-in IDs such as `claude-fable-5-1[1m]` silently become empty on bundle import.

I reproduced this with an export/import round-trip test that passes on the previous head and fails on `fd02b6e`. Please preserve the supported `[1m]` suffix and add coverage using the built-in model IDs.

dkw-oss: 🙏 849f562b1e659871002e84cdabe13c5aded3412c

p-linnane (approved):
Looks good. Thanks @dkw-oss!

PR #1042, opened 2026-09-14
Add opt-in pause-on-overage policy for model scans

### LINKED ISSUE #946: `pause_on_overage` option
The existing `downgrade_on_overage` option is helpful, but I'd like to completely pause on overage instead of just downgrading

abhinavgautam01:
Fixes #946

## Summary

Adds an opt-in `pause_on_overage` policy that pauses model scans when Claude reports subscription overage, instead of continuing on a cheaper model.

The option defaults to false and takes precedence over `downgrade_on_overage`.

## Changes

- Add `pause_on_overage` configuration and the `-pause-on-overage` CLI flag, preserving explicit CLI overrides.
- Stop active scan jobs while preserving session IDs, partial reports and workspaces for resumption.
- Pause queued skill and exposure scans and prevent new dispatch while the overage hold remains active.
- Resume policy-paused scans after the latest applicable reliable reset, using the existing auto-resume buffer.
- Leave scans paused for operator action when reset information is missing or implausible.
- Recover outstanding overage holds from persisted scan pauses after restart.
- Preserve manual cancellations, maintainer opt-outs, manual pauses and existing account-error handling.
- Display the pause state on the Jobs and Usage pages.
- Document configuration, precedence, recovery behavior and limitations.
- Add regression coverage for concurrent cancellation, dispatch gating, restart behavior, reset selection, configuration precedence, state preservation and UI rendering.

The policy relies on provider-reported overage signals. It is not a hard monetary budget and cannot guarantee zero overage charges before detection or while cancellation takes effect.

## Validation

- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `golangci-lint v2.13.2 run`
- `go mod tidy -diff`
- `gofmt` verification
- `git diff --check`


andrew (changes requested):
Thanks for this — the approach is right (reuse the account-pause resume timer, cancel with a distinguished reason, gate dispatch) and the test coverage is thorough. A few things to sort before it lands:

**Drop the unrelated CI commit.** `c4cd094b` (retry-build-command) is build infrastructure unrelated to #946. Please open it as its own PR if you want it considered; drop it from this branch.

**`ShouldPauseOnOverage()` runs a DB query per call.** It is called from the jobs handler, which re-renders on every SSE `scan-status` event, so this is a `SELECT ... WHERE status = ? AND error LIKE ?` per refresh. `ShouldDowngradeModel()` next to it is in-memory; keep this one the same cost class. The persisted-hold check is only needed to survive a restart, so read it once at worker start (`scheduleNextAccountResume` already scans the same rows — set a bool there) and make the runtime check `w.OnOverage() || w.persistedOverageHold`.

**`overageMu` serializes all dispatch and finalize whenever the flag is on**, even while not on overage. `preflightSkill` does several DB reads and possibly a queue write; `finalizeScan` writes the row, audit events and findings — holding a global mutex across those turns parallel workers into a queue.
- `startScanUnlessOverage`: the conditional claim in `startScan` already returns `errScanClaimLost` if the row was paused underneath it, so the mutex adds nothing here. Check state, pause rows if active, otherwise call `startScan` — no lock needed.
- `preflightSkillUnlessOverage`: fix the actual race by making `failScanPrereqs` conditional on `status = queued` instead of holding the mutex across the whole preflight.
- `finalizeScan`: only the `OveragePauseReason` branch touches overage state; do not take the mutex on the common path.

Smaller:
- `docs/usage.md` — the new section is longer than the rest of the file combined. Trim to what the flag does, precedence over `downgrade_on_overage`, and that a missing/implausible reset needs manual resume.
- `resumeAccountPaused` now also gates the update on `error` and `paused_until`, which changes the existing account-pause resume path. If that tightening is not needed for overage, revert to `id + status`.
- `pauseOverageRows` sets `finished_at` on a queued→paused row; the other pause paths do not. Drop it unless the resume path relies on it.

Please also squash to a single commit when you push the update.

abhinavgautam01:
Thanks, Addressed all points: overage checks now use cached state, ordinary scan work no longer holds the overage mutex and prerequisite failures only update queued scans. Restored the resume predicate, removed queued-pause finish timestamps, trimmed the docs and dropped the unrelated CI changes. Squashed into one commit; tests, race tests, vet and lint pass.

andrew (changes requested):
Thanks — all the round-1 points are addressed (CI split out, `ShouldPauseOnOverage` in-memory, `overageMu` off the ordinary dispatch/finalize path, `failScanPrereqs` conditional on `status = queued`, resume predicate restored, `finished_at` dropped, docs trimmed). Two small things from the fix-up:

- `internal/web/templates/job_list.html` — the new banner is `<section class="alert mb-4"><p>…</p></section>`; the account-paused banner directly above and the downgrade banner directly below are both `<div class="alert-{destructive,warning} mb-6"><i/><section><strong>Title.</strong> Body.</section></div>`. Match that structure (`alert-warning`, `mb-6`, `<strong>Subscription overage.</strong>`).
- `internal/worker/worker.go` — `scheduleNextAccountResume`'s DB-error branch now calls `setPersistedOverageHold(true, nil)` and `logOverageError` unconditionally, so an operator with `pause_on_overage` off who hits a transient startup DB error gets a `persist overage pause` log line for a feature they never enabled. Gate those two on `w.PauseOnOverage`; keep the `scheduleAccountResumeAfter` retry for both paths since it's an improvement over the old silent return.

abhinavgautam01:
Updated the banner to match the existing warning layout and gated the overage hold/logging on PauseOnOverage, keeping retries for both paths. Added regression coverage; tests and lint pass. Kept the PR as a single commit.

andrew (approved):

PR #1051, opened 2026-09-16
Retry Go tool installation failures in runner image builds

abhinavgautam01:
## Summary

Adds bounded retries to Go tool installation steps in `Dockerfile.runner` to tolerate transient Go module proxy and checksum service failures.

Split out from #1042 as requested during review.

## Changes

- Add a reusable shell helper that attempts a command up to three times, waiting 5 and 10 seconds between retries.
- Apply the helper to the pinned `git-pkgs` and `brief` installations.
- Preserve command arguments, environment variables and the final failure exit code.
- Keep Go module checksum verification enabled and leave tool versions unchanged.
- Run retry-helper tests in the runner-image workflow.

The helper retries any nonzero exit, not only network errors. Persistent failures still fail the build after three attempts.

## Validation

- Retry tests covering immediate success, recovery, exhausted retries, argument and environment preservation and missing-command handling.
- Shell syntax checks.
- Runner CLI version-pin checks.
- Offline Zizmor scan of the modified workflow.
- `git diff --check upstream/main...HEAD`.


andrew (changes requested):
The retry helper itself is fine. Three things to bring in line with the rest of the repo:

- `scripts/retry-build-command_test.sh` — the two existing shell tests (`scripts/release-resolver_test.sh`, `scripts/wait-for-runner-image_test.sh`) are `#!/usr/bin/env bash` with `assert_eq`/`fail` helpers that print expected vs actual on failure. This one uses bare `[ ... ]` under `set -eu`, so a mismatch exits 1 with no indication which `check` call or which assertion failed. Please rewrite `check` to use the same `assert_eq` pattern as the other two. Keeping `retry-build-command.sh` itself as `/bin/sh` is correct since it runs inside the `golang:trixie` stage.
- `.github/workflows/runner-image.yml` — the other `_test.sh` scripts run in `tests.yml`; this one is inside the arch matrix so it runs once per arch. Move the step next to the existing shell tests in `tests.yml`.
- `Dockerfile.runner:28` — "keep verification enabled" documents something the change doesn't do. Trim to `# Go proxy/checksum requests can fail transiently.`

abhinavgautam01:
Thanks, addressed all three points: switched the test to Bash with named expected/actual assertions, moved it alongside the existing shell tests in tests.yml, and trimmed the Dockerfile comment. The retry helper remains /bin/sh.

andrew (changes requested):
Thanks — the three earlier points are all addressed and the govulncheck wrap is a sensible extension.

One leftover of the same pattern:

`scripts/retry-build-command.sh:2` — `without changing their verification settings` documents something the script doesn't do (it never touches GONOSUMCHECK/GOFLAGS). Please trim to:

```sh
# Retry idempotent build commands.
```

abhinavgautam01:
Trimmed the comment as requested. Shell syntax checks and all three shell test suites pass.

andrew (changes requested):
Thanks — the round-2 comment trim is done and the helper/test/govulncheck wrap all look right.

Two things before this can land:

**`Dockerfile.runner`** now conflicts with main: #1052 added a `go install github.com/betterleaks/betterleaks` line in the same `go-tools` stage. Please rebase and wrap that line with the retry helper too (same proxy/sumdb path, same rationale). Resolved shape:

```dockerfile
COPY scripts/retry-build-command.sh /usr/local/bin/retry-build-command.sh
# Go proxy/checksum requests can fail transiently.
ARG BETTERLEAKS_VERSION=v1.8.1
RUN GOBIN=/out CGO_ENABLED=0 sh /usr/local/bin/retry-build-command.sh go install -ldflags "-X=github.com/betterleaks/betterleaks/version.Version=${BETTERLEAKS_VERSION}" github.com/betterleaks/betterleaks@${BETTERLEAKS_VERSION}
RUN GOBIN=/out sh /usr/local/bin/retry-build-command.sh go install github.com/git-pkgs/git-pkgs@v0.20.0
RUN GOBIN=/out sh /usr/local/bin/retry-build-command.sh go install github.com/git-pkgs/brief/cmd/brief@v0.13.0
```

**`internal/worker/patch_gate_test.go`** — the flake this fixes is real (`TestGitApplyCheck_ignoresRunnerInstalledFilter` on macOS: `lstat .../.git/objects/maintenance.lock: no such file or directory`, https://github.com/alpha-omega-security/scrutineer/actions/runs/35203347507) and the fix is correct, but it's a pre-existing main-branch issue unrelated to build retries, and squash-merge would bury it under this PR's title. Please open it as its own PR against main (titled for the flake, linking that run), let that merge first, then rebase this PR on top of it so CI stays green here.

abhinavgautam01:
@andrew, opened the maintenance-flake fix separately #1062, with the failing CI run linked. Tests and lint pass. I’ll wait for it to merge, then rebase this PR and wrap the Betterleaks install as requested.

andrew:
#1062 has merged (https://github.com/alpha-omega-security/scrutineer/pull/1062), so this is unblocked. Please rebase onto `main`, drop the `internal/worker/patch_gate_test.go` hunk (superseded by the `testutil.GitEnv` change), and wrap the `betterleaks` `go install` line with the retry helper as in the previous review — that should also clear the `Dockerfile.runner` conflict.

abhinavgautam01:
Rebased onto main after #1062, dropped the superseded patch-gate change, and resolved the Dockerfile conflict by wrapping Betterleaks with the retry helper. Tests and lint pass; image builds remain for CI.

PR #1062, opened 2026-09-18
Fix macOS patch-gate test flake caused by Git maintenance

abhinavgautam01:
## Summary

Fix a macOS CI flake in `TestGitApplyCheck_ignoresRunnerInstalledFilter` by disabling automatic Git maintenance and garbage collection in the temporary repositories created by `gateRepo`.

The fixture commits files before copying the repository, including `.git`, into a scan workspace. Background Git maintenance can remove `.git/objects/maintenance.lock` during that copy, causing `CopyTree` to fail with `lstat .../maintenance.lock: no such file or directory` before the filter-safety assertions run.

## Changes

- Set repository-local `maintenance.auto=false` and `gc.auto=0` before creating fixture commits.
- Keep the existing patch validation and runner-installed filter assertions unchanged.
- Leave production Git configuration and copy behavior unchanged.

## Context

Extracted from #1051 at Andrew's request so this pre-existing test flake is reviewed independently of the build-retry changes. The original failure is recorded in [the macOS CI job](https://github.com/alpha-omega-security/scrutineer/actions/runs/35203347507/job/105143112563).

After this PR merges, #1051 can be rebased onto it without carrying a separate test-fixture change.

## Validation

- `go test -race -gcflags='modernc.org/...=-d=checkptr=0' ./internal/worker -run 'Test(GitApplyCheck|GatePatch)' -count=20`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `golangci-lint v2.13.2 run`
- `gofmt` and `git diff --check`


andrew (changes requested):
Thanks for splitting this out. The mechanism is right — `git commit`/`git checkout` fork `git maintenance run --auto` by default (independent of the global/system config `testutil.GitEnv` already suppresses), that briefly creates and removes `.git/objects/maintenance.lock`, and `CopyTree`'s `WalkDir` can readdir the entry then fail the `lstat`.

But `gateRepo` isn't the only fixture with this window: `internal/worker/diff_rescan_test.go` does the same commit-then-`CopyTree` on a `.git` (`initGitRepo`/`gitCommit` at ~L265–278, copied at L48 and L151) and has the identical latent flake. Patching `gateRepo` alone means the next macOS failure is that test with the same three lines pasted in.

Please move the suppression into `internal/testutil/gitenv.go` instead, which already exists to make test git deterministic regardless of host — append to the returned env:

```go
"GIT_CONFIG_COUNT=2",
"GIT_CONFIG_KEY_0=maintenance.auto", "GIT_CONFIG_VALUE_0=false",
"GIT_CONFIG_KEY_1=gc.auto",          "GIT_CONFIG_VALUE_1=0",
```

and drop the three added lines and comment from `patch_gate_test.go`. Same net size, covers every fixture that copies a `.git`, no per-helper comment needed. `GIT_CONFIG_COUNT` is Git ≥ 2.31; both CI images are well past that.

abhinavgautam01:
Thanks, moved suppression into testutil.GitEnv and removed the per-fixture settings. The maintenance options are appended to preserve existing fixture URL rewrites. Added regression coverage; repeated fixture tests, the full race suite and lint pass.

andrew (approved):
Thanks. Appending after the ambient `GIT_CONFIG_COUNT` was the right call over what I suggested — hardcoding indices 0/1 would have clobbered the `url.insteadOf` / `protocol.file.allow` rewrites that `internal/skills/git_test.go` and `internal/worker/repo_cache_test.go` set via `t.Setenv` before layering `GitEnv()` on top. Rewrapped the doc comment to match its neighbours and amended.

Issue #728, opened 2026-07-27
Read scan transcripts back into the threat-model contract


andrew:
Each scan writes a transcript and we keep them, but nothing reads them back. When `security-deep-dive` wastes fifty turns on a file that turns out to be generated, or `verify` fails three times on the same missing build dependency, the next scan of that repo repeats the mistake.

Add a `reflect` skill that runs at the end of `triage`, reads the transcripts of the scans it enqueued, and writes short typed notes back into the repo's threat-model contract:

- tool failures: "sandbox is missing `libfoo-dev`; verify can't build"
- wrong assumptions: "researcher treated `params[:id]` as attacker-controlled; it's set by middleware in `app/middleware/tenant.rb`"
- successful patterns: "the working reproducers for this repo all go through `bin/dev-server`, not `rackup`"

The transcript files are large, so the skill greps them for tool errors, "cannot" / "unable" / "giving up" markers, and the last few turns of each scan rather than reading them whole. One note per stage per run at most. If a stage's transcript is missing or unreadable, record that instead of nothing so the output is never silently empty.

Consumers: `threat-model` folds the notes into the contract on its next run; `security-deep-dive` and `verify` already load the contract.

Source: [google/mantis](https://github.com/google/mantis) `mantis-reflect/SKILL.md` (Apache-2.0). Their version writes to a `learnings.jsonl` inbox that a KB-builder stage clears; we'd write straight into the threat-model contract since that's the persistent per-repo store we already have.

Related: #249 (threat-model refinement; this is one of the inputs that tool would test), #37 (the same transcript pass could compute tool-error rate and give-up rate as per-skill metrics).

andrew:
[openai/codex-security](https://github.com/openai/codex-security/tree/f22d4a36f26d16287bcdfd707b369116e02a08c3) does a narrower version of this that might be worth landing first: it stores analyst false-positive decisions with a required reason, then before a later validation run loads a bounded set of prior FPs for the same target and lets validation dismiss a match only if the recorded reason still holds against the current code. See [workbench feedback loading](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/workbench_feedback.py) and the [validation skill](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/skills/validation/SKILL.md).

[...]

PR #1065, opened 2026-09-19
Add analyst rejection feedback for subsequent validation

abhinavgautam01:
Part of #728

## Summary

Adds structured analyst rejection feedback that subsequent validation runs can reconsider against current code.

This implements the narrower analyst-feedback step suggested in #728. Transcript reflection and automatic promotion into
`known_non_findings` remain follow-up work.

## Changes

- Require an explicit decision and reason when rejecting a finding through the UI.
- Distinguish false positives from already-fixed and other non-actionable findings.
- Save the review, status transition and lifecycle history atomically.
- Snapshot the source scan, commit, fingerprint, repository-relative path and CWE at review time.
- Provide at most 20 relevant false-positive decisions to finding-scoped `revalidate` and diff-mode `security-deep-dive` runs.
- Match feedback within the same repository and relevant files, including old paths for renamed files.
- Exclude legacy reviews without source snapshots, superseded decisions, reopened findings and decisions retired by an earlier reopening.
- Require skills to independently cite current-code evidence before applying an old rejection reason.
- Document the review metadata and skill-context contract.

Feedback is historical evidence, not an automatic suppression rule. Full and exploratory audits remain unchanged and no decisions are automatically promoted into the threat model.

## Tests

- `go test ./...`
- Race-enabled regression tests for rejection, rollback, feedback selection and context staging.
- Full web race-test rerun.
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- `golangci-lint v2.13.2`: zero issues.
- `gofmt` and `git diff --check`.


andrew (changes requested):
Direction and tests look good; three mechanical items before merge:

- Reason length is validated in bytes (`len(reason) > MaxReviewReasonBytes` in `internal/db/audit.go`) but the dialog's `maxlength="4096"` counts characters, so a long multibyte reason passes the form and comes back as a plain 422 that loses the typed input. Count runes server-side, and make the docs consistent (docs/skills.md says characters, docs/database.md says bytes).
- `stageWorkspaceWithInputs` takes `feedback ...db.FindingFeedback` — variadic only to avoid updating the two test call sites. Make it a plain slice parameter and update the callers.
- The combined check in `AddFindingReview` returns one message for two different failures; a too-long reason on an `already_fixed` review is told "false-positive reviews require a reason". Split into two checks with distinct messages.

abhinavgautam01:
Thanks, addressed all three: rune-based reason limits with consistent docs, separate validation messages and an explicit feedback slice with callers updated. Added Unicode boundary and error-message regression tests.

internal/worker/finding_feedback.go
```diff
@@ -0,0 +1,51 @@
...
+	switch {
+	case skill.Name == "revalidate" && scan.FindingID != nil:
+		var finding db.Finding
+		if err := w.DB.WithContext(ctx).Where("repository_id = ?", scan.RepositoryID).First(&finding, *scan.FindingID).Error; err != nil {
+			return nil, fmt.Errorf("load finding for analyst feedback: %w", err)
+		}
+		if path := findingnorm.FindingPath(finding.SubPath, finding.Location); path != "" {
+			paths = append(paths, path)
+		}
+	case skill.Name == "security-deep-dive" && scan.RescanMode == db.ScanRescanModeDiff:
```
alexandre-daubois: These two cases match on string literals while the same package already declares `revalidateSkillName` (`novelty.go`) and `deepDiveSkillName` (`skill.go`) for exactly these names. A rename that updates the constants leaves this switch matching nothing, so analyst feedback stops reaching both skills while `novelty.go`'s own guard keeps working and nothing fails. You can switch both cases to the existing constants.

internal/web/templates/finding_workflow.html
```diff
@@ -257,6 +254,36 @@
...
+    <header><h2>Reject finding</h2></header>
+    <form method="post" action="/findings/{{$id}}/status" class="form grid gap-4">
+      <input type="hidden" name="status" value="rejected">
+      <div class="grid gap-2">
+        <label class="label" for="rejection-verdict-{{$id}}">Decision</label>
+        <select class="input" id="rejection-verdict-{{$id}}" name="verdict" required>
+          <option value="">Select a decision</option>
+          <option value="false_positive">False positive</option>
+          <option value="already_fixed">Already fixed</option>
+          <option value="uncertain">Other / not actionable</option>
```
alexandre-daubois, line 268: This option stores `uncertain`, and `ComputeAuditMetrics` counts a review as agreement whenever `verdict = automated_outcome`, so a rejection labelled `Other / not actionable` scores as agreement on every finding where `revalidate` also returned `uncertain`. Browser rejections wrote no `finding_reviews` row before this change, so the agreement rate on `/audit` now absorbs decisions the analyst never made about the automation. It may be worth storing rejection-dialog reviews under a verdict the agreement query skips, or leaving `automated_outcome` empty for them.

internal/db/db.go
```diff
@@ -1291,6 +1291,12 @@ type FindingReview struct {
 	// outcomes since there is nothing to compare to.
 	AutomatedOutcome string `gorm:"index"`
 	Reviewer         string
+	// Snapshot the observation being reviewed; later rescans may move the case.
+	SourceScanID       uint   `json:"source_scan_id"`
+	SourceCommit       string `json:"source_commit"`
+	FindingFingerprint string `json:"finding_fingerprint"`
+	FindingPath        string `gorm:"index" json:"finding_path"`
+	CWE                string `json:"cwe"`
```
alexandre-daubois, line 1301: These are the only json-tagged fields on the struct, so `GET /findings/{id}/reviews` answers `{"ID":1,"Verdict":"false_positive",...,"source_scan_id":0,"finding_path":""}`. The `FindingReview` schema this PR extends spells every key in snake_case, so a generated client picks up the five new fields and misses `id`, `verdict`, `reason`, `automated_outcome`, `reviewer` and `created_at`. Tagging the older fields here would make the response match the contract the same PR grew.

abhinavgautam01: All FindingReview fields now serialize with the documented snake_case names. The API regression test checks the actual response keys and values directly.

internal/web/templates/finding_workflow.html
abhinavgautam01: Rejection-dialog decisions now leave automated_outcome empty, excluding them from agreement calculations while preserving their feedback records. Added regression coverage for all three rejection options.

internal/worker/finding_feedback.go
abhinavgautam01: Both feedback guards now use the existing revalidateSkillName and deepDiveSkillName constants. Worker feedback tests pass.

andrew (approved):

PR #1067, opened 2026-09-21
Fix advisory exports including unrelated monorepo packages

### LINKED ISSUE #1049: Monorepo Finding Exports Apply Advisory Data to Unrelated Repository Packages
## Summary

OSV, CSAF and disclosure-bundle exports load every package in a finding's repository without consulting the finding's subproject attribution. A finding scoped to one monorepo component can therefore apply vulnerability information to packages belonging to unrelated components.

### Affected Code

- `internal/web/finding_osv.go`: `findingOSV` and `osvAffectedList`.
- `internal/web/finding_csaf.go`: `findingCSAF`, `buildCSAF` and its product-status/remediation builders.
- `internal/web/finding_bundle.go`: `bundleEntriesAt`.
- `internal/db/db.go`: `Finding.SubPath`, `Package.SubprojectID` and `Subproject`.

### Current Behavior

The three export paths query packages using only `repository_id = finding.RepositoryID`. They do not resolve `Finding.SubPath` or exclude packages attributed to other subprojects.

`osvAffectedList` creates an affected entry for each package with a supported ecosystem derived from its PURL. When the finding's fix version is a supported semantic-version anchor, that same fix boundary is added to each package entry.

The CSAF builder collects the repository's package product IDs and applies the finding's product status, scores and remediation to them. The status is not necessarily `known_affected`: it may be fixed, under investigation, or another classification depending on the finding. The bug is the overly broad package attribution, regardless of the selected status. CSAF export also requires recorded repository dependents.

Disclosure bundles reuse these builders with the same repository-wide package collection and inherit the attribution problem.

### Impact

An advisory about one monorepo package can incorrectly identify sibling packages as affected or fixed and recommend an unrelated fix version. This can mislead downstream consumers and advisory-processing tools.

### Reproduction to Add

1. Create a repository with `cli` and `web` subprojects.
2. Create a published package with a valid supported PURL for each subproject, setting each package's `SubprojectID` appropriately.
3. Create a finding with `SubPath = "cli"` and `FixVersion = "1.2.3"`.
4. Export `/findings/{id}/osv.json` and inspect `affected`: both packages are included instead of only the relevant package.
5. For CSAF, also seed a repository dependent, then inspect `/findings/{id}/csaf.json` for the same inappropriate inclusion in product classifications and remediation.
6. Inspect the advisory documents included in `/findings/{id}/bundle.tar.gz` for consistent behavior.

### Proposed Resolution

Use a shared, explicitly defined package-selection policy for per-finding advisory exports. When a finding has known subproject attribution, resolve the subproject within its repository and exclude packages positively attributed to unrelated subprojects.

Define conservative behavior for missing subprojects, unassigned packages, root-scoped findings and shared-code findings. A package with `subproject_id IS NULL` may simply have unknown attribution; it must not automatically be treated as affected solely because the subproject-specific query returned no packages. Likewise, a source path alone is not proof of the complete impact set for a shared-code vulnerability.

### Acceptance Criteria

- A finding associated with one subproject does not automatically include packages attributed to distinct sibling subprojects.
- OSV, CSAF and bundled advisory documents use a consistent selection policy.
- Tests use at least two subprojects and assert package identities, not merely output counts.
- Tests cover missing/unassigned attribution and explicitly defined root/shared-code behavior.
- Required database lookup failures remain errors rather than silently expanding the affected-package set.
- Generated documents continue to pass the existing OSV and CSAF schema validation.

abhinavgautam01:
Fixes #1049

## Summary

Scope published package identities in OSV, CSAF and disclosure-bundle exports to the finding's subproject instead of including every package in its repository.

## Changes

- Use one shared package-selection helper across all three export paths.
- Select only packages assigned to the finding's exact subproject within the same repository.
- Exclude sibling and nested subprojects unless explicitly selected by the finding's attribution.
- Avoid falling back to unassigned packages when a subproject is missing or has no assigned packages.
- Omit published package identities for root-scoped findings in repositories with recorded subprojects.
- Preserve existing unassigned-package behavior for root findings in repositories without recorded subprojects.
- Preserve repository-level OSV fallback, CSAF repository products, and explicitly recorded dependent assessments.
- Propagate database lookup errors instead of silently broadening package selection.
- Document the attribution policy and its limitations for undiscovered monorepos and shared-code impact.

## Coverage

- Exact and nested subproject attribution.
- Missing subprojects and subprojects without assigned packages.
- Root-scoped and shared-code findings.
- Cross-repository isolation and legacy root-package behavior.
- Package identities, fix ranges, CSAF classifications, scores and remediations.
- Standalone and bundled OSV/CSAF schema validation.
- Database lookup failures across all three export paths.

## Validation

- `go test ./...`
- `go vet ./...`
- `go vet -tags podman ./...`
- `go mod tidy -diff`
- golangci-lint v2.13.2: zero issues
- `gofmt`
- `git diff --check`


internal/web/finding_advisory_packages.go
```diff
@@ -0,0 +1,42 @@
...
+	} else {
+		var count int64
+		if err := gdb.Model(&db.Subproject{}).Where("repository_id = ?", f.RepositoryID).Count(&count).Error; err != nil {
+			return nil, err
+		}
+		if count != 0 {
+			return nil, nil
+		}
+		// Preserve root-package exports for repositories without subprojects.
+		q = q.Where("subproject_id IS NULL")
```
alexandre-daubois: Both branches key on `Package.SubprojectID`, which stays NULL for every row when the operator runs with `-monorepo-attribution=false`: `parseSubprojectsOutput` still writes the `subprojects` rows, but `reconcileSubprojectLinksIfEnabled` returns before linking anything. A repository with recorded subprojects then yields zero packages for a sub-path finding (`subproject_id = sub.ID` matches nothing) and zero for a root finding (`count != 0`), so OSV, CSAF and bundle exports lose every package identity and fall back to the repository GIT range, which is the opposite of the repo-wide behaviour that flag documents. You can key the narrowing on whether the repository has any package linked to a subproject, so the flag keeps its documented fallback.

internal/web/finding_osv.go
```diff
@@ -101,8 +101,8 @@ func (s *Server) findingOSV(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "failed to load finding references", http.StatusInternalServerError)
 		return
 	}
-	var pkgs []db.Package
-	if err := s.DB.Select("name, ecosystem, p_url").Where("repository_id = ?", f.RepositoryID).Find(&pkgs).Error; err != nil {
+	pkgs, err := findingAdvisoryPackages(s.DB, f)
```
alexandre-daubois: The shared helper drops the `Select("name, ecosystem, p_url")` this path carried, so an OSV export now reads every column of every selected package, `Metadata` (`gorm:"type:text"`) included, to build entries that use three of them. It may be worth taking a column list on `findingAdvisoryPackages` so OSV keeps its narrow read while CSAF and the bundle get full rows.

abhinavgautam01: The helper now accepts a column list, restoring OSV’s name, ecosystem, and p_url projection while retaining full rows for CSAF and bundles. Added query-level regression tests.

internal/web/finding_advisory_packages.go
abhinavgautam01: Narrowing now depends on existing package-to-subproject links, preserving repository-wide exports when packages remain unlinked. Added standalone and bundle regression coverage for root and subpath findings.

andrew (approved):

PR #1073, opened 2026-09-22
Add a merge script for multi-instance /reporting JSON exports

dkw-oss:
Combines exports from discrete scanner instances into one corpus-wide report: counts summed, averages weighted by scans_averaged, and an optional --severity check that every input carries the same exported floor. Tested by scripts/merge-reports_test.sh, run in CI.


scripts/merge-reports.py
```diff
@@ -0,0 +1,350 @@
...
+    periods = [r.get("period", {}) for r in reports]
+    period_keys = ", ".join(sorted({p.get("key", "?") for p in periods}))
+    starts = [p.get("starts_at") for p in periods]
+
+    return {
+        "activity_by_day": by_day,
+        "activity_by_model": by_model,
+        "activity_in_period": in_period,
+        "cost_averages_per_scan": cost,
+        "filters": first.get("filters"),
```
alexandre-daubois: The merged report copies the first input's `filters` even after the divergence warning fired, so merging a `?severity=medium` export with an unfiltered one writes `minimum_severity: "Medium"` over `findings` counts that include everything below that floor. The usage the module docstring gives redirects stdout to a file, so the stderr warning is gone the moment the merged report is archived and the JSON itself records no trace of the mismatch. You can set `minimum_severity` to null when the inputs disagree, or carry each input's recorded floor into its `sources` entry.

dkw-oss: Ty, please see d451be903e5e4ec0c4a1e2c3353231410e0b35c5

andrew (changes requested):
The merge rules and the severity handling look right, and the fields `reporting.go` exports today are all either additive counts or `avg_*` values with a `scans_averaged` weight, so the output is correct as things stand.

The problem is how it's tied to the exporter. The script hard-codes the `report.json` shape and `merge-reports_test.sh` feeds it hand-written JSON, so a field added to `internal/web/reporting.go` later (a median, a rate) gets silently summed and nothing fails. It would also be the first Python in the repo's tooling; `scripts/` is shell and the Go-side tools are Go.

Please port it to Go as `scripts/merge-reports/main.go`, with a Go test that builds two real exports through `s.Handler()` at `/reporting/report.json`, merges them, and asserts the totals and weighted averages. A change to the export then breaks the build rather than the merged numbers. The shell test, the `scripts/__pycache__/` line in `.gitignore`, and the extra CI step can go; `go test ./...` picks up the new test.

